PR #21, opened 2016-07-20
Control the patient delete button with a global entry, defaults to off

teryhill:
Control the patient delete button with a global entry, defaults to off

### DIFF
--- a/interface/patient_file/deleter.php
+++ b/interface/patient_file/deleter.php
@@ -183,7 +183,7 @@ function popup_close() {
  if ($_POST['form_submit']) {
 
   if ($patient) {
-   if (!acl_check('admin', 'super')) die("Not authorized!");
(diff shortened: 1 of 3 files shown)


aethelwulffe:
Looks totally clean to me, but what about the database default?

tmccormi:
not sure what you mean @aethelwulffe about the database default?

aethelwulffe:
Meaning set the field with a default =1 value...in the table structure 
when it created...

CREATE TABLE `list_o_crap`

(`blah` varchar(42) NOT NULL,

`active` bool DEFAULT 1);

aethelwulffe:
I redact my stupid comment.  All the same, as an upgrade, the user would have had the ability to delete, and now they will not...unless they go looking into globals admin.  Just something to keep in mind.  In reality, it just highlights the need for lots more "setting up your practice" documentation. *sigh*

PR #22, opened 2016-07-20
Billing screen change, take 2.

teryhill:
Need instructions on how to successfully add menu Items and changes need to made to this for them before it is committed to the codebase

Added code from Aron at MI2 for the modal re-open and view log screens.
Moved EOB to the left nav as posting


tmccormi:
This is from email conversations with Kevin and should help get pointing in the right direction.  Most of this was about DYNAMICALLY adding menu items form "plugin's"  but should help.

There are two options:
1. Update the menu_items and menu_tree tables with the new entries you want after populating either by loading the menu.sql file in contrib/util/menu or executing the build_menu.php file in the same directory from the command line.  (The net result ought to be the same as far as rows in the two tables.)

2. Dynamically add the new entries you want the way forms and the current modules get added...
Here's where the forms get added:
`https://github.com/yehster/openemr/blob/iframe-dev-1/interface/main/tabs/menu/menu_updates.php#L43`

you can add another hook to a call back similar to this:
`https://github.com/yehster/openemr/blob/iframe-dev-1/interface/main/tabs/menu/menu_updates.php#L11`

AND/OR

Most straightforward way to get your entry added is to add the entry directly to the base JSON list.
In this file:
https://github.com/yehster/openemr/blob/iframe-dev-1/interface/main/tabs/menu/menu_data.php

find the ACL entry in the array, and create a similar entry for
search for: in the JSON string

`{\"label\":\"ACL\",\"menu_id\"`

Then paste another object like:
`
{\"label\":\"Tags\",\"menu_id\":\"adm0\",\"target\":\"adm\",\"url\":\"/interface/tags_filters/index.php?action=tags\",\"children\":[],\"requirement\":0} `
into the JSON string.

AND this

Ken asked: Is there a way to place the menu item in a specific spot, for example, after the ACL item?

If I do something like this, will it just append to the Administration section?
```
$menu_update_map=array();
$menu_update_map["Visit Forms"]="update_visit_forms";
$menu_update_map["Modules"]="update_modules_menu";
$menu_update_map["Administration"]="update_administration_menu";

function update_administration_menu(&$menu_list) ...

```
Kevin said ..
Yes, because you are just doing an array push, but you can manipulate the $menu_list->children array in other ways to add it where ever you want.  (You could build a completely new array for example, push the existing ones until you get to where you want to add your new entry, add that, then push the remaining).

teryhill:
there needs to be a GUI or something easier. The user base just collapsed by 50% . It needs to be easy enough that the common Oemr user can do it.
Just my 2 cents

teryhill:
Got it . The posting menu is now a default under billing. Moving again

aethelwulffe:
Merge away!

PR #24, opened 2016-07-22
Added option and converted to standard security model

teryhill:
Added option to Cash Receipts Report to show Patient Name instead of
invoice number.

Also converted the Cash Receipts Report to standard security model.

### DIFF
--- a/interface/billing/sl_receipts_report.php
+++ b/interface/billing/sl_receipts_report.php
@@ -7,6 +7,7 @@
  * but I wanted to make the code available to the project because
  * many other practices have this same need. - rod@sunsetsystems.com
  *
+ * Copyright (C) 2016 Terry Hill <terry@lillysystems.com>
  * Copyright (C) 2006-2010 Rod Roark <rod@sunsetsystems.com>
  *
  * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


interface/billing/sl_receipts_report.php
```diff
@@ -133,12 +138,12 @@ function sel_diagnosis() {
...
 </script>
 
-<title><?php xl('Cash Receipts by Provider','e')?></title>
+<title><?php echo xlt('Cash Receipts by Provider')?></title>
 </head>
 
 <body class="body_top">
 
-<span class='title'><?php xl('Report','e'); ?> - <?php xl('Cash Receipts by Provider','e'); ?></span>
+<span class='title'><?php echo xlt('Report'); ?> - <?php echo xlt('Cash Receipts by Provider'); ?></span>
```
aethelwulffe: I guess this separation is just for the translation engine.  Seems like there is a better option, like running a translation matrix against contiguous names of reports and features.

library/globals.inc.php
```diff
@@ -802,6 +802,12 @@
       '2',                              // default = 2
       xl('This will Display the Invoice Number in the Sales Report or the Patient Name and ID or Patient Name and Invoice Number.')
     ), 
+
+    'cash_receipts_report_invoice' => array(
+      xl('Display Invoice Number or Patient Name in the Cash Receipt Report'),
```
aethelwulffe: Cash Receipts Report...forgot an 's'.
I think being able to have both of these would be good.  I have been using both name and invoice in most reports.

aethelwulffe:
Code looks clean, and runs.  Basic security bits are Milestone 1 relevant, as is improving anything related to financial reporting and billing.  This is self contained in any case, and recommend pulling it.  All syntax seems to pass.

PR #32, opened 2016-07-29
Fix globals.inc

aethelwulffe:
TESTED AND WORKING
@teryhill  please confirm.

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -258,7 +258,7 @@ $GLOBALS_METADATA = array(
     'online_support_link' => array(
       xl('Online Support Link'),
       'text',                           // data type
-      'http://open-emr.org/',
+      'http://libreehr.github.io',
       xl('URL for LibreEHR support.')
     ),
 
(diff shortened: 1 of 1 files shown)


teryhill:
looks good to me ...

aethelwulffe:
Three party confirmation sort of leaves it to Tony or Ken to merge...

library/globals.inc.php
```diff
@@ -259,7 +259,7 @@
       xl('Online Support Link'),
       'text',                           // data type
       'http://open-emr.org/',
```
tmccormi: Need to actually change the URL ... before we merge this one

teryhill: to what?

tmccormi: I presume our website for now... but www.libreehr.org is broken 404 not found ....
could temp set it to http://libreehr.github.io

aethelwulffe:
URL is changed.  Planned on doing those in the next round of the LIBERATION branch, but figured that was a point of discussion.

PR #33, opened 2016-07-29
Liberation

aethelwulffe:
After manually adding the globals fix, this branch installed, ran, and tested very well.  There is some odd stuff in documentation, like the "History" piece that describes the history of OpenEMR up to "current 2.0", but I think that can be snuffed.
  This may require a rebase, but if it is possible, I would rather take care of conflicts on the other branches instead of this one.
  to-be-done:  Update web links.


tmccormi:
Not sure what to do with this until the conflicts are resolved.   too many file changes to look at them all, but it looks like a refactor by 'sed' so, probably OK.

aethelwulffe:
Conflict resolution taken care of.

aethelwulffe:
D@MMIT.  I just resolved the conflicts on this huge mess, and now it's stale again.

aethelwulffe:
Looking messy, but it took hours of verifying the changes and testing them.  Just one conflict, and that was in globals.inc.php

tmccormi:
First merge using my cellphone!

PR #34, opened 2016-08-01
Fac acl usability

aethelwulffe:
It's late, so this branch might need to be re-created.
This fixes a FAC_ACL issue where a provider can't see their whole schedule due to the lack of an "All facilities" selection in the calendar.  It also addresses a little issue in patient.inc where there was a semicolon instead of a comma.  Noticed that and didn't want to pass it up.


tmccormi:
This is a mess.   Shouldn't it only contain the 	FAC_ACL_usability1 commit?   I suggest you cherry-pick that commit into a new clean branch and resubmit the pull request

PR #40, opened 2016-08-04
remove 4010 standard

aethelwulffe:
snuffing 4010 code from gen837

### DIFF
--- a/library/gen_x12_837.inc.php
+++ b/library/gen_x12_837.inc.php
@@ -367,13 +367,13 @@ function gen_x12_837($pid, $encounter, &$log, $encounter_claim=false) {
   //   $log .= "*** CMS ID is missing for payer '" . $claim->payerName() . "'.\n";
   // }
 
-  if (true) { 
-    // The 5010 spec says:
-    // "Required when the payer address is available and the submitter intends
-    // for the claim to be printed on paper at the next EDI location (for example, a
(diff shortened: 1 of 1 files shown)


aethelwulffe:
Quite willing to resolve any conflicts after Liberation.

library/gen_x12_837.inc.php
```diff
@@ -440,11 +367,12 @@ function gen_x12_837($pid, $encounter, &$log, $encounter_claim=false) {
   //   $log .= "*** CMS ID is missing for payer '" . $claim->payerName() . "'.\n";
   // }
 
-  if (true) { // !$CMS_5010
+  if (true) {
```
tmccormi: Any good reason to leave this non if statement in?

tmccormi:
First code review via cellphone

library/gen_x12_837.inc.php
aethelwulffe: As it does nothing at all?  Well shouldn't we keep a little useless legacy code in?  Ain't that tradition?
Making changes now.

aethelwulffe: It is a block placeholder for a piece we really should revisit.  Maybe I should just wrap it in a todo set of comment lines.

aethelwulffe:
Changed to include a review comment.  This section will be attended to in a full re-write of the gen and parse scripts.

PR #42, opened 2016-08-04
database upgrade for libre

aethelwulffe:
Accidentally committed to master.  Reverted and pushed to branch commit.
Changes tables to libre conventions for upgrading openemr.  Need further
discussion on upgrades, conversions, patches and all that.


tmccormi:
I think the title of the issue is misleading.   try and be more clear about that next time :-)
Perhaps " Change calendar table name prefix to libreehr_

PR #44, opened 2016-08-04
SQL patch file removal

aethelwulffe:
This removes outdated or other files used to upgrade very old versions
of OpenEMR.  Little point in an automated upgrade set for 2.8 to 4.3,
since (as tested last year) the complete sequence results in a broken
database anyway.  Also removed was a phpmyadmin bookmark set from
defaults.sql.  Ippf is still up for review.


aethelwulffe:
Issue #47

PR #45, opened 2016-08-04
change version scheme

aethelwulffe:
Added a 0.0.1 to 1.0.0 upgrade that takes the good stuff from 420-431,
then adds the table name changes at the end.  Changed the version of the
application and database to 1.  Removed the clickoptions file and
conversion, as that is a much earlier version change.


aethelwulffe:
Issue #47

sql_upgrade.php
```diff
@@ -110,7 +110,7 @@
 foreach ($versions as $version => $filename) {
   echo " <option value='$version'";
   // Defaulting to most recent version, which is now 4.2.1.
```
tmccormi: bad comment, refers to old version #.  may change comment to say forked from version 4.2.1 or drop it

version.php
```diff
@@ -17,16 +17,14 @@
 // is a database change in the course of development.  It is used
 // internally to determine when a database upgrade is needed.
 //
-$v_database = 164;
+$v_database = 1;
```
tmccormi: Has database.sql been changed to be version 1 in the table?

aethelwulffe:
database.sql does the following:
INSERT INTO version (v_major, v_minor, v_patch, v_realpatch, v_tag, v_database, v_acl) VALUES (0, 0, 0, 0, '', 0, 0);

aethelwulffe:
Changed version comment.

PR #49, opened 2016-08-04
remove legacy .js from root

aethelwulffe:
CategoryTreeMenu.js had been changed to DocumentTreeMenu a long time
ago.  The template was still looking at the old location, where the
original file existed.  This all needs testingand investigation to see if the template is even used, but I am not seeing any issue with the document categories myself.


aethelwulffe:
Well, it IS best to not be too close to an old trailer that's on fhir...

PR #56, opened 2016-08-05
Trying to do this in chunks so it can be checked

teryhill:
Trying to do this in chunks so it can be checked with out much time
involved


aethelwulffe:
Testing good.

Issue #54, opened 2016-08-05
Setup.php has wrong email address


tmccormi:
_you would email us at **president@oemr.org** if you have installed this software_

Do we have a email or should we just change/drop this whole sentence and refer them to the website and github accounts?

teryhill:
website and/or github is my vote

aethelwulffe:
This is resolved fully now.

PR #60, opened 2016-08-05
change contact information #54

aethelwulffe:
Issue #54


tmccormi:
Personally not fond of this default choice and my customers all surveyed as prefering the light.css theme as more modern and cleaner ...

aethelwulffe:
Clean I would argue with.  White tends to blend with white.  If there is 
no consensus, I suppose I am stuck putting a style choice in the install 
process, because with the tabs interface (and frames), I got cries of 
outrage.  I find "Modern" is an insulting choice of words to use as a 
preferential comparative statement as well.  Turn it green and black, 
and it becomes a VT100.  Modern. Some of us like having divisions 
between choices.

It is important enough to me to invest in a configuration choice upon 
installation.  I will come back to this.

PR #62, opened 2016-08-05
Last chunk I think

teryhill:
Last chunk I think


interface/patient_file/history/encounters.php
```diff
@@ -152,14 +152,6 @@ function showDocument(&$drow) {
     htmlspecialchars( xl('Document') . ": " . basename($drow['url']) . ' (' . xl_document_category($drow['name']) . ')', ENT_NOQUOTES) .
     "</td>\n";
 
-  // skip billing and insurance columns
-  if (!$GLOBALS['athletic_team']) {
```
tmccormi: is this possibly backward?  Is says "IF NOT athletic team"

interface/patient_file/history/encounters.php
```diff
@@ -152,14 +152,6 @@ function showDocument(&$drow) {
...
-    echo "<td colspan=5>&nbsp;</td>\n";
```
tmccormi: That is: shouldn't this line be left in?

interface/patient_file/summary/add_edit_issue.php
```diff
@@ -702,55 +686,6 @@ function divclick(cb, divid) {
...
-  <td valign='top' nowrap><b><?php echo xlt('Grade of Injury'); ?>:</b></td>
-  <td>
-<?php
-echo generate_select_list('form_injury_grade', 'injury_grade', $irow['injury_grade'], '');
-?>
-  </td>
- </tr>
-
- <tr<?php if (! $GLOBALS['athletic_team']) echo " style='display:none;'"; ?> id='row_injury_part'>
-  <td valign='top' nowrap><b><?php echo xlt('Injured Body Part'); ?>:</b></td>
```
tmccormi: Same here ...  this code should be here if NOT atheltic_team

interface/patient_file/summary/add_edit_issue.php
```diff
@@ -788,20 +723,6 @@ function divclick(cb, divid) {
   </td>
  </tr>
 
- <tr<?php if (! $GLOBALS['athletic_team']) echo " style='display:none;'"; ?> id='row_returndate'>
```
tmccormi: Here too

teryhill:
Yep added back both lines

interface/patient_file/summary/add_edit_issue.php
teryhill: i think the !globals is for the style . the Injured body part is the athletic code

interface/patient_file/history/encounters.php
teryhill: this one I fixed

aethelwulffe:
Merge it.  I wanna delete more stuff.

interface/patient_file/summary/add_edit_issue.php
```diff
@@ -723,6 +723,15 @@ function divclick(cb, divid) {
...
 
+ <tr id='row_returndate'>
+   <td>
+   <input type='hidden'  name='form_return' id='form_return' />
+   <input type='hidden'  name='row_reinjury_id' id='row_reinjury_id' />
+  <img  
+    id='img_return'/>
+  </td>
+ </tr>
+
```
teryhill: this is a fix that was put in by someone and it works but it will need to be investigated down the road

teryhill:
so can I add annotate diagram? or are we only in delete mode 8^)

Issue #61, opened 2016-08-05
includes directory is useless


aethelwulffe:
include_once(dirname(__FILE__) . "/../../includes/config.php");
This contains:
<?php
// Copyright (C) 2010 Rod Roark <rod@sunsetsystems.com>
//
// This program is free software; you can redistribute it and/or
// modify it under the terms of the GNU General Public License
// as published by the Free Software Foundation; either version 2
// of the License, or (at your option) any later version.

// $GLOBALS['OE_SITE_DIR'] is set in globals.php.
require_once $GLOBALS['OE_SITE_DIR'] . "/config.php";
?>
This is stupid.  Either we use this directory or get rid of it.  Which one?  Please Vote.

tmccormi:
That is left over from legacy.   Needs to be refactored.  is config.php actually used for anything anymore

aethelwulffe:
This was dropped with #63

Issue #48, opened 2016-08-04
Clean up application root directory.


aethelwulffe:
We have lots of messy stuff in the application root directory.  I would like to see that, as a start, get moved somewhere appropriate.

teryhill:
What about cleaning up all the js, jquery stuff with all the different versions

aethelwulffe:
I figure hitting it in small pieces is a start.  Keeps the "issues" 
manageable.  Yes, the .js is a major need, and should be gone after.  
Should that mean that all .js should go in one pile, or just the min.js 
standard libraries?

PR #63, opened 2016-08-05
get rid of legacy includes dir

aethelwulffe:
#48 #61


library/freeb/xmlrpc_server.php
```diff
@@ -2,9 +2,7 @@
 require_once("xmlrpc.inc");
```
tmccormi: anything using freeb should be removed entirely.   There might be one place in billing the saves files to that dir the need refactorung, but freeb XMLRPC has not be used since version 2.8

PR #70, opened 2016-08-09
Issue 5 dropphpmyadmin

tmccormi:
This drops phymyadmin but allows the user to install that or any other SQL admin tool and link it into the Admin->Other->Database menu item (or disable the link altogether)

Includes a document on how to install stand alone phpmyadmin and globals defaults to that.


tmccormi:
I can squash this commit before merging  ...

PR #71, opened 2016-08-09
Drop PHPMyAdmin, add configurable SQL Admin, fix libreehr.org links

tmccormi:
Much better commit for #5


tmccormi:
Has anyone had a change to look this over?

aethelwulffe:
Everything is working ...or not working as it may be....just fine.  Did we already remove the PMA bookmarks from database.sql?

tmccormi:
Don't know about PMA books marks but I'll take a look real quick

tmccormi:
no, but I'll do that now.

PR #72, opened 2016-08-09
Drop PHPMyAdmin, add configurable SQL Admin, fix libreehr.org links

tmccormi:
Updated pull to include dropping PMA books, but keeping the sql in case it's needed


aethelwulffe:
Adding those tables are part of the installation process for phpmyadmin, and are not necessarily compatible between versions.

tmccormi:
There where some OpenEMR specifics in the PMA tables, so I just retained them as a separate file for now and removed them from database.sql

PR #75, opened 2016-08-11
Remove base Zend Framework - issue 6

tmccormi:
This should safely remove the embedded installation of ZAF2 and allow the separate install out side OpenEMR.  config/init_autoload.php will have to be edited.


aethelwulffe:
Good here, but where is this autoload of which you speak?  Lots of composer stuff has the "autoload" string in it, but I am not finding the /config/init_autoload.php anywhere.

aethelwulffe:
Oh, I get it.  As a package, the zend stuff must be modified.   Then double my upvote.  Terry, give this your OK, and let's cut down the dang package. :racehorse:

PR #83, opened 2016-08-12
Replace all date DEFAULTS with NULL instead of 0000-00-00 and other S…

tmccormi:
STRICT MODE changes
Added lname, fname and pid to patient seach list

### DIFF
--- a/library/options.inc.php
+++ b/library/options.inc.php
@@ -2390,9 +2390,9 @@ function display_layout_rows($formtype, $result1, $result2='') {
 function display_layout_tabs($formtype, $result1, $result2='') {
   global $item_count, $cell_count, $last_group, $CPR;
 
-  $fres = sqlStatement("SELECT distinct group_name,seq FROM layout_options " .
+  $fres = sqlStatement("SELECT distinct group_name FROM layout_options " .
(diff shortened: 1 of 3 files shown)


aethelwulffe:
I still can't test this at the moment...at least not in php7.  One question though:  Do we actually need a default for these values?

tmccormi:
I'm not sure it matters if there is a default or not, but one thing is sure you can't write a '' (empty string) to a date/time field in STRICT mode.  It will only accept NULL or valid dates.

That might make this branch a show stopper and cause me to agree to just changing the my.cnf to turn of STRICT mode

```
ERROR: insert failed: INSERT INTO form_encounter SET date = '2016-08-15', onset_date = '', reason = 'test encounter', facility = 'Your Clinic Name Here', pc_catid = '10', facility_id = '3', billing_facility = '3', sensitivity = 'normal', referral_source = '', pid = '15168', encounter = '2', provider_id = '1'

Error: Incorrect datetime value: '' for column 'onset_date' at row 1
```

tmccormi:
I am convinced that we can't do this.  Risk is too high.  Need to document my.cnf config must be strict mode off.

Issue #99, opened 2016-08-16
new sql err: Clicking the 'Issues' link at top of pt summary screen


htuck:
Clicking the 'Issues' link at top of pt summary screen - top portion of panel has error; bottom portion has headings for display but no other issue types listed.
	Add
	Medical Problems
		ERROR: query failed: SELECT `date` FROM `lists_touch` WHERE pid=? AND type=?
		Error: Table 'libreehr_socrates.lists_touch' doesn't exist
		/opt/www/vhosts/gponline-emr.vu2vu.com/emr_gponline_dev/libreehr/library/lists.inc at 127:sqlQuery
		/opt/www/vhosts/gponline-emr.vu2vu.com/emr_gponline_dev/libreehr/interface/patient_file/summary/stats_full.php at 170:getListTouch(14364,medical_problem)

(header columns:)
	Title 	Begin 	End 	Coding (click for education) 	Status 	Occurrence 	Referred By 	Modify Date 	Comments 	Enc
![issuelink](https://cloud.githubusercontent.com/assets/8339489/17718320/13370f2e-63c8-11e6-963b-7a4ce5d47952.png)

tmccormi:
Here's the creation error.   I am sadden by the fact that setup runs without reporting these errors...

CREATE TABLE `lists_touch` (
  `pid` bigint(20) default NULL,
  `type` varchar(255) default NULL,
  `date` datetime default NULL,
  PRIMARY KEY  (`pid`,`type`)
) ENGINE=MyISAM
MySQL said: Documentation

#1171 - All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead

teryhill:
That was changed to 
DROP TABLE IF EXISTS `lists_touch`;CREATE TABLE `lists_touch` (  `pid` bigint(20) NOT NULL default '0',  `type` varchar(255) NOT NULL default '',  `date` datetime default NULL,  PRIMARY KEY  (`pid`,`type`)) ENGINE=MyISAM ;

In database.sql

tmccormi:
Interesting ... loading database.sql directly in mysql 5.7 command line has the same behavior, that is it fails to create tables with errors and does not report it..

If I set --verbose mode I get the ERRORS as shown.
```
$ grep ERROR /tmp/logfile-sql.txt 
ERROR 1171 (42000): All parts of a PRIMARY KEY must be NOT NULL; if you need NULL in a key, use UNIQUE instead

[...]

PR #102, opened 2016-08-17
Fix issues and inconsistency in database.sql - issue #99

tmccormi:
A lot of this is deleted meaningless comment line separators and

- adding DROP IF EXISTS to all CREATES
- adding quotes on all table and field identifiers
- fixing the lists_touch table defaults


tmccormi:
I think I should change all ENGINE= to InnoDB (that's a recent update in OpenEMR)

teryhill:
There were some issues with the autoincrement and incrementing of the tables under InnoDB Be wery wery karful

tmccormi:
I made the INNODB changes.   There are some notes about auto-increment changes in the DB ...  not many and none seemed relevant to me.   Specifics, perhaps?   I vaguely remember the discussion on SF

teryhill:
interface/billing/edit_payment.php for one

 				 {
 -				  sqlStatement("insert into ar_activity set "    .
 +					 sqlBeginTrans();
 +					 $sequence_no = sqlQuery( "SELECT IFNULL(MAX(sequence_no),0) + 1 AS increment FROM ar_activity WHERE pid = ? AND encounter = ?", array(trim(formData("HiddenPId$CountRow"   )), trim(formData("HiddenEncounter$CountRow"   ))));
 +					 sqlStatement("insert into ar_activity set "    .

aethelwulffe:
I guess I need to take a live database, dump the inserts, bring in the sites directory, then try to process era's against this to be able to test the innoDB changed.  That is where the biggest issues were when this was last tried.  @teryhill I don't see that MAX value query.  Is that something out of the OpenEMR codebase?

tmccormi:
Yes those MAX value changes are part of recent OpenEMR.  No reason we can't make just the effected files MyISAM instead of INNODB, mixed ENGINE types are supported

aethelwulffe:
Selecting MAX value for autoincrement is not a great idea.
```
UPDATE table_1 
SET marked_by_connection_id = CONNECTION_ID(), 
    marked_time = NOW()
WHERE worktime < UNIX_TIMESTAMP() AND marked_by_connection_id IS NULL;
```
The issue with the above is that if the session dies, you gotta have a cron job to clean up.
Using another table to store a value, then only allowing new inserts from the "temporary" table, provided that they are also marked by time, and if used, and are afterwards deleted after the row is pushed to the final table is a pretty dependable process, but you have to make sure that the whole application talks to the ar_session table the same way.
Soooo...I guess I have said nothing.  Locked tables are a pain. all that.  What is the downside to having mixed engine and engine defaults in the table creation?  Also, are we ensuring that default collation is UTF8 general?

tmccormi:
I am no SQL expert .... here's what I think.
1) Agree that using MAX sucks, autoincrement is better, fixing the DB to be INNODB compliant by NOT using auto-increment field in compound primary keys (if I understand the problem)  @yehster does...
2) Transactions ...  we could allow MySQL to have transaction commit and roll back.
3) Mixed engines seem to be fine, I've got lots of sites using it that way.

aethelwulffe:
I hate having an incongruous foundation, but yeah, I think we need to use both engines...in a controlled way.

tmccormi:
It risk avoidance ...  OK for now so we can get to RELEASE which is nearly a month late...

tmccormi:
Any feedback here? I would like to get this merged

teryhill:
Pull request 130 is where the changes for the max were introduced, I would think leaving the tables that they used Max on as MyISAM would be prudent

tmccormi:
This is what I find ...  Look right?

1. ar_activity
2. ar_session
3. claims
4. procedure_order_code
5. procedure_answers

teryhill:
It should be any tables that uses the compound key and I think there were just a few files. So that looks right, I would question if billing should be included but I am not sure what the key is for it . I am working on statement changes.

tmccormi:
billing did not show up with a MAX() function anywhere in that pull 130 so I have pushed these for now.

aethelwulffe:
I will test tonight when I get back.

tmccormi:
Did you delete my branch instead of merging it? @aethelwulffe

aethelwulffe:
Didn't intend to do either one.  Please re-publish!  Mia Coupa Mia Coupa!

aethelwulffe:
Up arrow autocomplete is bad.  Switching back to github windows application.... :octocat:

aethelwulffe:
Merged, and testing again (php5.7).

tmccormi:
Thanks!

Issue #108, opened 2016-08-24
SQL upgrade fails


tmccormi:
Even if there is nothing to do, it should not fail...

**RENAME `openemr_module_vars` TO `libreehr_module_vars`
ERROR: query failed: RENAME `openemr_module_vars` TO `libreehr_module_vars`

Error: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '`openemr_module_vars` TO `libreehr_module_vars`' at line 1**

/opt/www/vhosts/gponline-emr.vu2vu.com/emr_gponline_dev/libreehr/library/sql_upgrade_fx.php at 504:sqlStatement
/opt/www/vhosts/gponline-emr.vu2vu.com/emr_gponline_dev/libreehr/sql_upgrade.php at 55:upgradeFromSqlFile(0_0_1-to-1_0_0_upgrade.sql)

aethelwulffe:
Here...I am wrapping it in them tagses #EndIf

aethelwulffe:
We have a merge here, is this issue closed?

tmccormi:
I have not tested it yet for final, yet

Tony

-- 
Please be aware that e-mail communication can be intercepted in 
transmission or misdirected. Please consider communicating any sensitive 
information by telephone.  The information contained in this message may be 
privileged and confidential. If you are NOT the intended recipient, please 
notify the sender immediately with a copy to hipaa-security@mrsb-ltd.com and 
destroy this message.

aethelwulffe:
I have tested this a multitude of times.  Closing issue

PR #109, opened 2016-08-24
#108 possible fix

aethelwulffe:
I have not had the existing code fail, but...I ain't using the cutting
edge of the butter knife.  Try these wrapper tags.

### DIFF
--- a/sql/0_0_1-to-1_0_0_upgrade.sql
+++ b/sql/0_0_1-to-1_0_0_upgrade.sql
@@ -71,6 +71,38 @@
 --  #IfNotListOccupation
 --    Custom function for creating Occupation List
 
+#IfTable
+RENAME `openemr_module_vars` TO `libreehr_module_vars`;
+#EndIf
+
+#IfTable openemr_modules
+RENAME `openemr_modules` TO `libreehr_modules`;
+#EndIf
+
+#IfTable openemr_postcalendar_categories
+RENAME `openemr_postcalendar_categories` TO `libreehr_postcalendar_categories`;
+#EndIf
+
+#IfTable openemr_postcalendar_events
+RENAME `openemr_postcalendar_events` TO `libreehr_postcalendar_events`;
+#EndIf
+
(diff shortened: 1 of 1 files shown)


sql/0_0_1-to-1_0_0_upgrade.sql
```diff
@@ -1624,14 +1624,38 @@ INSERT INTO list_options (list_id,option_id,title) VALUES ('lists','issue_subtyp
...
+#IfTable openemr_postcalendar_topics
 RENAME `openemr_postcalendar_topics` TO `libreehr_postcalendar_topics`;
+#EndIf
+
+#IfTable openemr_session_info
 RENAME `openemr_session_info` TO `libreehr_session_info`;
+#EndIf
+
+#IfTable openemr_module_vars
 RENAME `openemr_module_vars` TO `libreehr_module_vars`;
```
aethelwulffe: Oops.  Right, move up high...  Going for try 3

sql/0_0_1-to-1_0_0_upgrade.sql
```diff
@@ -71,6 +71,38 @@
 --  #IfNotListOccupation
 --    Custom function for creating Occupation List
 
+#IfTable
+RENAME `openemr_module_vars` TO `libreehr_module_vars`;
```
tmccormi: Doesn;t the syntax require the word TABLE?   As in
RENAME TABLE `openemr_module_vars` TO `libreehr_module_vars`;

tmccormi: http://dev.mysql.com/doc/refman/5.7/en/rename-table.html

aethelwulffe:
OK, this "jump on something then push then jump on something else then..." stuff isn't working.  I think I may turn off my phone.  Try4 seems to actually be complete.  I don't think there are any collisions with the rename at the head of the file now.  SED (or CE as it is on my machine) can be used for great good or great harm....

PR #111, opened 2016-08-25
Statement changes

teryhill:
Statement changes #10


aethelwulffe:
Any chance of removing the "sl_" from names so that we recognize it as cleaned up?

aethelwulffe:
I thought I had some trouble with other people's documentation viewable 
in the portal...but I logged out and logged in again and...it wasn't 
there.   Hmm....

PR #115, opened 2016-08-25
removed WS and more accounting calls

teryhill:
removed WS and more accounting calls #110 more yunk removed


tmccormi:
This commit is a bit mixed, removing CVSDemo install as well as WS cleanup.   Just think it better if we try not to mix tasks too much in the future.  will accept as is and adjust the commit message

PR #116, opened 2016-08-25
remove stuff needed for PHP version 5.2 and below

teryhill:
remove stuff needed for PHP version 5.2 and below #110


tmccormi:
Code looks OK.   Will pull and test this before merging, I think

aethelwulffe:
I will pull immediately after merging.

PR #129, opened 2016-08-26
removing offsite portal

teryhill:
removing offsite portal Need to look at the care coordination module for removal also. 

Have not tested this yet  #110


aethelwulffe:
That was fast.  Yay!

tmccormi:
RIp Torn!

teryhill:
I will start on CMS after a nap

PR #132, opened 2016-08-26
remove CMS Portal

teryhill:
remove CMS Portal #110 Clean up of CMS stuff. Will need to do another pass after testing. I have not tested this yet.


tmccormi:
Looks good from a code perspective.  Went all the way into the languages ... above and beyond :-)

aethelwulffe:
Tested last night (new install).  I think you might have actually gotten everything.

Issue #133, opened 2016-08-27
Change to Flow board originally by MDsupport


teryhill:
Change to Flow board originally by MDsupport

teryhill:
Keep this around for a while

aethelwulffe:
Wanna add the close button to this issue?

teryhill:
Got one more thing I want to send up having to do with the flow board. (I think it will be well received)

aethelwulffe:
Oh Goodygoodygoody!

teryhill:
Taking longer that expected closing this will open another one later for the enhancements

PR #134, opened 2016-08-27
moves provider to the top of screen if single provider

teryhill:
moves provider to the top of screen if single provider From MdSupport #133


tmccormi:
Are we ready to merge this now?

teryhill:
@aethelwulffe did you get a chance to test this? It works fine on my windows setup

teryhill:
When this is commited do not close the issue

PR #139, opened 2016-08-31
Support forum Link

aethelwulffe:
Added https://forums.librehealth.io/c/7-support ref issue #81


teryhill:
What do we need to start using as the one for our main site and the wiki?

aethelwulffe:
Tony?  Can we go for a temp that redirects to the real one later?

PR #141, opened 2016-08-31
removal of camos form ref #110 and #7 

aethelwulffe:
If it is desired to remove the CAMOS form, here it is!  Voted in the Milestone 1 discussion as one of the forms least likely to succeed.


tmccormi:
I think we should not waste time on the languages stuff.  They are generated files from a a keyword source list managed by Brady in google docs.  And leaving words that will never be used does not break anything.

Issue #66, opened 2016-08-06
Progress note forms need major overhaul


aethelwulffe:
1990 called, and they want their progress notes back.

teryhill:
The portal work that is bring done will address some of that. We use pnotes for the secure messaging.

aethelwulffe:
Closing this.  There is a merge against it, and it is a very well known need that is not linked to a specific project at the moment.

PR #149, opened 2016-09-02
Encounter form rename

aethelwulffe:
Ref issue #66 Renaming from 'newpatient' to 'patient_encounter' to get rid of some obtuseness in a core feature.  This may be met with great love, or questions of "why?", but I am throwing it out there.
  Testing has not begun, though I have done these same basic changes to another repo in the past.  Of course, this ...unfortunately...breaks some compatibility with other forks.  I have included the libreehr conversion queries though.


aethelwulffe:
Tested install and function of encounter related thingses.  Everything seems to work.
The installer script starts life by shoving a ton of notices to the log...and I had to increase the php timeout again and start over (did 600 seconds), but no "newpatient" related errors.
Moving on to test the upgrade script.

tmccormi:
This looks fine to me.   @teryhill have you had a chance to look it over?

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
```diff
@@ -43,9 +43,9 @@ public function validcredential($credentials){
 			$pres = $appTable->zQuery("SELECT * FROM patient_access_offsite WHERE portal_username=?",array($credentials[6]));
             $prow = $pres->current();
             
-			$newpatient_to_approve = 0;
+			$patient_encounter_to_approve = 0;
 			if(!$prow['portal_pwd']){
```
teryhill: These are actually new patients to approve from the portal

sql/libreehr_upgrade.sql
```diff
@@ -7,3 +7,5 @@ RENAME TABLE `openemr_postcalendar_limits` TO `libreehr_postcalendar_limits`;
 RENAME TABLE `openemr_postcalendar_topics` TO `libreehr_postcalendar_topics`;
 RENAME TABLE `openemr_session_info` TO `libreehr_session_info`;
 RENAME TABLE `openemr_module_vars` TO `libreehr_module_vars`;
+UPDATE `forms` SET `form_name` = 'Patient Encounter' WHERE `form_name` = 'New Patient Encounter';
+UPDATE `registry` SET `name` = 'Patient Encounter', `directory` = 'patient_encounter' WHERE `directory` = 'newpatient';
```
teryhill: Will this take care of the existing data?

tmccormi: Why are we bothering with "official" upgrade code?   We should be focused on NEW Install only.  We have no users that need upgrades.   Having something unofficial in your tool kit for reusing test data is fine, but don't make it part of the release

aethelwulffe:
Well, one, it does do the upgrade thingie...and without that, I need to re-write my test data generator a bit etc...
  Can we keep the libreehr upgrade bit for a little while?  We can toss it out of the release if you want...

aethelwulffe:
Personally, I would like...for short term...to be able to convert the friends and family side of all of it at least.

interface/modules/zend_modules/module/Carecoordination/src/Carecoordination/Model/EncounterccdadispatchTable.php
aethelwulffe: oops.  I caught some of that...but I guess not all.

Correcting now.

sql/libreehr_upgrade.sql
aethelwulffe: Move to contribs?

aethelwulffe:
Do I need to pull out the upgrade script portion of this to get it merged?

tmccormi:
I'll merge it but we need to think harder about what we are doing for this release.  It seems like cruft to me.

aethelwulffe:
O...K...  Had to look up cruft.

PR #151, opened 2016-09-05
Added appointment hooks, mod to menu system to dynamically add items …

kchapple:
…from multiple plugins, mod to user-group to dynamically add fields using user settings

### DIFF
--- a/library/pluginsystem/PluginSystem.php
+++ b/library/pluginsystem/PluginSystem.php
@@ -76,7 +76,7 @@ class PluginSystem
         $composerJsonPath = realpath( __DIR__."/../../composer.json" );
         $composerJson = file_get_contents( $composerJsonPath );
         if ( $composerJson === false ) {
-            throw new Exception( "$composerJsonPath/composer.json is not readable" );
+            throw new \Exception( "$composerJsonPath/composer.json is not readable" );
         }
         $composer = json_decode( $composerJson, true );
 
--- a/library/pluginsystem/PluginSystem.php
+++ b/library/pluginsystem/PluginSystem.php
@@ -76,7 +76,7 @@ class PluginSystem
         $composerJsonPath = realpath( __DIR__."/../../composer.json" );
         $composerJson = file_get_contents( $composerJsonPath );
         if ( $composerJson === false ) {
-            throw new \Exception( "$composerJsonPath/composer.json is not readable" );
+            throw new \Exception( "$composerJsonPath is not readable" );
         }
         $composer = json_decode( $composerJson, true );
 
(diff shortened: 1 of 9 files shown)


library/pluginsystem/PluginSystem.php
```diff
@@ -76,7 +76,7 @@ public function init()
         $composerJsonPath = realpath( __DIR__."/../../composer.json" );
         $composerJson = file_get_contents( $composerJsonPath );
         if ( $composerJson === false ) {
-            throw new Exception( "$composerJsonPath/composer.json is not readable" );
+            throw new \Exception( "$composerJsonPath/composer.json is not readable" );
```
aethelwulffe: EDIT:  I thought this was a goofup, until I realized that PHP must have stupidly decided to use a single backslash as the operator for a namespace....

aethelwulffe:
I would vote this a thumbs-up...but I have no idea what this is intended to do.

library/pluginsystem/PluginSystem.php
kchapple: Global namespace. 

Sent from my iPhone

> On Sep 5, 2016, at 12:38 PM, Art Eaton <notifications@github.com> wrote:
> 
> In library/pluginsystem/PluginSystem.php:
> 
> > @@ -76,7 +76,7 @@ public function init()
> >          $composerJsonPath = realpath( __DIR__."/../../composer.json" );
> >          $composerJson = file_get_contents( $composerJsonPath );
> >          if ( $composerJson === false ) {
> > -            throw new Exception( "$composerJsonPath/composer.json is not readable" );
> > +            throw new \Exception( "$composerJsonPath/composer.json is not readable" );
> EDIT: I thought this was a goofup, until I realized that PHP must have stupidly decided to use a single backslash as the operator for a namespace....
> 
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

-- 
Please be aware that e-mail communication can be intercepted in 
transmission or misdirected. Please consider communicating any sensitive 
information by telephone.  The information contained in this message may be 
privileged and confidential. If you are NOT the intended recipient, please 
notify the sender immediately with a copy to hipaa-security@mrsb-ltd.com and 
destroy this message.

aethelwulffe: Yes, a distinction with a difference I suppose.  Still scary when if dawned on me.  Scope and namespace have always been interchangeable terms with me, and I have mental issues with scope in web scripting.  IF loops don't seem to contain a local variable to it's scope.  Globals are not accessible in child functions, and now there is a php namespace without fine-grain scope control...all very weird.  I will shut up now.

Does this tie into the fancy appointment stuff going on elsewhere?

kchapple: Yes. The plugin that adds the video conferencing functionality is on my github right now. https://github.com/kchapple/vidyo

samlikins: Ken,

Though it's not a critical or even a issue, the error is wrong.

This:

>            throw new \Exception( "$composerJsonPath/composer.json is not
readable" );

Should be:

>            throw new \Exception( "$composerJsonPath is not readable" );

The filename is already in the string.
Sam

PR #152, opened 2016-09-06
Added hook for after appointment

kchapple:


aethelwulffe:
Function already exists, just no hook?

kchapple:
@aethelwulffe Somehow this got missed in the last commit. This is to add custom functionality after the appointment list on demographics page. In my vidyo plugin, I add a "Join" button to join the video conference.

PR #154, opened 2016-09-06
Claim status codes

aethelwulffe:
#10  Billing remarks, claim status, and claim status code categories updated or added


aethelwulffe:
Currently the eob_process file looks at a sparse handful of codes to determine if the record is a denial or not, then colors it red.  Sometimes it does useful stuff like stating "follow up manually".  The categories can help drive more intelligent processing of the claims instead of depending on some hard-coded logic.
  This, of course, is integral to all the claims clearinghouse work at a very low level.  Since that is such a huge, monumental advance for LibreEHR over all other known billing systems, I am sure lots of people will want to integrate this stuff into their forks pretty quickly so that they can stay on top...before they get left behind.

aethelwulffe:
Adding a lot of flavors of stuff, and doing some clean-up formatting.

tmccormi:
Here's a trick... adding ?w=1 to the URL of the code review turns on ignore whitespace changes
https://github.com/LibreEHR/LibreEHR/pull/154/files?w=1

aethelwulffe:
https://github.com/tiimgreen/github-cheat-sheet

aethelwulffe:
For chrome:

https://chrome.google.com/webstore/detail/github-diff-whitespace/lhbcdehjihmbiafeodkfnbndleijnnhp?hl=en-US

That gives you a button to ignore whitespace.

Firefox developer has that all on it's lonesome.

PR #155, opened 2016-09-07
remark_codes.php array syntax errors #10 

aethelwulffe:
Apparently, I had not tested this code.


aethelwulffe:
Now this is real-world tested!

Issue #68, opened 2016-08-08
Remove adLDAP


tmccormi:
This was offered on OpenEMR..  another good thing to clean out
remove adLDAP from library
D library/adldap/README.txt (9)
D library/adldap/adLDAP.php (725)
D library/adldap/adLDAP_auth.inc (261)
D library/adldap/adLDAP_conf.inc (24)
D library/adldap/adLDAP_sync.php (182)
D library/adldap/adLDAP_sync.sh (9)

aethelwulffe:
Note that there is a lot of total junk in adodb.
files like "chris.php" and others that start with localhost db connections, and the string 'C0yote71' in them are next on the chopping block.  What do we actually need out of the adodb directory?

tmccormi:
we could ditch adodb and use the current standard libraries instead and use PDO

aethelwulffe:
Please!

for cr@p's sake, please give the d@mn project PDO.  Pretty. Freaking. 
Please.

How many times do I hear "Well, if we were just using PDO...."

aethelwulffe:
http://tonylandis.com/php/adodb-pdo-convert-refactor-code/
This is what I have had bookmarked since the day it was posted.
Somewhat hackish, yes, but perfectly legitimate.

tmccormi:
looks like a good start.  ALL sql access should be going through our own wrapper in order for logging to work, so it should not be toooooo hard tod directly convert the calls to PDO

PR #158, opened 2016-09-09
library/adldap and adodb fluff #68

aethelwulffe:
First and final round of ldap removal, round 1 of adodb.  adldap is junk, but some small part of adodb requires retention and upgrade since 2008.
#110


tmccormi:
Anyone had a chance to test this?

tmccormi:
Looks like all deletes.... so that's OK, testing?

aethelwulffe:
I ran it when testing.  Nothing hinges on the ldap stuff.  I did NOT do an install with it

PR #159, opened 2016-09-09
Allow External URLS that open as separate windows, add specific code …

tmccormi:
…for sql_admin_tool_url

OK this works in FireFox.  Chrome still blocks it on XFrame = DENY, even though I have the configuration set to allow (which you can see when I run via firefox ...

$cfg['AllowThirdPartyFraming'] = true; # Enable Third Party Framing (X-Frame-Options)

### DIFF
--- a/interface/main/tabs/js/tabs_view_model.js
+++ b/interface/main/tabs/js/tabs_view_model.js
@@ -203,9 +203,16 @@ function menuActionClick(data,evt)
                 alert("This encounter is locked. No new forms can be added.");
                 return;
             }
-        }        
-        navigateTab(webroot_url+data.url(),data.target);
-        activateTabByName(data.target,true);        
+        }
+        if (data.url().indexOf("http")===0)
+        {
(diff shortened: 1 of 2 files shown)


aethelwulffe:
I still ain't got Chrome (need a throw-away VM), but I will test this now.  Of course, the other versions worked on FireFox anyway, and I don't expect any bad news.

aethelwulffe:
in globals.inc.php we also have:
line
 80 $sql_admin_tool_url  = 'http://localhost/phpmyadmin'; //not my real URL!
1890    'sql_admin_tool_url' => array(
      xl('URL path to SQL Admin Tool'),
      'text',                         // data type
      $sql_admin_tool_url,            // default set at top
        xl('URL path to SQL Admin Tool such as PHPMyAdmin or MySQL WorkBench'),
    ),
 where was that committed?

aethelwulffe:
Looks like it works.

PR #164, opened 2016-09-10
Rename Patients menu item to Finder, add option to globals

tmccormi:
regarding #163

### DIFF
--- a/interface/main/left_nav.php
+++ b/interface/main/left_nav.php
@@ -980,8 +980,8 @@ $(document).ready(function(){
 
   <li class="open"><a class="expanded" id="patimg" ><span><?php xl('Patient/Client','e') ?></span></a>
     <ul>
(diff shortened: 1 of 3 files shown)


aethelwulffe:
I could say maybe Create, Add, Create Patient, Create Chart, Add Patient, Add Patient/Search, New Patient or something else like that, but "Finder" is fine if we make it a default locked tab for someone so that they get used to what that is, 'cause if it is in the Patient.  Patient Finder as a title is a good, but when "Patient" is already the menu item, the context is clear.

So, do we have:
 interface/new/new_patient.php
interface/main/finder/dynamic_finder.php
interface/main/finder/patient_select.php
...all covered?

  The CDR and the AMC (and of course PQRS!) all use the patient_select.

aethelwulffe:
..in any case, I say merge immediately.  I like cruft like this.

PR #165, opened 2016-09-10
Change Facility to Place of Service

aethelwulffe:
Billing Facility and Facility are really poorly understood by users.
This naming convention is responsible for a lot of confusion and billing
failures. 
The only quandary is what to do to ensure that the Place of Service is the Location Code source.  I have seen this one tiny change help several businesses get things together.


aethelwulffe:
Thank you.

PR #169, opened 2016-09-14
Added a bit about plugins and hooks to the readme file

kchapple:


tmccormi:
Thanks for the info.   

I would not have put this in the top level Readme.md file.  Should be part of the documentation dir in a Plugin-Dev-Process .md  I think.

Easy enough to move.  I'll take care if that.

tmccormi:
Moved this to #171

PR #171, opened 2016-09-14
Update Licenses and Add PLugin FAQ

tmccormi:


aethelwulffe:
OK, read it all.
Yes, thank you for moving stuff to /documentation.
I would recommend using a file extension with any readme files.
RTF files are really nice for use with installers and the like.
Top level man pages (err. readme) are good places for a directory of other documentation locations and subjects, kind of like a table of contents.

COPYRIGHT_AND_LICENSE
```diff
@@ -1,7 +1,6 @@
 LICENSE
-Unless otherwise stated, this code is subject to the terms of the GNU General
-Public License version 3 or higher (GPL), a copy of which is included here for
-your convenience at license.txt .
+Unless otherwise stated, this code is subject to the terms of the Mozilla Public License v 2.0 a copy of which is included here for
+your convenience at mpl_license.txt
```
teryhill: May be a good Idea to add the HealthCare disclaimer here also?

tmccormi: I think it's fine where you put it.

tmccormi:
rename the COPYRIGHT_AND_LICENSE file with .txt for @aethelwulffe

PR #172, opened 2016-09-14
Move menu JSON to sites dir as a nicely formatted file and add templa…

tmccormi:
and add template variable replacement


tmccormi:
fixed conflicts and merged

Issue #79, opened 2016-08-11
More cleanup the acknowledgment page needs to be condensed


teryhill:
Could put a general acknowledgment to the OpenEMR Community with our Thanks ... and leave it at that, tracking only the new contribs to the new project

teryhill:
@tony @aethelwulffe File is there . This issue can probably be closed after the merge

PR #173, opened 2016-09-14
changes to add health disclaimer

teryhill:
changes to add health disclaimer in conjunction with #79

### DIFF
--- a/acknowledge_license_cert.html
+++ b/acknowledge_license_cert.html
@@ -12,7 +12,9 @@
 <h3>License information</h3>  
 <div class="paragraph">   
 Unless otherwise stated, this code is subject to the terms of the <a href="license.txt">
-GNU General Public License version 3 or higher (GPL)</a>, a copy of which is included here for your convenience.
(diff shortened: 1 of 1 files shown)


acknowledge_license_cert.html
```diff
@@ -12,7 +12,9 @@
 <h3>License information</h3>  
 <div class="paragraph">   
 Unless otherwise stated, this code is subject to the terms of the <a href="license.txt">
-GNU General Public License version 3 or higher (GPL)</a>, a copy of which is included here for your convenience.
+GNU General Public License version 3 or higher (GPL)</a>, a copy of which is included here for your convenience.<br>
```
tmccormi: Need to change this to MPL

teryhill: see if that works. I am assuming from your post it is called mpl_license.txt instead of license.txt

tmccormi:
looks good.

teryhill:
Keep this open until I hear back from Larry. I will have some changes to the license statements

Issue #186, opened 2016-09-18
sl_eob_search error


aethelwulffe:
sl_eob_search.php
line 88 Menu item "Posting"
breaks here:
global $STMT_TEMP_FILE_PDF;
$pdf =& new Cezpdf('LETTER');//pdf creation starts <breaks here line 88
Test with FEES>POSTING

aethelwulffe:
This is the last bug that does not have a commit against it.

teryhill:
What is the error? Is the PDF library's still there?

aethelwulffe:
Yes, the file is still there.
  Error is related to "NEW".
[Mon Sep 19 13:01:36.985089 2016] [:error] [pid 7120:tid 1792] [client ::1:65119] PHP Parse error:  syntax error, unexpected 'new' (T_NEW) in C:\\xampp\\htdocs\\LibreEHR\\interface\\billing\\sl_eob_search.php on line 88, referer: http://localhost/LibreEHR/interface/main/tabs/main.php?url=main_info.php

aethelwulffe:
In php7 the assignment of variables by reference is deprecated.  There is nothing wrong with the code, it is just meaningless to php7.

teryhill:
take out the & and see if that fixes it for PHP7

aethelwulffe:
I removed the &.
Now it works.
I have a pull in already.

aethelwulffe:
@teryhill Looking back, the deprecation warning has been popping up since PHP5 came on the scene.  The project was ignoring this stuff for a very very long time.

aethelwulffe:
I checked this out with NetBeans (which I don't use).  It doesn't give any warnings about it.

PR #188, opened 2016-09-19
PHP7 upgrade for Posting Screen #186 

aethelwulffe:
variable assignment by reference is implied


aethelwulffe:
This is a php7 thing.  Best to go looking for =& in the code base to fix any others.

aethelwulffe:
#186

Issue #190, opened 2016-09-19
Globals Navigation Menu selection can go


tmccormi:
No need to select what kind of menu to use anymore

![navigation](https://cloud.githubusercontent.com/assets/285835/18651641/59f6b5d0-7e81-11e6-9de6-de5e6749e72a.png)

teryhill:
The width is set as a default of 150 (if it is empty) in main_screen.php and used in main_title.php. I would suggest leaving it in those files until we are sure we can jerk those programs.

tmccormi:
We can just hard code it and drop the option entry

teryhill:
00104: $nav_area_width = '130';
00105: if (!empty($GLOBALS['gbl_nav_area_width'])) $nav_area_width = $GLOBALS['gbl_nav_area_width'];

teryhill:
all done?

PR #192, opened 2016-09-20
Removed Layout Style #190

teryhill:
removed layout style from globals no longer needed #190


tmccormi:
should also remove the left nav column width config.

Issue #203, opened 2016-09-21
global nav area removal broke EHR


aethelwulffe:
Reproducible.
Install LibreEHR.  Menu is messed up big-time, like every version of the navigation system we have ever had, from "original", "Tree", and Tabs all at one time.  No sliding menu though!
Reverted removal of Modules change.  No effect.  Reverted Global nav area change...fixed.
Should look at this again.
Will re-create windows installer after.
I go sleep.

aethelwulffe:
---------- Find in Files ----------
> Searching for the string '$nav_area_width'...
C:\AlphaFiles\bidenis\REPOS\LibreEHR\interface\main\main_screen.php(104,1): $nav_area_width = '130';
C:\AlphaFiles\bidenis\REPOS\LibreEHR\interface\main\main_screen.php(158,38):      <frameset cols='<?php echo attr($nav_area_width) . ',*'; ?>' id='fsbody' frameborder='1' border='4' framespacing='4'>
C:\AlphaFiles\bidenis\REPOS\LibreEHR\interface\main\main_screen.php(165,46):      <frameset cols='<?php echo  '*,' . attr($nav_area_width); ?>' id='fsbody' frameborder='1' border='4' framespacing='4'>
C:\AlphaFiles\bidenis\REPOS\LibreEHR\interface\main\main_screen.php(177,33): <frameset cols='<?php echo attr($nav_area_width); ?>,*' id='fsbody' frameborder='1' border='4' framespacing='4' onunload='imclosing()'>

More places use the variable $nav_area_width

aethelwulffe:
RETREAT!  RETREAT!

tmccormi:
Merged.  Test with Master please

aethelwulffe:
Exactly the same.
Maybe I didn't pull from the right branch or something.
Maybe that wasn't the issue.

teryhill:
Check and make sure you have the correct code. The only other thing was the other variable that was removed.

teryhill:
It is probably the other commit. I will do some checking

teryhill:
set ($GLOBALS['concurrent_layout']) to 3 in the globals.php

aethelwulffe:
Did.  No change.  Maybe that is because the user preferences are now overriding it, and I don't have an interface to change that directly.  I'll look, but I am really running out of time today.

teryhill:
I will work on it as soon as I get back

teryhill:
See if the reversal does it

[...]

PR #204, opened 2016-09-21
Undo the removal of Navigation Area Width

teryhill:
Undo the Navigation Area width #203


tmccormi:
Change ref to Master, it was set to development_101

PR #206, opened 2016-09-22
images and docs for packages/installer

aethelwulffe:
what it says.


aethelwulffe:
Yes, this is windows specific... but just for the installer package.  .MD has the same issue that extensionless files do.  The user's system has never had the file association made.  You would expect that an ubiquitous application like "notepad" would have those associated by default, but that isn't the case.  Doing a search on my machine for .MD (non-sensitive) as an extension reveals nothing that isn't in my repo.  In any case, I also want to ensure the rtf opens in a non-browser, as I don't want the information to disappear in a tab when they are trying to cross-reference.
Using anything other than a bitmap would require me to use another library in this particular instance, as well as a rebuild of the compiler.
...While RTF isn't a "baseline", the systems know what to do with them.  Open up LibreOffice and try to do a "save as" for MD.  Ain't there.  I had never heard of it before now myself.
  Sorry for the extra files (though I can delete a LOT more to make up for them), and there is no harm in just adding them before each build.  The icon files I would certainly keep (at least one of them).
  This is just the first serious run, and will certainly be tweaked a lot more.  I want to see how much I can legally rip out of the additional packages.

aethelwulffe:
Aw shucks.  The RTF is a fail anyway.  Didn't include the source for the images.  *sigh*  Maybe I will switch to an archive format like ODT or Doc, or just say heck with it and do pdf or htm.
The rtf in question is really the Quick Start Guide.  The other one is for use inside the installer window itself...and that one works fine.

Issue #209, opened 2016-09-22
Database still silently failing to install some tables


tmccormi:
This is on the new ehr.librehealth.io deployment.   The display wierdness is also evident here.
![ehr librehealth io](https://cloud.githubusercontent.com/assets/285835/18764379/04355150-80c6-11e6-8f48-c31c936fe936.png)

tmccormi:
This is a 16.04 ubuntu, mysql  Ver 14.14 Distrib 5.7.15, for Linux (x86_64) using  EditLine wrapper

tmccormi:
I'm going to dump it and do a clean install in case Robby did something wrong, but he installed from development_101, so it should have been good.

teryhill:
I am going to send a commit backing out the change for layout style

tmccormi:
Sorry, that made no difference at all on the ehr.librehealth.io site.

teryhill:
just for kicks go into globals and change the concurrent layout save and see if that changes things

aethelwulffe:
Wait a minute:  Is this issue due to tables not existing?  I can verify that in the last master I built the windows install from, that table certainly gets created.  Looking at it now.
That does not mean that the database on the demo got loaded fully/properly.  It very well may be that the time-outs and query limits caused the install to break (silent is still bad) but the database >does< have good queries.

tmccormi:
I just reinstalled and re-ran setup and the table did not get created.  
* Table 'libreehr.libreehr_postcalendar_events' doesn't exist

aethelwulffe:
Running master (As it is) tables get created.  Running from the layout options and INNODB database changes, that branch loads everything AND the menu looks right.  Still has the "modules" and "external" piece, and doesn't have that do_action stuff, but it is good otherwise.

aethelwulffe:
Just restored the dem_layout_options branch to diff off of.

tmccormi:
This is the list of tables that silently failed to be created.
> `batchcom`
> `drug_inventory`
> `drugs`
> `amendments`
> `amendments_history`
> `log_comment_encrypt`
> `patient_tracker`
> `patient_tracker_element`
> `insurance_data`
> `libreehr_postcalendar_events`

tmccormi:
I think this is a mysql version/setup difference on the new server, but needs to be figured out

teryhill:
What version is mysql?

tmccormi:
[...]

PR #210, opened 2016-09-22
Reversing the change to remove concurrent layout options

teryhill:
Reversing the change to remove concurrent layout options #209


tmccormi:
Wiping the DB and re running the setup did appear to fix the frame display issue.

PR #217, opened 2016-09-28
Add Legal Business Entity Name to Facilities

aethelwulffe:
Uses the legal business entity name in EDI, letting the users call their
facilities other names that identify the location code, such as "Office,
Home, ALF, Hospital, etc...".
  This corrects a lot of trouble, and should be added to the CMS-1500 code as well.
  The decision to use alias instead of altering the behavior of name was due to extent of it's use, and there is a reasonable argument for that in any case.

### DIFF
--- a/interface/usergroup/facility_admin.php
+++ b/interface/usergroup/facility_admin.php
@@ -139,20 +139,22 @@ function displayAlert()
          <tr>
           <td width='150px'><span class='text'><?php xl('Name','e'); ?>: </span></td>
           <td width='220px'><input type='entry' name='facility' size='20' value='<?php echo htmlspecialchars($facility['name'], ENT_QUOTES) ?>'><font class="mandatory">&nbsp;*</font></td>
+          <td width='150px'><span class='text'><?php xl('Legal Entity','e'); ?>: </span></td>
+          <td width='220px'><input type='entry' name='alias' size='20' value='<?php echo htmlspecialchars($facility['alias'], ENT_QUOTES) ?>'><font class="mandatory">&nbsp;*</font></td>
+          </tr>
+         <tr>
+         <td><span class=text><?php xl('Address','e'); ?>: </span></td><td><input type=entry size=20 name=street value="<?php echo htmlspecialchars($facility["street"], ENT_QUOTES) ?>"></td>
           <td width='200px'><span class='text'><?php xl('Phone','e'); ?> <?php xl('as','e'); ?> (000) 000-0000:</span></td>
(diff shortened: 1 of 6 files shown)


tmccormi:
Should we wait for the CMS 1500 support?

aethelwulffe:
I think that getting this in as a prereq would be a good idea.  The changes to the HCFA are more extensive.  Basically, I am supplementing the current code, then tossing it.  I will keep the original to print "batches", and use the new code in single claim mode until it has a test result.  The new system takes you to a form that displays the data on a virtual CMS1500.  The help text for each field displays what the data source for that box is.  How that affects changes in the system elsewhere is stuff i am still figuring out.

aethelwulffe:
Figured out I really need to put a LOT of other things into the system before addressing the CMS1500.  If this can be merged up as-is (since it is a perfectly workable feature) that would help ease developing on a new branch.

interface/usergroup/facility_admin.php
```diff
@@ -162,16 +164,17 @@ function displayAlert()
...
-	?>
-        <tr>
-            <td><span class=text><?php xl('State','e'); ?>: </span></td><td><input type=entry size=20 name=state value="<?php echo htmlspecialchars($facility{"state"}, ENT_QUOTES) ?>"></td>
+	?>     
             <td><span class=text><?php xl('Tax ID','e'); ?>: </span></td><td><select name=tax_id_type><option value="EI" <?php echo $ein;?>><?php xl('EIN','e'); ?></option><option value="SY" <?php echo $ssn;?>><?php xl('SSN','e'); ?></option></select><input type=entry size=11 name=federal_ein value="<?php echo htmlspecialchars($facility{"federal_ein"}, ENT_QUOTES) ?>"></td>
         </tr>
         <tr>
-            <td><span class=text><?php xl('Country','e'); ?>: </span></td><td><input type=entry size=20 name=country_code value="<?php echo htmlspecialchars($facility{"country_code"}, ENT_QUOTES) ?>"></td>
-            <td width="21"><span class=text><?php ($GLOBALS['simplified_demographics'] ? xl('Facility Code','e') : xl('Facility NPI','e')); ?>:
+        <td><span class=text><?php xl('Zip Code','e'); ?>: </span></td><td><input type=entry size=20 name=postal_code value="<?php echo htmlspecialchars($facility{"postal_code"}, ENT_QUOTES) ?>"></td>
```
tmccormi: never mind this file is really rearranged a lot, makes it harder to review

aethelwulffe:
Required re-arranging as it is in a restricted two-column format.  Otherwise, it looks like shit.

Issue #220, opened 2016-10-04
Document Categories - EDIT not working


tmccormi:
Administration->Practice->Documents->EDIT Categories

Does not present the Tree for editing.   tested on release demo site.

![edit-doc-categories-broken](https://cloud.githubusercontent.com/assets/285835/19095091/a805dc88-8a48-11e6-91fb-0f20a939e4c7.png)

tmccormi:
Treemenu not defined

![practice-document-error](https://cloud.githubusercontent.com/assets/285835/19095293/d00a2ad0-8a49-11e6-8f8a-2de655291018.png)

tmccormi:
Did not intend to close this one

tmccormi:
Anyone have any thoughts here.  Nothing obvious to me.

aethelwulffe:
Well, we know where that menu is defined, and where the libraries (that were moved I think) are.  I will go looking.

aethelwulffe:
CategoryTreeMenu.js is not in the root directory.  DocumentTreeMenu.js still is.

tmccormi:
There is not TreeMenu.js in base openemr either, only DocumentTree, I assumed it was a rename.  There is no Treemenu,js in either code base at all.

tmccormi:
did someone rename it or something?

aethelwulffe:
BPUC has both...look for blame.  I am dumping it in to test.

aethelwulffe:
Dumping in CategoryTreeMenu.js that is.

tmccormi:
There is no such thing as 
tony@dragonfire:/opt/gitrepos/LibreEHR (master)$ find . -name CategoryTree\* -print
./library/classes/CategoryTree.class.php

aethelwulffe:
Yes, that fixes it.  Pushing to a branch here in a minute.

aethelwulffe:
Want on dev or master?

tmccormi:
Dev is fine, you should be about to undelete from the repo (versus) an new copy ..
but whatever  works

aethelwulffe:
I am an idiot.  This did NOT fix this.  I totally misunderstood the issue.

aethelwulffe:
I see that something is looking for these libraries at library/js instead of the root, and they were not there.  Also, this seems to have issues:
<?php require_once($GLOBALS['srcdir'].'/js/DocumentTreeMenu.js');?>

aethelwulffe:
[...]

PR #226, opened 2016-10-05
Document category edit error

aethelwulffe:
View the document category edit template.

### DIFF
--- a/templates/document_categories/general_list.html
+++ b/templates/document_categories/general_list.html
@@ -16,7 +16,8 @@
 <script type="text/javascript">
 var deleteLabel="{xl t="Delete"}";
 </script>
--<?php require_once($GLOBALS['srcdir'].'/js/DocumentTreeMenu.js');?>
+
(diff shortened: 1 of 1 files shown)


aethelwulffe:
#220 
You can revert the last commit on this...I think...
More importantly, I need to get the js upgrade (all onto the same library) as there are some issues with jquery.1.3.2.  I replaced all that, and nothing seemed to break...but I still have testing against every file (72) that calls it.

templates/document_categories/general_list.html
```diff
@@ -16,7 +16,8 @@
 <script type="text/javascript">
 var deleteLabel="{xl t="Delete"}";
 </script>
--<?php require_once($GLOBALS['srcdir'].'/js/DocumentTreeMenu.js');?>
+
+<script type="text/javascript" src="/LibreEHR/interface/../library/js/DocumentTreeMenu.js"></script>
```
teryhill: Where is this pointing to /interface or /library or both I am confused

PR #227, opened 2016-10-05
Jquery upgrade initial pass

aethelwulffe:
Custom applictions like CKEditor and the Care module have had the path file name changed, but not the full path to the location of the upgraded jquery library.  I se this as an initial step toward centralizing libraries, and not breaking anything that is actually used.


tmccormi:
This is not a good branch.  It has commits that are already in the db, lots of them

aethelwulffe:
So much for re-basing.  I will try again.

Issue #228, opened 2016-10-06
Letter Pop up..  White Screen version 1.0.0


tmccormi:
[Thu Oct 06 23:19:40.802596 2016] [:error] [pid 20043] [client 37.228.227.191:6948] PHP Parse error:  syntax error, unexpected 'new' (T_NEW) in /opt/ehr/interface/patient_file/letter.php on line 146, referer: https://ehr.librehealth.io/interface/main/tabs/main.php?url=main_info.php

aethelwulffe:
We already knew that ezpdf has issues in this environment.

teryhill:
Sent up a fix

aethelwulffe:
Sent up additional fix across the code base.
#230

tmccormi:
This can be closed I assume

PR #229, opened 2016-10-07
Fix for PHP 7 error in letter popup

teryhill:
Fix for PHP 7 error in  letter popup  #228


aethelwulffe:
You were a few minutes ahead of my branch.  I am afraid that mine has this same fix in it.

tmccormi:
This is why you should ASSIGN the bug to yourself before working on it... so we aren;t doing redundant work :-)

Issue #236, opened 2016-10-10
Unable to add appointments on ehr.librehealth.io


tmccormi:
ERROR: insert failed: INSERT INTO libreehr_postcalendar_events ( pc_catid, pc_multiple, pc_aid, pc_pid, pc_title, pc_time, pc_hometext, pc_informant, pc_eventDate, pc_endDate, pc_duration, pc_recurrtype, pc_recurrspec, pc_startTime, pc_endTime, pc_alldayevent, pc_apptstatus, pc_prefcatid, pc_location, pc_eventstatus, pc_sharing, pc_facility,pc_billing_location,pc_room ) VALUES (?,?,?,?,?,NOW(),?,?,?,?,?,?,?,?,?,?,?,?,?,1,1,?,?,?)

Error: Incorrect integer value: '' for column 'pc_multiple' at row 1

/opt/ehr/library/encounter_events.inc.php at 332:sqlInsert
/opt/ehr/interface/main/calendar/add_edit_event.php at 123:InsertEvent(Array)
/opt/ehr/interface/main/calendar/add_edit_event.php at 584:InsertEventFull()

tmccormi:
Mode is Strict ... 

SELECT @@sql_mode;
+-------------------------------------------------------------------------------------------------------------------------------------------+
| @@sql_mode                                                                                                                                |
+-------------------------------------------------------------------------------------------------------------------------------------------+
| ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION |
+-------------------------------------------------------------------------------------------------------------------------------------------+

aethelwulffe:
No zero date kills a lot of stuff.  I think your user addition bug is also the same root issue.

tmccormi:
Yes.  however I changed the setting to permissive, restarted mysql and still can't save.

 mysql> set sql_mode="";
Query OK, 0 rows affected, 1 warning (0.00 sec)

mysql> SELECT @@sql_mode;
```
+------------+
| @@sql_mode |
+------------+
|            |
+------------+

```

aethelwulffe:
Same error attempting to save an all day availability for a provider.
[...]

PR #237, opened 2016-10-10
calendar_autoinc hack extension for INNODB

aethelwulffe:
#236 This addresses the issue of trying to add +1 to a NULL value, but might
be better served by just re-establishing autoincrement in the INNODB
table like we did for the others.

### DIFF
--- a/interface/main/calendar/add_edit_event.php
+++ b/interface/main/calendar/add_edit_event.php
@@ -91,7 +91,9 @@ function InsertEventFull()
             // obtain the next available unique key to group multiple providers around some event
             $q = sqlStatement ("SELECT MAX(pc_multiple) as max FROM libreehr_postcalendar_events");
             $max = sqlFetchArray($q);
-            $new_multiple_value = $max['max'] + 1;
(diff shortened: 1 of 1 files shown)


tmccormi:
I guess I'm OK with either.   @teryhill, your thoughts?

teryhill:
auto inc is better to me . less chance of some one accidentally messing it up , but you guys have been at it longer than me 8^)

aethelwulffe:
Sounds like a vote in a particular direction.  I'll see if I can get the DB hooked up in...damn.  Got to get back to work.  Later.

tmccormi:
I'm confused are we accepting this or switching to a DB based solution?

aethelwulffe:
No, a pure DB solution is not the answer.  We have to use the MAX+1 when dealing with multiprovider events etc... and this has been the solution since 2011.  Using autoincrement would still require you to get the id, then use it, as you have to link it to things.  This is as good of an answer as I can come up with.

tmccormi:
OK.  I'll accept that. :-)

PR #240, opened 2016-10-14
This will allow control over deletion of payments

teryhill:
This will allow control over deletion of payments. #10 This allows a point of control for the payment deletion. This wold be an automatic function of a month end routine if one existed (will soon hopefully)


teryhill:
I have it at 2 of my client sites no issues from them

PR #260, opened 2016-11-12
no stomp menu experiment

aethelwulffe:
set all menu items to unique menu id's
This pull request is for discussion, but the branch is for tinkering and possibly a merge after some review etc...

### DIFF
--- a/sites/default/menu_data.json
+++ b/sites/default/menu_data.json
@@ -131,7 +131,7 @@
     "children": [
       {
         "label": "Fee Sheet",
-        "menu_id": "cod2",
+        "menu_id": "bil0",
         "target": "enc",
         "url": "/interface/patient_file/encounter/load_form.php?formname=fee_sheet",
         "children": [],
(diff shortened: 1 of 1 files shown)


tmccormi:
I'm not seeing what I expected, still lots of things in the Reports menu that step on the previous report.  I'm going to recheck my install, I might not have gotten the code change I expected

aethelwulffe:
Well check it, and lemme know, cause maybe I left some changes out of the commit or some other craziness.

aethelwulffe:
An even more important bit is the attaching tabs to the form element (instead of aligned left), increasing the size of the tab controls (and tabs) and making the tab size configurable resizable.  After that is still the formatting of the contents of tabs.
  Right now, I wouldn't recommend the layout for regular use.

tmccormi:
I think a simple short term solution would be to allow there to be a configurable limit of the number of tabs in the foreground.  Default to 3 max.

PR #261, opened 2016-11-12
HTML5 cms1500 form

aethelwulffe:
html conversion of a cms1500 to use as the basis of an interactive
graphical form.


aethelwulffe:
having wicked issues getting time to work on this, yet getting hounded about it daily...

teryhill:
We need to evaluate the current fee sheet and see what fields we need to add to make this happen and put them in a new include file for us to start working out of to make our changes

aethelwulffe:
We have no separate repo for general "stuff" anymore.  We lost that in the change-over.  Since this isn't going into contribs, and we don't have a dev resource spot, It goes no-where.  I am going to drop this request and keep it on my machine in a nice dark hole.

teryhill:
can you zip up a copy and send it to me?

aethelwulffe:
Sure, sending your way...I think I can restore my local branch still.

Issue #264, opened 2016-11-16
Kevin's EDI version 2


teryhill:
I will be committing this in the near future

PR #272, opened 2016-11-17
Kevins edi v2

teryhill:
This is Kevins EDI changes #264 This is a large commit because of the assets that he used. I tested it on my system (windows) some one will need to test it on the other OS flavor. Looks good on the screen with 1 additional tab.


aethelwulffe:
First impressions, yes, this is big.
Second, there is a LOT of css and fonts...not that I mind fontawesome and all that, but it may be better to land some of this in a more organized fashion or place it into a separate module structure...if that is at all doable.  It would take a lot of work, I know.
We have another jquery UI and a ton of other things with directory structures delineated by version numbers.  My opinion is that dependencies like this need to have a standardized name, one copy, and to keep that copy up to date, example "jquery.js".  It seems very very messy.  If it is "dude we must have this, look at all the things it does, let's just do it" then fine, we will clean and refactor later.
  The next thing is a plain language answer for "what does this get us?"  This addresses the decision point above.  What functions does this add that will alter the average user experience?

aethelwulffe:
I cannot believe that the actual application requires even a tiny fraction of the 1400 files of fonts, css and other fluff...or even all the js libraries.

aethelwulffe:
This has a lot of stuff supporting the parsing of x12 files and history based on the files.  To start with, there is the eligibility bit.  What use is it, when the EHR can't currently create a valid 270 file to begin with?  What does using the old EDI files as a document database give us anyway?  Shouldn't the base system have the data we need?  Isn't the goal to import/export the exchange data properly, using the x12 format only as an exchange media?

teryhill:
Don't you have to start some where. I agree that this is not what we would like to have but it gets us started. I am looking at the 270 file now to work on getting it to an acceptable file by the clearing houses then the creation of the 276.

teryhill:
This will atleast allow the viewing of the 837 file in a human readable format for the loops. I have tried reaching Kevin to get an Idea of where he is headed with this and to get some Ideas of what we can expect to see down the road if anything but I have not heard from him.

I Can remove this commit

aethelwulffe:
Hey, I ain't passing judgement yet, I am just asking questions.  I think it is a piece we should keep around...somewhere...in any case.
So you know, there is a very deep php-based edi parsing tool in progress.  It's primary use is import of data from 837's, but it has classes for all the loops and elements we will encounter in other file types.  Moreover, it is far more comprehensive, much smaller, and can accommodate bastard versions of 837's in common use.  Simply viewing an 837 in human readable form is not as big of a deal as this mass of code seems to represent.  The main issue aside from the fluff is that I can already see a few errors, omissions and exceptions in the parsing of the files.

teryhill:
How close is that module?

aethelwulffe:
Really?  Extremely close.  There are tweaks to do in the back end...you 
know, you keep going along and finding something else that is just not 
*quite* right.  Once that is there, the front-end for the first use case 
can be worked on (where the fun starts) and anything like that can be 
scaled or scoped out.  The next issue is that while I am a source of a 
lot of the conceptual logic, I am not the author/designer.  That means 
that this project ain't been paid for, and I have no right to contribute 
it.  I hope that issue will be solved soon enough though, getting the 
developer well paid, getting our targeted application out of it, then 
making the whole back-end and the front-end structure available for 
end-user type applications and a way to make some sweeping changes to 
our whole era/edi structure.  Everyone wants this, since it promotes the 
project and should promote system adoption and the business thereby.  We 
are just starving and grinding our teeth over that money thing.

teryhill:
So no where near able to commit it 8^)

aethelwulffe:
Ask me next wednesday...or even after 4pm today...;-)

teryhill:
I do have a dollar in my bank account if that will help. Let me know when you want to start working on getting it in LibreEHR.

aethelwulffe:
On second thought, maybe we should merge this after the big 
Kanji/Hee-Haw translation gets merged..:-P

teryhill:
BR549

aethelwulffe:
Who was that with Opie?

PR #273, opened 2016-11-18
Returned globals init hook that got removed

kchapple:
This is the hook that allows for globals to be dynamically added by a plugin. At some point it got removed, probably by a conflict-resolving merge.


tmccormi:
Ken, this is merged into development _101  (not master) just a reminder

Issue #269, opened 2016-11-16
Changes for Misc Billing form to handle needed items


teryhill:
I will be committing this in the near future

aethelwulffe:
I am absolutely positive that I will need your assistance when diving into the interactive 1500 again.  To lighten the load, we can make sure it satisfies the current specific needs of your LSS customers.  I am pretty sure that if it satisfies Florida and NC medicaid, we will be catching most every issue possible.

aethelwulffe:
None of which sounds like I can get back to this week...

teryhill:
We may need to just talk it out a couple of times. That should let us develop an attack plan

teryhill:
This is actually the code that is currently being run on LSS 4.2.1 and 4.2.2

PR #274, opened 2016-11-19
Misc_Billing_Form Changes.

teryhill:
These changes allow the EPSDT options, The box 17 options, The
electronic resubmission of claims. #269

### DIFF
--- a/interface/forms/misc_billing_options/new.php
+++ b/interface/forms/misc_billing_options/new.php
@@ -1,5 +1,41 @@
 <?php
+/*
+ *  new.php for the creation of the misc_billing_form
+ *
+ *  This program creates the misc_billing_form
+ *  The changes to this file as of November 16 2016 to add needed items to the HCFA Print and Electronic Transmission
+ *  are covered under the terms of the Mozilla Public License, v. 2.0
+ *
+ * @copyright Copyright (C) 2016 Terry Hill <terry@lillysystems.com>
+ *
+ *
+ * LICENSE: This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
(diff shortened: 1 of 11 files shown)


aethelwulffe:
pushing to ARSP prod in an hour or so.

aethelwulffe:
In the form, where does provider qualifier code come from (17a)?  Getting a blank dropdown.  Probably because I am testing it in an older code set.  I did notice that the older code has exclude from insurance (dang useful thing) and it doesn't seem to be here....

teryhill:
I will send the exclude stuff up in a commit after this on is in place( Tony wants things seperate) so don't worry it will be there.

I will send the code up to create the list for those provider qualifiers

aethelwulffe:
That's the only missing chunk I have seen so far...but billing ppl arrive on monday...

teryhill:
I just pushed the code for that

tmccormi:
If Art approves it I can merge it while I'm waiting for my next flight...

aethelwulffe:
OK.  Exhausted, but the stuff looks like it runs.  It will might be Wednesday before it sees real-life testing *here* anyway, but I am sure it can be visually evaluated before then.  Some features we are simply not users of, and I won't be able to do real-world with it.
  The CMS-1500 print branch is the missing piece here in any case, so I am going to try to test these in conjunction with each other in a prod, but the stuff seems to _run_, but that is as far as I can take it at this point.
  My vote is to merge up, hit the squashed print branch, then see what we get ASAP, so I approve this merge request.

teryhill:
I have it running on 10 or so sites that use it and have been for five or six months.

aethelwulffe:
Yeah, but Terry, you-all up there could be suffering from high-latitude sickness and be hallyooosenatin'.

Issue #268, opened 2016-11-16
Changes to the hcfa print and x12 files for Taxonomy codes


teryhill:
I will commit this in the near future

aethelwulffe:
Thang-Kew!

teryhill:
Now for a more detailed description. This commit puts the provider Taxonomy code in 24I and 33B along with the referrer's Taxonomy code in box 17a

It also adds the provider taxonomy in the 2310b loop and in the 2000A loop

PR #276, opened 2016-11-19
Added taxonomy code for caid

teryhill:
Added taxonomy code for caid #268


teryhill:
Sorry not sure why that is in there but the taxonomy stuff is in it also

Issue #280, opened 2016-11-21
EDI 270 creation program still on the 4010 standard


teryhill:
EDI 270 creation program still on the 4010 standard needs to be brought up to the 5010 standard

teryhill:
Fixes per clearing house review

aethelwulffe:
That ISA element that identifies the delimiter used in the file can be a caret if you want.  Of course if you are saying it will be one, then you use the other...that does not work.

teryhill:
Lookie what I got Fellers

ISA*00*0000000   *00*0000000000*ZZ*55577889       *ZZ*753951258      *161130*1833*^*00501*000000001*1*T*:~
GS*HR*55577889*753951258*20161130*183348*2*X*005010X279A1~
ST*276*000000003*005010X279A1~
BHT*0022*13*LibreEHRTest*20161130*1833~
HL*1**20*1~
NM1*PR*2*Anetna*****46*753951258~
HL*2*1*21*1~
NM1*41*2*13*****46*753951258~
HL*3*2*22*0~
NM1*1P*2*Lilly Systems & Solutions*****XX*113399772~
HL*4*2*22*0~
DMG*D8*19830227~
NM1*IL*1*Barker*Stella****MI*123456789~
TRN*1*1234501*9000000000*5432101~
DTP*472*D8*20161115~
SE*15*000000003~
GE*1*2~
IEA*1*000000001~

An EDI 276 request generated thru LibreEHR

aethelwulffe:
Oh very very nice!
You're a good boy, oh yes you are!
5010 strings don't match, no delimiters (Anetna?) :) but good work man.
Getting 270/271 is the real biggie for most places though.  Verifying insurance is a big deal.

aethelwulffe:
You have full merge on this.  Closing

PR #281, opened 2016-11-21
Changes to bring the edi 270 creation up to 5010 standard

teryhill:
Changes to bring the edi 270 creation up to 5010 standard. #280


teryhill:
I will keep you updated on the testing.

Issue #283, opened 2016-11-28
Add CMS 1500 margins as a user Preferences and cleaned up globals


teryhill:
Add CMS 1500 margins as a user Preferences and cleaned up globals.

PR #284, opened 2016-11-28
Added CMS Margins to User Preferences and cleaned up globals

teryhill:
Added CMS Margins to User Preferences and cleaned up globals. Nothing
here that OpenEMR should want. #283 (Chevy Engine)

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -86,15 +86,27 @@ else {
 // xl('Appearance')
 // xl('Locale')
 // xl('Features')
+// xl('Report')
+// xl('Billing')
+// xl('Claim')
+// xl('Esign')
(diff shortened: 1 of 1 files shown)


teryhill:
The majority of the change is just removing white space and cleaning up the looks of the file. The CMS changes are actually lines 134 and 135.

library/globals.inc.php
```diff
@@ -86,15 +86,27 @@
 // xl('Appearance')
 // xl('Locale')
 // xl('Features')
+// xl('Report')
```
tmccormi: These are new lines 89-103... they should have already been there.  Makes we worried that something else in a previous commit was deleted ...

Issue #285, opened 2016-11-29
Billing widget not displaying correct numbers.


teryhill:
Customers have complained that the billing widget gives the wrong numbers sometimes. This has been happening for years.  This is a sample of the common error.
![billing_error](https://cloud.githubusercontent.com/assets/4655366/20695162/ff0eee9e-b5b7-11e6-9c80-ef58e9fd2efb.PNG) 

The error can get interesting as seen here.

![billing_error_2](https://cloud.githubusercontent.com/assets/4655366/20695229/4e04403a-b5b8-11e6-9969-796942999e58.PNG)

I will be sending up a code change to fix this soon

PR #286, opened 2016-11-29
Billing widget fix

teryhill:
Fix for long standing error in billing widget. This has been around a while in OpenEMR. #285

### DIFF
--- a/library/patient.inc
+++ b/library/patient.inc
@@ -4,9 +4,6 @@
  *
  *  This program patient.inc includes functions for manipulating patient information.
  *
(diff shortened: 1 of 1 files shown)


library/patient.inc
```diff
@@ -1856,7 +1854,7 @@ function get_patient_balance($pid, $with_insurance=false) {
           "FROM ar_activity WHERE " .
           "pid = ? AND encounter = ? AND payer_type = 0",
           array($pid, $encounter));
-        $ptbal = $insarr[0]['copay'] + $brow['amount'] - $drow['payments'];
```
teryhill: @tmccormi  this is the fix

Issue #241, opened 2016-10-19
Tags system Color Picker JS missing


tmccormi:
When adding a new tag the "color picker is blank", afterwards if you click on 'None' in the color column you get this ...
**Not Found
The requested URL /LibreEHR/interface/tags_filters/javascript; was not found on this server.**

![tag-color-javascript-missing](https://cloud.githubusercontent.com/assets/285835/19532685/ac70f672-95f1-11e6-8b24-7874a548986f.png)

tmccormi:
@kchapple can you take a look at this please - it was a clean install of LibreEHR master branch

aethelwulffe:
Status?

tmccormi:
Pull #336 fixed the color selector

PR #296, opened 2016-12-06
fixed plugin to run migrations, and added migration to add list optio…

kchapple:
…ns for tag colors and status (#241)


teryhill:
@tmccormi This looks ok to me I just think the license information need to be updated and added

tmccormi:
Why is this closed?  I haven;t merged it yet

kchapple:
@tmccormi needs license info as @teryhill pointed out

PR #297, opened 2016-12-06
Development 101 ken

kchapple:
On second though, I don't think it's necessary to add license to every file. Nobody does that (except openemr) and just creates extra effort. Many files don't have license info anyway. Thoughts?


tmccormi:
We decided with the LibreHealth team and SFC that we needed to include the license headers.
See attached file that can be used to insert the correct License and disclaimer.

[MPL2-License-Header-Snippet.txt](https://github.com/LibreHealthIO/LibreEHR/files/634891/MPL2-License-Header-Snippet.txt)

I'll check this into the repo as well today.

tmccormi:
I've merged this.  Will update the licenses in a separate pull request.

tmccormi:
@kchapple  Question, does the Migration_2016_12_05_14_48_11.php automatically run with sql_upgrade.php? 
If so this needs version.php updates if not we need some model/method or documentation on how to run migrations and how the customer would know that they need to be run..

PR #313, opened 2016-12-13
Remove the remaining old upgrade scriipts

tmccormi:
Regarding #234  clean up


aethelwulffe:
Well, I am saving my copy of the libre upgrade script.

PR #318, opened 2016-12-14
Use PID as primary criteria vs. PUBPID

aethelwulffe:
Propose a "PubPID Module"? :)

### DIFF
--- a/controllers/C_Document.class.php
+++ b/controllers/C_Document.class.php
@@ -660,7 +660,7 @@ class C_Document extends Controller {
 			
 			//validate that the pod exists
 			$d = new Document($doc_id);
-			$sql = "SELECT pid from patient_data where pubpid = '" . $file['patient_id'] . "'";
+			$sql = "SELECT pid from patient_data where pid = '" . $file['patient_id'] . "'";
 			$result = $d->_db->Execute($sql);
 			
 			if (!$result || $result->EOF) {
@@ -745,7 +745,6 @@ class C_Document extends Controller {
 		//move to new patient
 		if (is_numeric($new_patient_id) && is_numeric($document_id)) {
 			$d = new Document($document_id);
-			// $sql = "SELECT pid from patient_data where pubpid = '" . $new_patient_id . "'";
 			$sql = "SELECT pid from patient_data where pid = '" . $new_patient_id . "'";
 			$result = $d->_db->Execute($sql);
 			
--- a/controllers/C_Prescription.class.php
+++ b/controllers/C_Prescription.class.php
@@ -323,7 +323,7 @@ class C_Prescription extends Controller {
 		$pdf->ezText('');
 		$pdf->line($pdf->ez['leftMargin'],$pdf->y,$pdf->ez['pageWidth']-$pdf->ez['rightMargin'],$pdf->y);
 		$pdf->ezText('<b>' . xl('Medical Record #') . '</b>',6);
-		$pdf->ezText(str_pad($p->patient->get_pubpid(), 10, "0", STR_PAD_LEFT),10);
+		$pdf->ezText(str_pad($p->patient->get_pid(), 10, "0", STR_PAD_LEFT),10);
 		$pdf->ezColumnsStop();
 		if ($my_y < $pdf->y){
 			$pdf->ezSetY($my_y);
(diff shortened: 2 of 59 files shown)


controllers/C_Document.class.php
```diff
@@ -660,7 +660,7 @@ function queue_action_process() {
 			
 			//validate that the pod exists
 			$d = new Document($doc_id);
-			$sql = "SELECT pid from patient_data where pubpid = '" . $file['patient_id'] . "'";
```
tmccormi: THis might require a major conversion step/script.   Changes to update the documents table at least.  (for people that have existing sites)  just FYI

aethelwulffe: I think that is just "owner" or foreign_id in the documents table.  I have done this in the past and everything was cool...but that was only due to the fact that pubpid=pid.
  The conversion script would need to include that, yes...but I noticed that the Libreupgrade script was just pulled out of the code base by someone....  :)

aethelwulffe: Yes, it is foreign_id.
This would require a conversion script that says:
UPDATE documents d
        INNER JOIN
    patient_data p ON p.pubpid = d.foreign_id 
SET 
    d.foreign_id = p.pid;

tmccormi:
Which is why it was more of an FYI than a showstopper :-)

tmccormi:
just need the database.sql tweak (and perhaps layout_options)

aethelwulffe:
Keeping pubpid...I didn't write it out of the whole idea.

aethelwulffe:
This would also affect ZEND stuff if I ripped it out now.

tmccormi:
OK...  I think we could drop it in favor of the user adding a custom field to layouts for the specific external ID they want to track...  but it's fine.

Does the SQL write still duplicate the PID into the PUBPID field if empty?

aethelwulffe:
It always was supposed to, but that was not working due to the old INNODB issue with incrementing a null value.  I think this is taken care of with this.

aethelwulffe:
I have NOT tested this shit yet!!!!

tmccormi:
@teryhill what do you think?

tmccormi:
will test before merging :-) ... you too ...

teryhill:
Gives me the ibee jebees .

aethelwulffe:
Scary, yes...but I have also done the whole "find and replace everywhere" deal without actually reviewing code, and everything still worked, even the double pid selects in the queries.
  This is a little neater than that.
PUBPID gives me the willies already.

aethelwulffe:
Well, documents are broken in this branch...but they are also broken in 101!
Please check this, and see if it is just an issue with the db I am using, or if treemenu is having issues.

tmccormi:
I am able to upload documents in my local dev 101 install 
document

aethelwulffe:
Must be my database.
I will check...

aethelwulffe:
[Wed Dec 14 20:49:23.750085 2016] [core:error] [pid 4828:tid 1872] (20025)The given path contained wildcard characters: [client ::1:51369] AH00036: access to /libre101/controller.php?practice_settings&document_category&action=list failed (filesystem path 'C:/xampp/htdocs/libre101/controller.php?practice_settings&document_category&action=list')
and
4 -0500] "GET /library/js/DocumentTreeMenu.js HTTP/1.1" 404 1321 "http://localhost/libre101/controller.php?practice_settings&document_category&action=list" "Mozilla/5.0 (Windows NT 10.0; WOW64; rv:50.0) Gecko/20100101 Firefox/50.0"

This is with development101

tmccormi:
Your tree issues are not related to this pull request....  should I merge this or not?

aethelwulffe:
I thought I was the only one testing this.  If you have verified you can upload documents on your system with this branch, then fine, do it.  I may or may not have time to figure this out this weekend to see what is broken due to a house full of people.  I have a few minutes this morning only because my body is too shitty to lead the morning trail run with the rest of the family...so they left me all alone (muahahahahhaha).
  Anyone else tested?

teryhill:
I have not tested this

aethelwulffe:
Do documents work on your 101?  Can you switch branch code?

It may be that I need a new DB install due to a recent change. We got no 
upgrader, so...

aethelwulffe:
Nope.
Installed server and everything from the old .exe distro, tried that, dumped the DB, replaced it with 101, and no documents.
That issue has nothing to do with this branch, so long as it doesn't cause issues with fixing things later, let's merge it.  Everything else seems no worse than before, and some things are clearly fixed.

tmccormi:
Just to be clear (once more) this does not make any changes to the database itself correct, @aethelwulffe ?

aethelwulffe:
No, not at all, negative.

tmccormi:
Then I'm going for it...

PR #322, opened 2016-12-16
Added a selection for 'All' and added 'ptkr_date_range', to user prefs

teryhill:
Added a selection for 'All' to status section in edit_globals and added 'ptkr_date_range' to user prefs.


aethelwulffe:
Tests fine on top of the other stuff.

teryhill:
It tests well on my systems

Issue #326, opened 2016-12-20
DocumentTree menu gone again


tmccormi:
I'm getting this error again on new installs:
GET https://libreehr.vu2vu.com/demo/libreehr/DocumentTreeMenu.js 
controller.php?document&list&patient_id=1015:27 Uncaught ReferenceError: TreeMenu is not defined
    at controller.php?document&list&patient_id=1015:27

They were moved in this commit
24c1ea1 Locate DcoumentTree_Menu.js (#252)
D       CategoryTreeMenu.js
D       DocumentTreeMenu.js
A       library/js/DocumentTreeMenu.js

They code still expects them to be in the root dir as near as I can tell

tmccormi:
This fixes it ... but it's a hack: 
in the EHR root
# sudo ln -s library/js/DocumentTreeMenu.js DocumentTreeMenu.js

tmccormi:
Did I mention how much I hate javascript?   I'm getting a 404 not found on a file that is CLEARLY there...

![whereswaldo](https://cloud.githubusercontent.com/assets/285835/21373592/f0baf0da-c6d4-11e6-9055-6ce39f47c8b9.png)

```
tony@dragonfire:/var/www/LibreEHR/templates/documents$ ls ../../library/js/DocumentTreeMenu.js 
../../library/js/DocumentTreeMenu.js

```

tmccormi:
OK it appears to be fixed.  Pull request coming, please test..

```
index e48ea6d..4f0aba4 100644
--- a/templates/document_categories/general_list.html
+++ b/templates/document_categories/general_list.html
@@ -14,10 +14,10 @@
 </style>
 {/literal}
 <script type="text/javascript">
-var deleteLabel="{xl t="Delete"}";
+var deleteLabel="{xl t='Delete'}";
 </script>
 
-<script type="text/javascript" src="../../library/js/DocumentTreeMenu.js"></script>
+<script type="text/javascript" src="library/js/DocumentTreeMenu.js"></script>
 <table>
        <tr>
                <td height="20" valign="top">{xl t="Document Categories"}</td>
diff --git a/templates/documents/general_list.html b/templates/documents/general_list.html
index 4e61ca5..b13ac49 100644
--- a/templates/documents/general_list.html
+++ b/templates/documents/general_list.html
@@ -7,7 +7,7 @@
 {literal}
 {/literal}
 
[...]

PR #330, opened 2016-12-21
Fix path and minor syntax error #326

tmccormi:
For #326

### DIFF
--- a/templates/document_categories/general_list.html
+++ b/templates/document_categories/general_list.html
@@ -14,10 +14,10 @@
 </style>
 {/literal}
 <script type="text/javascript">
-var deleteLabel="{xl t="Delete"}";
+var deleteLabel="{xl t='Delete'}";
 </script>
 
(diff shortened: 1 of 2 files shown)


templates/document_categories/general_list.html
```diff
@@ -14,10 +14,10 @@
 </style>
 {/literal}
 <script type="text/javascript">
-var deleteLabel="{xl t="Delete"}";
+var deleteLabel="{xl t='Delete'}";
 </script>
 
-<script type="text/javascript" src="../../library/js/DocumentTreeMenu.js"></script>
+<script type="text/javascript" src="library/js/DocumentTreeMenu.js"></script>
```
aethelwulffe: Not sure where this template is called from and how that affects the src...but if globals.php is called with it, then we have a path variable for the document root that can be passed to javascript.  I am thinking we need to pass that to js globals in any case.
If this works...then fine.  We should still review the globals and session variables some time in the future with JS in mind.
Good catch on the quotes.

PR #331, opened 2016-12-23
Changed to reflect table

teryhill:
Field in table 25 char only using 10 in program. User needed to send note on drug due to new 2017 hcpcs requirement. This field could stand to be expanded in the table to a larger value

### DIFF
--- a/interface/forms/fee_sheet/new.php
+++ b/interface/forms/fee_sheet/new.php
@@ -314,7 +314,7 @@ function echoLine($lino, $codetype, $code, $modifier, $ndc_info='',
 
     if ($code_types[$codetype]['claim'] && !$code_types[$codetype]['diag']) {
       echo "  <td class='billcell' align='center'$usbillstyle><input type='text' name='bill[".attr($lino)."][notecodes]' " .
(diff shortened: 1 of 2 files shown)


teryhill:
@tmccormi @aethelwulffe I think the field size in the billing table needs to be changed to 50 or 75 char instead of the current value of 25 char.

teryhill:
how do I need to change the field size then. We don't have an upgrade script any more, do we?

tmccormi:
Just edit database.sql - only supporting new installs officially.    Upgrade and migrations scripts with the 1.01 release.

teryhill:
So what do we want 50 or 75 or more @tmccormi @aethelwulffe

tmccormi:
I'm not sure what it's purpose is even.  So I abstain from voting.

teryhill:
This is the explanation from the clearing house

noc_explain

tmccormi:
unclear what kind of note you would put on a single service code.  What is the clearing house limit on the note record?

teryhill:
This is what they say. I personally think 50 should be plenty.

image

tmccormi:
Sounds good then

aethelwulffe:
The x-12 field limit is 80 for a NOC description in a SV1 segment.

PR #333, opened 2016-12-28
fix to insurance inactivate and change web address

teryhill:
fix to insurance inactivate when creating a new patient.


interface/patient_file/summary/demographics.php
```diff
@@ -29,7 +29,7 @@
  * @package LibreEHR
  * @author Brady Miller <brady@sparmy.com>
  * @author Terry Hill <teryhill@librehealth.io>
- * @link http://www.libreehr.org
```
tmccormi: bad url.  it http://librehealh.io No WWW

teryhill: Changed that commit on the way. I have this code running on a customers site . That is how I found the problem originally.

PR #354, opened 2017-01-23
Add global setting for page preference for second tab when starting app

tmccormi:


library/globals.inc.php
```diff
@@ -5,65 +5,16 @@
...
-//   Serbian                        // xl('Serbian')
-//   Sinhala                        // xl('Sinhala')
-//   Slovak                         // xl('Slovak')
-//   Somali                         // xl('Somali')
-//   Spanish (Latin American)       // xl('Spanish (Latin American)')
-//   Spanish (Spain)                // xl('Spanish (Spain)')
-//   Swedish                        // xl('Swedish')
-//   Turkish                        // xl('Turkish')
-//   Ukrainian                      // xl('Ukrainian')
-//   Vietnamese                     // xl('Vietnamese')
```
teryhill: You know these are here to translate the languages

library/globals.inc.php
```diff
@@ -82,34 +33,15 @@
...
-// xl('Security')
-// xl('Notifications')
-// xl('CDR')
-// xl('Logging')
-// xl('Miscellaneous')
-// xl('Portal')
-// xl('Connectors')
-// xl('RX')
-// xl('PDF')
-
```
teryhill: The same with these , to allow translation of the tabs

tmccormi:
They are just comments?  why would the be needed?  I can, of course put them back if some external process using them, but geez.

teryhill:
Translation routines look for the xl where ever they are

teryhill:
Don't see any obvious problems in the code, does it test well?

tmccormi:
Just found a bug.  It's not letting the tab 1 get changed.  it's hardcoded to calendar somewhere.  gonna ping Kevin

tmccormi:
interesting. It works as a global setting, but not as a user_pref setting

teryhill:
what is in the globals table for the records

tmccormi:
doesn't appear to be writing to the globals table at all for those defaults.  Do you see anything obvious in the globals.inc.php that I'm blind too?

tmccormi:
weird,  it doesn't save default_tab_2 in user settings and it ignores default_tab_1, but not always...
user_settings

teryhill:
just for kicks change the name of the globals to first_default_tab and second_default_tab

tmccormi:
what good is that, the code uses the current names...  I don't understand what you are trying to see.

teryhill:
if it was being truncated some where but since you see them now scratch that

tmccormi:
I have new code to test and update ... please hold

aethelwulffe:
Tell me when it is at test stage.
  I have a lot of catching up to do

tmccormi:
see issue #355 I'm testing my branch merge now and will push when I have decent results

tmccormi:
This is now testing well on my side, and ready for other to poke at... @aethelwulffe @teryhill

aethelwulffe:
OK,

Seemed to work.  I get billing report now.

tmccormi:
Two tabs seems good to me.  Now that I know how yehster did it we could add a third.   Better would be to have a 'save last' function for users (like demographics).  Then the user would be able to override even the defaults ... future.

Issue #361, opened 2017-01-26
Mods to Globals order of operation


tmccormi:
Globals don't behave the way I thought.I thought that the initialization process worked to first get defaults from the globals array and then look in the database. That way you wouldn't need to save globals to get the defaults of globals that are newly added. But you have to go and save them in order to take effect. 

So what I was seeing was what happens when there's no global for default screens set. What do you think about enhancing globals so that the user doesn't have to go to edit-globals and save for newly added globals to be set?  --Ken

teryhill:
the global default should be added to /interface/globals.php in the section starting about line 343

aethelwulffe:
What is the status of this issue?  We have a commit here, but I am not sure if that is the whole issue.

aethelwulffe:
Can we get a better definition of what this issue is to address (as a specification for a work ticket)?

PR #364, opened 2017-01-29
changes for a new install with default panes not set

teryhill:
changes for a new install with default panes not set #362  #361


tmccormi:
This is acceptable for the time being, until we think of something better...

PR #366, opened 2017-01-30
Demographics - add CSS ID to sections for later control and some cleanup

tmccormi:
This is just some cleanup and adding some IDs to sections that should have ID's so we can do styling or other DOM control sugar.


tmccormi:
Anybody got any issues with this one? @aethelwulffe @teryhill ?

aethelwulffe:
No. 
Looks good.

Issue #368, opened 2017-01-30
Globals ... commit missing


tmccormi:
@terry somehow code from hide_billing_widget global added on 8/25 by Terry, Commit 7ac4285 is missing in the latest changes

PR #369, opened 2017-01-30
Added back hide billing widget

teryhill:
Added back hide billing widget #368


tmccormi:
does the demographics.php still have the code?

tmccormi:
I see it.  merging... quick work

PR #371, opened 2017-01-31
added ids to widget rows, removed auto-loading of past encounters tab

kchapple:


teryhill:
I am assuming that this is for some of that magical stuff that does nothing for us now

tmccormi:
Yes, but it's good practice to include ID in HTML for CSS and the hooks give us power to do plugins ...
more plugins in more places!  That's my motto

tmccormi:
Removing the autoloading encounters history tab could be debatable.  I don't like it myself and you can open it with a click on the title bar view history.

What do y'all  think?

teryhill:
It ain't in demographics or options so we must not use it

tmccormi:
Not sure if that was sarcasm or not .... :-)

teryhill:
no not sarcasm, I don't  interact regularly with the code it is in so I don't think it is applicable to me yet

aethelwulffe:
Mongo dunno.
I say give it a try.  If it is configurable as a default tab/widget, then there you go.

teryhill:
Is there a standard format y'all want for them, "Yes, but it's good practice to include ID in HTML for CSS and the hooks give us power to do plugins "

aethelwulffe:
$widgetId = str_replace( " ", "_", $label ) . "_widget";

So, looks like right now, whatever the $label is, it removes spaces and 
adds _widget.

I suppose that is some sort of standard...so I guess you have to make 
sure not to use the same label for two things, hunh?  No sanity check there.

tmccormi:
Different topic, but yes, ID should be unique and CLASS for common  CSS.

Issue #374, opened 2017-02-01
Left_nav needs to go bye-bye


aethelwulffe:
We are done with that code.  daemon_frame is also toast, as is a lot of the associated stuff for old layouts in the patient_encounter and other areas.

PR #375, opened 2017-02-01
deprecate left_nav

aethelwulffe:
leave left_nav to leverage more levity.
left_nav and daemon_frame gone.
#374


tmccormi:
if only you dropped the concurrent_layout and leftnav width and other crap that goes with it, I'd be in hog heaven

aethelwulffe:
I'll get along with it.  I think concurrent layout is a bit of a job, 
and some scary stuff happened when I went chop-chop.

aethelwulffe:
I tried, and something went funny.  Thought about making a comment that 
it needed doing...so maybe I can try again...or you can push one?

tmccormi:
wouldn't be the first time we tried that an reverted ...

PR #376, opened 2017-02-01
Nation Notes as a Module

aethelwulffe:
Now adding Nation Notes back in as a module:
Added two "mod theme" files in /themes to get the ball rolling for themes for mods...and...
Core code changes:

In library/options.inc.php:
if ($GLOBALS['mod_nn'] == true){
  require_once(dirname(dirname(__FILE__)) . "/modules/nation_notes/nn_library_options.inc");}
else{
[main file contents]
}

if ($GLOBALS['mod_nn'] == true){
  require_once(dirname(dirname(__FILE__)) . "/modules/nation_notes/nn_lbf_new.inc");}
else{
[main file contents]
}

In interface/main/themes/ltr_mod_themes:
@import url("../../modules/nation_notes/nn_themes/nn_ltr_theme.css");

In interface/main/themes/rtl_mod_themes:
@import url("../../modules/nation_notes/nn_themes/nn_rtl_theme.css");

Remaining tasks:  
*Add globals flag or registration for module
* add css selection controls for mods as well as LTR/RTL configuration,
* add script to run the database upgrade as well as the acl upgrade.


teryhill:
Why????? I use it in LBF's

aethelwulffe:
I was restructuring it as a module and adding a hook system (once I could figure out what that was) while getting rid of about 50% of the files in the package.
I also though that it wasn't very popular due to how hinkey it seems to be.  On the other side of that, most everyone I deal with is using structured data, and another way of creating narrative mish-mash unsearchables hasn't been very popular.  LBF's are almost impossible to do complex reports against.  The queries become enormous, and you can never be sure that someone isn't just going to change the form on you!
  This is all part of the component sprawl project we have been talking about.  The resources splattered a good bit, and the footprint was huge.  I figured it could be done better, and it looked like a good candidate module.

teryhill:
Ok

aethelwulffe:
We can talk about this long-about 4PM, right?  I am just throwing stuff 
out there to get the organization and directory structure project kinda 
moving along.  I don't really have it in for NationNotes or anything.  I 
was just seeing a lot of discussion and no moving feet, so I put some 
stuff out there.  If nothing else, this commit shows where every 
component of Nation Notes is, right?

aethelwulffe:
Now I have added Nation Notes back in as a module.  I like the model of two commits on the branch when ripping out and re-arranging.  It seems to help keep me organized real nice.  Having the blame to find when/how something was first dumped into the pile is good too!

PR #377, opened 2017-02-01
Dump unused fonts

aethelwulffe:
HTML2PDF is hard-coded in the ony place that uses it (custom_reports) to
use the dejavusans font set only.  The rest will not be called.
Hopefully this is the first stage to replacing html2pdf with fpdf or
something else.


teryhill:
Let's not dump this until the release is out

aethelwulffe:
These fonts are not used by anything, anywhere.

aethelwulffe:
This is a 2.3% reduction of the code base file numbers.  With Nation Notes, it is 12.6%

Issue #379, opened 2017-02-01
Error when re-opening a claim


teryhill:
Get the following error when re-opening a claim on the demo site . This results from the removal of the auto increment in the file. the updating of the record was not working. 

![image](https://cloud.githubusercontent.com/assets/4655366/22528834/fda7951a-e8a2-11e6-83e1-a1daf19b59de.png)

PR #380, opened 2017-02-01
This is to fix an error in the reopen of the claims

teryhill:
This fix is needed because of the change in table type and the removal
of the auto increment in the claims table.
Original changes from OpenEMR. Matrix and Brady (I think) #379


aethelwulffe:
I like this fix.  The actual autoinc change came from me in this project with the INNODB bit.

PR #384, opened 2017-02-02
Nicer Calendar Icon

tmccormi:
Any objections to this look?

calendar-picker-icon


aethelwulffe:
I much prefer this one:
index

aethelwulffe:
Only possible issue is alignment, as the old was 24px and this is 16.  
valign would be nice.

Issue #335, opened 2017-01-04
Taxonomy code giving rejections


teryhill:
Apparently I am ahead of caid on the code changes . We will need to revert them or ,since we have just a very few clients, we can change it as needed.

Let you guys make the call on it.

aethelwulffe:
Can I please have more info on this?

PR #385, opened 2017-02-02
added ordering Physician from Misc Billing Form

teryhill:
Added code to fully support the Box 17 enhancements to the paper form in
the electronic submission. #335


teryhill:
Yep this is a stop gap till we revamp the fee sheet

PR #386, opened 2017-02-02
Database changes added statuses, added qualifier code and fixed field…

teryhill:
… in claims

Database changes added statuses, added qualifier code and fixed field in
claims #379


teryhill:
The last two are already handled in the code

aethelwulffe:
Just wondering if we should handle them all in the same way...and maybe even the facilities and all that as well...eventually.  Everywhere that uses an NM1 element really.  Frankly, I would like to see "primary" in insurance_data get changed to "P" etc... and all the values used in x12 also used in our data fields.

teryhill:
Sounds like a design spec it starting to ooze out

PR #387, opened 2017-02-02
Changes to Annotatable Diagrams

teryhill:
Changes to Annotatable Diagrams #356


teryhill:
Still need to make them auto install

aethelwulffe:
So is this a complete commit or no?

tmccormi:
Yes, I merged it.  Making it a preinstalled default form can be a separate pull.

Issue #383, opened 2017-02-02
Admin->Other->Language - Edit Def crashes with SQL error


tmccormi:
Edit Definitions ...
Enter a term, click Search

![language-edit](https://cloud.githubusercontent.com/assets/285835/22532436/f2556b12-e89b-11e6-9e25-1fb478cc0d1d.png)

teryhill:
This does not happen on the demo site, nor on my local instance

tmccormi:
It has to do with the database being created using COLLATE utf8mb4_unicode_ci which is appropriate for Europe...  and the system allows you to select that at setup, so the editor should work with alternative collations ...

aethelwulffe:
Sorry,

My machine has the whole dot as a floating point thing going on. Can't 
REALLY test that without a EuroPuter.;-)

tmccormi:
I'm not using a Euro putter .. just a website that is hosted in EU...
I'm not sure yet if it got set that way on purpose or it's some kind of default we did not notice...

aethelwulffe:
Closed the commit against this without merge.  If it is not really an issue, I suggest we close this.

tmccormi:
It's still an issue for anyone that creates a db with charset that is not utf8_general_ci

aethelwulffe:
Do we want to check and set a global for the characterset value?

PR #390, opened 2017-02-02
Make language tool multi-lingual #383

aethelwulffe:
Check for what the actual collation is instead of just assuming it.
This could use more refinements, and need to make sure mysqli actually
works, which it probably doesn't.  This is a WIP commit.


aethelwulffe:
mysql_client_encoding($this->_connectionID) might work.
I need to check this.
In any case, it sounds like a function that needs to be added to the connection library.

PR #391, opened 2017-02-03
Move HTML2PDF to modules directory

aethelwulffe:
Removing this component from another project to /modules.
This module is called from one location in the code:
/interface/patient_file/report/custom_report.php


aethelwulffe:
This includes paring down the unused fonts.  If some composer or other build stuff could go away too, that would be nice, so that it is just the working bits.

PR #393, opened 2017-02-03
fixed bug in migration that was chopping the t character off the end …

kchapple:
…of #If* statements


aethelwulffe:
Pretty simple.
  I have no idea what anything in these directories six levels down are for or related to, but the change makes sense.

PR #396, opened 2017-02-06
Move HTML2PDF to modules directory

tmccormi:
Removing this component from another project to /modules.
This module is called from one location in the code:
/interface/patient_file/report/custom_report.php


aethelwulffe:
bump

tmccormi:
just to be clear, it's just a move, not a massive code change

interface/patient_file/report/custom_report.php
```diff
@@ -46,7 +46,7 @@
 $PDF_OUTPUT = empty($_POST['pdf']) ? 0 : intval($_POST['pdf']);
 
 if ($PDF_OUTPUT) {
-  require_once("$srcdir/html2pdf/vendor/autoload.php");
+  require_once("../../../modules/html2pdf/vendor/autoload.php");
```
tmccormi: backwards relative paths ... someday we need a modules location constant ..

teryhill:
It appears to just be a move. what do you want to call the modules global?

aethelwulffe:
How about $modules ?

aethelwulffe:
Or $modules_root or $modules_dir
..I like those better

teryhill:
$globals['modules_dir']

aethelwulffe:
Added $GLOBALS['modules_dir'] to the branch.  This will break it, and will only be testable when the additional global is added.  I am doing this under the assumption that this is a reviewed requirement, and not something that we were going to change after committing.

teryhill:
@tmccormi  It appears to be a move . I up voted this

aethelwulffe:
I will be marking in the new globals, but those are in the nation notes branch at the moment.  It would be a second commit merge to do those first, then four more commits to the two branches in question...so I would like to just test this again as-is, then if the nation notes gets merged, I can just add one more commit to this to add the global paths.  This does not require a toggle.  I am mostly doing it to make integrating the PDF functions in the whole application to one module we can maintain and use it's documentation to track everywhere it is used (and needs test of hooks when changed).  Then the call to the PDF modules can be added to the API we will have one day.

interface/patient_file/report/custom_report.php
```diff
@@ -46,7 +46,7 @@
 $PDF_OUTPUT = empty($_POST['pdf']) ? 0 : intval($_POST['pdf']);
 
 if ($PDF_OUTPUT) {
-  require_once("$srcdir/html2pdf/vendor/autoload.php");
+  require_once($GLOBALS['modules_dir'] . "/html2pdf/vendor/autoload.php");
   $pdf = new HTML2PDF ($GLOBALS['pdf_layout'],
```
teryhill: I did a search and this is used in sl_eob_search.php also.

aethelwulffe: Uh, I only see one of the other PDF functions there, which is class.ezpdf.php which is an extension of class.pdf.php, both of which are still in /library/classes

aethelwulffe: Am I wrong?  "HTML2" does not show up in that file...

teryhill: The change to add Ray's statements had that change. Is it not still in the Codebase?

teryhill: line 94 of sl_eob_search.php

aethelwulffe:
Line 94 of interface/billing/sl_eob_search.php reads:    while(!feof($file))
Apparently either Ray's changes never made it in, or perhaps it could be that my branch is crap because it has been sitting here for months.
  I am not re-building, re-basing this or anything like it.  If it takes a second commit to make the sl_eob bit work, fine.  I am investing way too much time working on my stale pull requests.

teryhill:
I checked and the code is in the main code base.

aethelwulffe:
You are right.  sl_eob has changed since then, and this feature was added.

aethelwulffe:
Should I just burn this?

teryhill:
no commit it and then fix sl_eob on another commit,

aethelwulffe:
It isn't cool to commit your own pulls.  Red five standing by.

teryhill:
Where is Blue Leader @tmccormi

aethelwulffe:
I know where Gold Leader is....

tmccormi:
blueleader is going to trust the team on this one ...

PR #397, opened 2017-02-06
Nation Notes Module

tmccormi:


aethelwulffe:
21 days on the pull list...and now conflicts.  Crap.

aethelwulffe:
OK, I am dumping this commit completely in 24 hours.  If you don't want a commit, please DOWN-VOTE IT, so I can not worry about maintaining and rebasing it.

teryhill:
@aethelwulffe Has this been running some where? I don't have a problem with the changes. Just have not downloaded them or tested them.

aethelwulffe:
Not "having a problem with" is different than "think it is worth anything".  It has to be considered a net value in some way.  I assumed that this sort of thing was what we were looking for to move out of the realm of the spaghetti monster.
Please give honest opinions before I bother re-factoring to a new branch again.  It is 713 files.  That stuff isn't easy!

teryhill:
@aethelwulffe  do you have this working? 713 file is a lot.

aethelwulffe:
The earlier "Toss nation notes" commit is running on what is *the only* production instance of LibreEHR I know of.
Almost all of the 713 files involved represent trash and fluff that was included with the original commit.  If you exclude unneeded files, contributions to the project start looking pretty small!  The "real" bit of the code is a few thousand lines.  I dumped out 112,000 even while keeping the functionality as an optional module.
I am going to re-enable the module on the production instance, then turn it off again as a test.  I will also install the branch from ground zero, test it by loading imported data from claim files (which just add to, not replace the db), then enable the module and test it by trying to make some sort of nation Notes form element.
  I will report back with the process list and all that.

aethelwulffe:
nation1

Installed, ACL is fine.
Table check is SAT.
HOWEVER:  in rectifying the conflicts, the added user field `suffix` is creating an error.  SQL and that one little bit of code are at different stages, which makes this branch goofy and stale.  I will try to overcome that.

aethelwulffe:
Clean install when using lower case:
image
The create new patient feature is broken just the same as it was in the base version of this branch.

aethelwulffe:
Re-basing a twig branch from master to get the full picture.

aethelwulffe:
Table total numbers were correct.  Two tables used by Nation Notes were not installed (as intended) giving the correct number.

teryhill:
There is a fix for the create new patient in the pr's.

aethelwulffe:
What are the PR's?

Sorry, jumping back and forth between this and getting counselor's 
ancient laptops to run...

teryhill:
Pull request's

aethelwulffe:
@teryhill 
Looks like when you added the provider select function to options.inc.php, you put a close bracket in the same spot I did...and when I corrected the conflicts, there wasn't syntax checking happening...so we wound up with a missing close bracket.

aethelwulffe:
Everything good now...just getting a deceased notice DECEASED (17228 days ago)

PR #398, opened 2017-02-06
Kevins edi v2

tmccormi:


aethelwulffe:
This should be a module if we pop it in (I am willing to work it if need be) but we need to decide on what to do with this pull request.

teryhill:
It will be a swap for the old EDI stuff. So if you want to remove the old and make this a module go fer it

aethelwulffe:
This commit would take a lot of work to get it to mesh with LibreHealthEHR at this point...  The old edi history stuff is totally out of date, but was only a dozen or so files.  This is a half a million lines of code added to the project.  The first issue I have is that I'm scared that trying to diff it all will freeze my browser!
  I am going to take a deeper look into the branch and see what can be done with this.  Probably the usual issue of taking an existing project framework, making use of it to display a few items nicely on the page, and including a huge bloat of files just to get a couple of functions worth of visual stuff.

teryhill:
It will not take much it is mostly self contained. The assets are newer versions of some of the existing ones and a few of them were introduced in portal. 

Terry

aethelwulffe:
Looks like most of the code is in the form of developer kit demo files and the like buried in each version of jquery, and...well in every project resource in the stuff.
  Looks like we could just put the work in and get the assets straight...or put it in that plugin repo like we have been talking about and not worry about it.  It does seem like it could have a *very* light footprint if we get the assets located...in which case I would still stick it in modules, but leave it cannon. It is something that has to get updated according to an external standard (x12) and it is already a bit out of date on a few things in 997 I spot checked.

teryhill:
been around since Nov 16

PR #402, opened 2017-02-06
Make language tool multi-lingual #383

aethelwulffe:
Check for what the actual collation is instead of just assuming it.
This could use more refinements, and need to make sure mysqli actually
works, which it probably doesn't.  This is a WIP commit.

### DIFF
--- a/interface/language/lang_definition.php
+++ b/interface/language/lang_definition.php
@@ -57,9 +57,10 @@
 <?php
 
 // set up the mysql collation string to ensure case is sensitive (or insensitive) in the mysql queries
+
 if (!$disable_utf8_flag) {
   $case_sensitive_collation = "COLLATE utf8_bin";
-  $case_insensitive_collation = "COLLATE utf8_general_ci";
+  $case_insensitive_collation = $mysqli->character_set_name();
 }
 else {
   $case_sensitive_collation = "COLLATE latin_bin";


interface/language/lang_definition.php
```diff
@@ -57,9 +57,10 @@
 <?php
 
 // set up the mysql collation string to ensure case is sensitive (or insensitive) in the mysql queries
+
 if (!$disable_utf8_flag) {
   $case_sensitive_collation = "COLLATE utf8_bin";
-  $case_insensitive_collation = "COLLATE utf8_general_ci";
+  $case_insensitive_collation = $mysqli->character_set_name();
```
tmccormi: This doesn't seem right.  why would you hard code one and lookup the other...  seems like just reading the database setting and using that across the board would be the best approach

aethelwulffe: You are right.  Should that not be something that the controller grabs for session data, or am I still thinking too small and too late?  Would that not be better than checking it here every time?
I imaging that the libreehr user has permissions to ask the database what the collation is.

tmccormi:
I would think so, we should just be able to get that info for the database.   The tricky widget is when individual tables have different character sets and collations (which I have seen in upgrades).  but that might not be a problem for LEHR going forward.

aethelwulffe:
We add that to the migration checklist "give all tables the same collation"

aethelwulffe:
Dropping this commit in 24 hours.

PR #403, opened 2017-02-06
Theme and asset consolidation

aethelwulffe:
Moving assets used for styling of the application to interface/themes.


tmccormi:
This looks fine.  @aethelwulffe Have you tested or do you need me to?

aethelwulffe:
I tested while developing, but there are many twists between here and there.  I like to see the stuff run on linux as well.  Otherwise I have to set my machine to case-sensitive directories and some other stuff.
  If you are set up to pull down and run easily, please do test.

tmccormi:
That's what  figured, will do.

aethelwulffe:
bump

aethelwulffe:
Tested each and every theme on Ubuntu16.04 and Win10 Enterprise.  All works.

aethelwulffe:
This was desktop 16.04 with permissions pretty loose, but I don't think that will affect anything.  Primary linux concern was the default case-sensitive issue.

PR #404, opened 2017-02-06
Resource directories

aethelwulffe:
Assets directory for third-party resources and libraries not maintained by LHEHR.


teryhill:
@kchapple @tmccormi @aethelwulffe  Is this structure ok with everyone? I can't bring in portal until this is settled.

teryhill:
@kchapple @tmccormi @aethelwulffe this is the current structure used inside of patient portal, This is some of the assets used in the portal. This is the basically the same structure used in OpenEMR.

datastructure

aethelwulffe:
Terry, do you see any issues with using assets/jquery or assets/js/jquery for the jquery package directories, assets/fonts for stuff like font-awesome, assets/js/bootstrap for bootstrap stuff, etc?  I would just like to keep the top level assets directory itself clean of anything by categories of assets, and prefer sub-categories to be marked below that if we are talking about a huge ton of stuff in one directory.

aethelwulffe:
For the directory location globals, I would rather concatenate several variables together rather than have to fully path each one.
For instance, if you want to change the name of the assets directory, you could just change that database variable, then the global for the directory path of everything else under it is automatically taken care of.

teryhill:
I do not see a problem, we could use a globals for jquery and a separate one for bootstrap, just use the globals for the top directory so it would be like this globals\jquery\jquery-min-2-2-0\index.js  I think it would be best if the structure was not too deep.

aethelwulffe:
Shall we go ahead with this commit then as-is?

aethelwulffe:
Terry.  Merge this if it is where you want it.

PR #406, opened 2017-02-07
Added headers and changes some of the translation

teryhill:
Added headers and changes some of the translation general cleanup


tmccormi:
Terry - these programs were not authored by you.  You should remove that line from the header.   Your cut and paste template needs a tweak for this kind of update.

Issue #410, opened 2017-02-10
PFB Change . Customer wants to see checked out items


teryhill:
Currently you can set a roll off timer for the checked out items so while the flow board is running they roll off (disappear). This is handy for the normal working of the flow board. This customer wants to go in and periodically check the checked out items bus does not want to go to gloabls to reset the switch. When a status is selected IE checked out then the roll off timer is circumvented.

aethelwulffe:
Looks like a good feature, and is +/- 53.  No lines added!

PR #411, opened 2017-02-10
Customer wants to be able to see the checked out items if they are re…

teryhill:
…quested

Customer wants to be able to see the checked out items if they are
requested. If you set the roll off timer for the checked out items they
are removed from the screen This user wants to see them later with out
resetting the globals flag. #410


aethelwulffe:
The flow board is your baby, and it didn't crash when tested.  I have no further ability to verify the operation in a clinic, but nothing seems to be too radical here.

aethelwulffe:
Merging at my level since the team is down to skeleton.

Issue #412, opened 2017-02-10
Patient list by referrer


teryhill:
I brought in report helper written by Rod Roark for the maintaining of
the titles when scrolling the report on the screen. This will give a
list of patient demographics by the referring provider also added the
new datetime picker.

PR #413, opened 2017-02-10
Patient list by referrer

teryhill:
I brought in report helper written by Rod Roark for the maintaining of
the titles when scrolling the report on the screen. This will give a
list of patient demographics by the referring provider also added the
new datetime picker. #412


library/options.inc.php
```diff
@@ -3159,5 +3160,32 @@ function lbf_current_value($frow, $formid, $encounter) {
...
+# If you specify the $allprov it will list everyone that has an entry in the NPI field.
+function genProviderSelect($selname, $toptext, $default=0, $disabled=false, $allprov=false) 
+{
+  if($allprov == 1) {
+  $query = "SELECT id, lname, mname, fname FROM users WHERE " .
+    "( authorized = 1 OR npi != '' ) " .
+    "AND active = 1 AND ( info IS NULL OR info NOT LIKE '%Inactive%' ) " .
+    "ORDER BY lname, fname";
+  } else { 
+     $query = "SELECT id, lname, mname, fname FROM users WHERE " .
```
aethelwulffe: I think that using "LIKE" in here is a very bad idea predicated on people using particular text.  There is a provider type available, authorized, and active.  Ensuring that the provider types selection in the user admin or in the address book or other GUI's allow the user to clearly say "this is a provider of some type of service", and using the other active and authorized status markers are what we should go with.  I have found text like "Her Medicaid License Was Inactive During Last March" in a user info box before, which messed lots of things up in a very mysterious fashion.

aethelwulffe:
Gimmie your input on that comment Terry.  This bit you have here is something that should have been done years ago in any case, so I am anxious to see it merged.

teryhill:
looking at it now Art

teryhill:
made the change Art. So if referrers are added to the address book they need to have an entry in the NPI field to show up in the referrer search Providers from the user file will have to be authorized.

aethelwulffe:
The other issue is if they don't use NPI's at all in whatever country, what do we do?
I suppose we can hit that at a different scope.

teryhill:
does not matter the software does not validate the NPI so you could put anything in there as long as something is there 8^)

aethelwulffe:
True....with the right translation.  Then again, picking a provider type 
and using that would probably work a lot better without being a "hidden 
feature".  How many times do we hear "Dr. x doesn't show up..." or "How 
can I get Administrator or Jerry the Janitor to stop showing up?".

Issue #372, opened 2017-01-31
need to make a directory to house the tools used in system development


teryhill:
currently there are jquery files and tools stored in the /library/js directory. These are what is stored there:

ajax_functions_writer.js
ajtooltip.js
blink
calendarDirectSelect.js
clickmap.js
common.js
datatables
DocumentTreeMenu.js
fancybox
fancybox-1.3.4
jquery-1.2.2.min.js
jquery-1.4.2.min.js
jquery-1.4.3.min.js
jquery-1.5.js
jquery-1.6.4.min.js
jquery-1.7.2.min.js
jquery-1.9.1.min.js
jquery-2.2.0.min.js
jquery-calendar.css
jquery-calendar.js
jquery-ui-1.7.1.custom.min.js
jquery-ui-1.8.21.custom.min.js
jquery-ui-1.8.6.custom.min.js
jquery-ui.js
jquery.1.3.2.js
jquery.autocomplete.pack.js
jquery.bgiframe.min.js
jquery.cookie.js
jquery.dimensions.pack.js
jquery.easydrag.handler.beta2.js
jquery.gritter.min.js
jquery.js
jquery.livequery101.js
jquery.treeview-1.3
jquery.treeview-1.4.1
jquery121.js
jscolor
jscolor-1-4-5
knockout
qtip
SearchHighlight.js

in this directory (interface\tags_filters\assets)  the following assets are listed

jQuery JavaScript Library v1.9.1 (multiple copies)
Bootstrap v3.3.6
Bootstrak editable - v1.5.1
data_table
jquery.datetimepicker
jquery.validate
moment.min.js
bootstrap3-typeahead.js
Select2 4.0.2
jQuery Migrate - v1.1.0
RequireJS 2.1.5
composer (also in /library/vendor)

I think we should decide on a directory and move all the stuff there to cut out the replication

aethelwulffe:
How does this sound?:

Make a location.

Start moving them one-by-one (or more) with a commit, tests, and search 
results posted for each commit.

Immediately, you could create that location (or re-use library/js) for 
the portal.

I tend to prefer migrating to a new directory name because when you move 
it /and/or upgrade the version/ and you have not checked everything out, 
something doesn't work...and we find out about it

teryhill:
[...]

PR #415, opened 2017-02-11
initial commit for assets 

teryhill:
initial commit. making sure I am following the correct guidelines. Where do we want things like Knockout, bootstrap, backbone,etc. I understand the jquery's go under js but not sure about the others #372


aethelwulffe:
Should we change /js to /dependencies or /lib?  I have the funny feeling 
that most of the dependencies are in fact going to be javascript, but do 
they really need to be separated by type if they are something else?  I 
think /js is a pretty familiar concept.  Since Knockout, Bootstrap, 
Backbone, Gritter, Jquery, Ajax and all that are all javascript, it 
seems to take care of everything.  (?)

teryhill:
I don't have an issue with putting them under js .

aethelwulffe:
I suppose there was already talk about using the global variables to 
determine what version of a dependency should be used etc... so maybe 
including the old packages is part of the plan.  Please remind me of 
what was discussed on the phone call.

aethelwulffe:
My understanding of JQUERY:

Version 1.x supports ancient I.E.

Version 2.x has the same API as 1.x, while dropping the old IE 
compatibility.

Version 3.x works a little differently.

JQuery-Compat 3.x takes over the job of the 1.x for compatibility of 
IE8, Opera12, and Safari5 and other junk.

JQuery 3.x does not have the compatibility bits.

Both 3.x varieties use the same public API, and otherwise match features 
by version.

https://github.com/jquery/jquery-migrate/  is a tool we can use to 
identify and migrate jquery versions.

In my opinion, we should be able to test everything with this, upgrade 
code where needed and all that.

teryhill:
I think we should, since we are beginning this , put the latest versions in the asets dir and migrate

aethelwulffe:
I will post the above to a jquery upgrade issue.
Please look up the difference between the slim and regular versions of the standard and compatibility versions (4 flavors) and pull down the .min versions of each for this commit.
I will get pieces into play to get the upgrade moving in a more professional manner than the trial and error method.

teryhill:
sounds like a plan

Issue #416, opened 2017-02-11
JQUERY UPGRADE


aethelwulffe:
My understanding of JQUERY:

-  Version 1.x supports ancient I.E.
-  Version 2.x has the same API as 1.x, while dropping the old IE compatibility.
-  Version 3.x works a little differently.
-  JQuery-Compat 3.x takes over the job of the 1.x for compatibility of IE8, Opera12, and Safari5 and other junk.
-  JQuery 3.x does not have the compatibility bits. 
- Both 3.x varieties use the same public API, and otherwise match features by version. 

https://github.com/jquery/jquery-migrate/ is a tool we can use to identify and migrate jquery versions. In my opinion, we should be able to test everything with this, upgrade code where needed and all that.

Tasks:

- [x] Place the package from https://code.jquery.com/jquery-3.1.1.min.js into the assets/js directory.
- [x] Add the jquery_current directory to globals.
- [x] Identify all uses of the deprecated jquery functions listed below.
- [x] Modify all instances of deprecated API calls to use the latest versions.
- [x] Create and retain a maintenance/change log for JQUERY.

Check off deprecated functions as they are searched for and upgraded.

1.3
- [x] jQuery.boxModel
- [x] jQuery.browser

 1.7
- [x] deferred.isRejected()
- [x] deferred.isResolved()
- [x] .die()
- [x] jQuery.sub()
- [x] .live()
- [x] .selector

 1.8
- [x] .andSelf()
- [x] deferred.pipe()
- [x] .error()
- [x] .load()
- [x] .size()
- [ ] .toggle()
- [x] .unload()

 1.9
- [x] jQuery.support
 
 1.10
- [x] .context

 3.0
- [x] .bind()
- [x] .delegate()
- [x] jQuery.fx.interval
- [x] .unbind()
- [x] .undelegate()

teryhill:
I sent up 3.1.1 and the migrate tool. Do you want to send the ones that are currently used in portal and upgrade them later?

teryhill:
Any particular global name you want ?

aethelwulffe:
No opinion.  Make it unambiguous. (means "talk plain").

teryhill:
so calling it $globals['the stuff that does the cool stuff] is probably out huh?

$GLOBALS['js_path']
$GLOBALS['images_path']
$GLOBALS['css_path']
$GLOBALS['fonts_path']

aethelwulffe:
[...]

PR #417, opened 2017-02-11
These are the assets used in portal

teryhill:
The blink is for PFB we will look at removing jquery 1.5 1nd 1.11 #416


teryhill:
just maybe the jquery 1.5 and 1.11 along with 1.19 and 2.0 as soon as things are refactored. If we are going to run on tablets we need some of the stuff.

teryhill:
The current asset directory adds 15.2 meg to the overall project. Not a huge increase in my opinion

teryhill:
The project is still 100 meg smaller than the O

aethelwulffe:
Well, I am pretty sure that when this is done, a lot more will get snuffed than you just added.
There shouldn't be any issues with merging this (protests) as we all seemed to have a good idea of what is wanted, so I think we should keep moving and let this get to the test point.

Issue #419, opened 2017-02-11
Bring in new Misc_billing_form in to Libreehr


teryhill:
Bring in new Misc_billing_form in to Libreehr.

PR #421, opened 2017-02-12
Added the link to the assets. Dropdown select for state.

teryhill:
Added the link to the assets. Dropdown select for state. #419

### DIFF
--- a/interface/usergroup/addrbook_edit.php
+++ b/interface/usergroup/addrbook_edit.php
@@ -87,7 +87,7 @@ td { font-size:10pt; }
   // 2 = Person Centric
   // 3 = Company Centric
   $sql = sqlStatement("SELECT option_id, option_value FROM list_options WHERE " .
-   "list_id = 'abook_type'");
+   "list_id = 'abook_type' AND activity = 1");
   while ($row_query = sqlFetchArray($sql)) {
    echo "type_options_js"."['" . attr($row_query['option_id']) . "']=" . attr($row_query['option_value']) . ";\n";
   }
@@ -141,6 +141,7 @@ td { font-size:10pt; }
   $form_fname = invalue('form_director_fname');
   $form_lname = invalue('form_director_lname');
   $form_mname = invalue('form_director_mname');
+  $form_suffix = invalue('form_director_suffix');
  }
  else {
   // Person centric
@@ -148,6 +149,7 @@ td { font-size:10pt; }
   $form_fname = invalue('form_fname');
   $form_lname = invalue('form_lname');
   $form_mname = invalue('form_mname');
+  $form_suffix = invalue('form_suffix');
  }
 
   if ($userid) {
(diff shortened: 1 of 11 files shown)


teryhill:
Yep I can add suffix. @tmccormi is doing the date picker changes in his big commit don't want to mess him up.

aethelwulffe:
Gotcha.

We are going to need suffix, valedictory or something like that on the 
patient reports (for export, record requests, etc...) as well.   Folks 
are having to add stuff to their last name, which screws up billing.

teryhill:
image

teryhill:
I just pushed a commit with the changes to the address book and the misc billing form. I can get it to display in users but not to save. I will work on that next week, but right now You can add the suffix in the address book

interface/usergroup/addrbook_edit.php
```diff
@@ -87,7 +87,7 @@ function invalue($name) {
   // 2 = Person Centric
   // 3 = Company Centric
   $sql = sqlStatement("SELECT option_id, option_value FROM list_options WHERE " .
-   "list_id = 'abook_type'");
+   "list_id = 'abook_type' AND activity = 1");
```
tmccormi: Does this mean Active users are the only thing that can be selected here?  or does activity=1 mean something else?

tmccormi: Oh, nevermind this is list options type selector, not user active.   Is "activity" really the field name, does not seem correct.

aethelwulffe: I think it means "not deleted".

tmccormi: I wish developer would use nouns that actually have some relevance to the use to the value...

Activity = "was there any activity on the thing in the past?"
Active = "Is this thing temporarily available for use or NOT"
Deleted = "Is this thing permanently disabled for users"

aethelwulffe:
Look, "activity" is used all over the damn place, and most of the time it is underused, duplicated, or not used at all.  By all means start an issue that will diverge the database over this.  I love that stuff.  I can make it happen.

teryhill:
What's the trick to getting the new field in user_admin to save. I have the field there it displays the information from the table but it will not save to the table.

aethelwulffe:
I don't know at the moment.  I have always just used the address book 
interface to make modifications.  If you look at Ken's facility ACL 
code, there will be something there, but it will likely be a very odd 
multi-select thang wrapped up in some other stuff.  I can see if I can 
tell you anything later this afternoon.

teryhill:
The address book works for that. I could make the field read only in users or just leave it out and let them make the change from address book

aethelwulffe:
We are going to want to modify the user interface pretty soon 
anyway...like getting rid of the UPIN and other things.  Does that user 
interface fun off of the layout engine in any way?

teryhill:
I am pretty sure it does not use layout's.

teryhill:
I can go ahead and drop the Upin if that is needed

aethelwulffe:
User_admin and user_admin_add and all that have a lot of duplication and kind of spaghetti stuff going on.

aethelwulffe:
Yeah, UPIN was discontinued ten years ago...so yeah.  Doesn't have to be 
this commit, but should be on a list at least, and taken out of the DB.

aethelwulffe:
Terry, do you want to call this commit done, and put the other items on a different work-list?

teryhill:
Yes lets and work on the other stuff later just make sure it gets on the List

teryhill:
Did some cleanup and removed the ability to enter a upin. left it there in the table till we remove the checks for it in the billing side. I think this is mergeable after some one else tests it.

teryhill:
@aethelwulffe @tmccormi This is ready from my stand point. Just need y'all to look at it.

Issue #423, opened 2017-02-15
fixed an inconsistency reported by Harley in the PFB


teryhill:
fixed an inconsistency reported by HT. in every other instance the word
Category is used except on the main PFB screen it is called visit type,
changed it to category to be in line with the code base

PR #424, opened 2017-02-15
fixed an inconsistency reported by HT

teryhill:
fixed an inconsistency reported by HT. in every other instance the word
Category is used except on the main PFB screen it is called visit type,
changed it to category to be in line with the code base #423


teryhill:
it is called category in the add_edit_event file and as appointment categories in the calendar setup. Hence the change

PR #425, opened 2017-02-16
Changes to work correctly with global

teryhill:
Harley noticed that the flow board was not properly following the global variable "Providers See Entire Calendar". We discussed and made the changes to follow the setting.  #423


tmccormi:
Not sure if this is good or not.   What if they what to see the Calendar for general overview but not have their daily flow/process mixed with other providers work?

teryhill:
it follows what the calendar currently does. If you have multiple providers and don't want them seeing each other patients you turn off the global. If you want them to see the entire calendar you turn it on. Then they can select which provider they want to see in PFB

teryhill:
The current issue is that it did not block the providers when it should have. They can still turn it on and see everything.

tmccormi:
OK, I'll bite

PR #434, opened 2017-02-28
Add collapsing behaviour to tags in demographics

### LINKED ISSUE #405: tags need to have collapse and expand on the demographics page
tags need to have collapse and expand on the demographics page. this will maintain the current look, all others have the collapse and expand dialog . @ken

azwreith:
Fixes #405 

I think the demographics page can be further refined by modifying the `expand_collapse_widget` function to accept an array of Buttons and using that for the tags and other stuff in the future.

Also, I noticed a tiny bug: When you have no tags and you click on "Add" and then "Cancel", the "No tags found" message disappears. It's replicable on the online demo. Not sure if there's an issue already created for it.

Thanks!

### DIFF
--- a/interface/tags_filters/views/tags_demographics.php
+++ b/interface/tags_filters/views/tags_demographics.php
@@ -17,16 +17,23 @@
                            href="javascript;"
                            class="css_button_small tags-cancel-button"><span>Cancel</span></a>
                     </td>
-                    <td><span class="text"><b>Tags</b></span></td>
+                    <td>
+                        <a href="javascript:;" id="tags_container" class="small" onclick="toggleTagsIndicator()">
+                            <span class="text"><b>Tags</b></span>
+                            (<span class="indicator">expand</span>)
+                        </a>
+                    </td>
                 </tr>
                 </tbody>
             </table>
         </div>
 
         <?php if ( acl_check( 'patients', 'demo', '', 'write' ) ) { ?>
-            <div class="notab" id="TAGS">
-                <img id="tags-ajax-loading"style="padding:8px;" src="../../pic/ajax-loader.gif">
-                <div id="tags-content"></div>
+            <div id="tags_ps_expand" style="display: none;">
+                <div class="notab" id="TAGS">
+                    <img id="tags-ajax-loading"style="padding:8px;" src="../../pic/ajax-loader.gif">
+                    <div id="tags-content"></div>
+                </div>
             </div>
             <script type="text/javascript">
                 $(document).ready( function () {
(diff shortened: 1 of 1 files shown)


teryhill:
@kchapple What do you think ?

teryhill:
@azwreith   Can you show me an example of what you are thinking about on the modification to the expand_collapse_widget

azwreith:
@teryhill I was thinking of modifying the function to support multiple buttons (for eg: Tags has Add, Submit, Cancel). So we can modify the function to something like this:
```
function expand_collapse_widget($title, $label, $buttonLabel, $buttonLink, $buttonClass, $linkMethod, $bodyClass, $auth, $fixedWidth, $forceExpandAlways=false) {
...
for(i = 0;  i < $buttonLabel.length; i++) {
    // generate buttons, links, etc
}
...
```
where `$buttonLabel, $buttonLink, $buttonClass, $linkMethod` are array objects.

teryhill:
@azwreith I think that would be a good concept. Do you have a vision of what other changes to demographics would be useful or more appealing to look at by the user ( a more modern look and feel)?

kchapple:
If you're using tags at all I don't know why you'd ever want them hidden. But that's just me. 

Ken

teryhill:
@kchapple  consistency , the users use issues, demographics and other items on the screen and they can hide them.

aethelwulffe:
@Ken, I hear you, but expansion of sections is a per-user configurable item, so having the collapse vs. force open type behavior might be a good thing in some cases.  Freeing up screen space is a good option to have for specialized users.

azwreith:
@teryhill I think `Billing`, `Labs`, and `Vitals` would look good if they were to be aligned with the other text. To show you what I mean, here a screen shot:
screenshot from 2017-03-02 07-16-53

Also, wouldn't it be nice if the entire site were to use Bootstrap? That is a separate issue though ^^"

teryhill:
Yes bootstrap would be great, but it is for another day.

aethelwulffe:
Really, Demographics should go above billing for most folks, but, like we said, that is for another day.

azwreith:
@aethelwulffe I can't the permission to stop users from viewing tags anywhere :sob:

aethelwulffe:
I am sure there might be something deep deep in the TAGS code stuff somewhere that sets the permissions (I assumed that Ken might have put it right into the user admin UI), but if you can't turn them off, I bet no-one else can!  Without further worry...I wouldn't worry about it.

teryhill:
@tmccormi  this has 2 up votes

aethelwulffe:
Clean commit, clean code, tested and approved.  Merging now.  Removing the GSOC tag, as GSOC has not started...even though the hard-chargers have already started hitting this stuff!

Issue #435, opened 2017-02-28
another set of date picker changes for New patient search and add


teryhill:
date picker changes cause patient add to malfunction

teryhill:
This is not a bug. I had a setting in globals turned on that made it appear to me to be a malfunction.I had it set to mandatory fields only

PR #436, opened 2017-02-28
Working on Date picker

teryhill:
Working on Date picker #435 

I am missing something can you two @tmccormi  @aethelwulffe  look at this and see if you can see what I am missing.


teryhill:
Never mind this can be merged.

PR #450, opened 2017-03-01
Datepicker cleanup

teryhill:
last that I see until further testing. #435

### DIFF
--- a/sites/default/menu_data.json
+++ b/sites/default/menu_data.json
@@ -502,13 +502,6 @@
             "children" : [ ],
             "requirement" : 0
         }, { 
-            "label" : "Referrals",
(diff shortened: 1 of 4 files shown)


sites/default/menu_data.json
```diff
@@ -502,13 +502,6 @@
             "children" : [ ],
             "requirement" : 0
         }, { 
-            "label" : "Referrals",
```
tmccormi: This report was removed on purpose as it does not work...

teryhill: Yep It got back in. I have the report and was going to look at it and see what I could do to fix it. I removed this so there was no broken link

Issue #332, opened 2016-12-28
URL in headers for LibreHealth.io


tmccormi:
Some of the files have headers that reference www.libreehr.org

At some point they should all be converted to be the librehealth.io page instead.

aethelwulffe:
I'll SED that tomorrow.

teryhill:
Most of this is done there may be a very few left, Will change them as I find them.

PR #452, opened 2017-03-02
Changed the package and link where necessary

teryhill:
Changed the package and link where necessary #332 Doing it in smaller chunks to make review easier and prevent stomping


interface/billing/sl_eob_invoice.php
```diff
@@ -18,11 +18,11 @@
  * You should have received a copy of the GNU General Public License
```
aethelwulffe: sl_....argh!  One day we will be rid of it!

interface/code_systems/dsmiv_howto.php
```diff
@@ -15,7 +15,7 @@
  * You should have received a copy of the GNU General Public License
```
aethelwulffe: Delete this file.  It is based on the bad assumption that DSM4 codes actually exist.  It is DSM5 now anyway, and it only refers to ICD10...DSM is a guide, not a code set.

interface/code_systems/icd9_howto.php
```diff
@@ -15,7 +15,7 @@
  * You should have received a copy of the GNU General Public License
```
aethelwulffe: Snuff this file.

interface/eRxGlobals.php
```diff
@@ -15,193 +15,193 @@
  * General Public License along with this program.
```
aethelwulffe: Would live to see eRX go into the modules dir...some day.

interface/main/onotes/office_comments.php
```diff
@@ -13,7 +13,7 @@
  * You should have received a copy of the GNU General Public License
```
aethelwulffe: Do onotes work?

teryhill:
No code change just changes to link and package and reference to libreehr

teryhill:
@tmccormi  Can you review this?

PR #453, opened 2017-03-02
Add phone number formatting

### LINKED ISSUE #439: Phone number formatting in demographics eg (123)456-7890.
Phone number formatting in demographics example (123)456-7890. This should be user definable using a global variable.

azwreith:
Fixes #439 

A limitation of this code is that it assumes all formatting options to be added in the future to be in `3 3 4` , i.e, US format. An advantage of this is that people can add new formats by just including it in the `globals.inc.php` file. If need be, I can expand this code for the general case, but that model would require extra effort for people using the system to add new formats (since they'll have to write regex). Opinions?

The check for phone numbers is done (for now) using layout_options.edit_options value

EDIT: Please check lines 829-842 of global_inc.php and lines 1705-1715 of options_inc.php for the diffs

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -512,7 +512,7 @@ $GLOBALS_METADATA = array(
        'num',
        '3',
       xl('If YMD is selected for age display, switch to just Years when patients older than this value in years')
-    ),      
+    ),
   ),
 
   // Features Tab
@@ -613,7 +613,7 @@ $GLOBALS_METADATA = array(
        '0',                             // default
       xl('Set a facility cookie to remember the selected facility between logins.')
     ),
-    
+
 
     'discount_by_money' => array(
       xl('Discounts as Monetary Amounts'),
@@ -651,7 +651,7 @@ $GLOBALS_METADATA = array(
        '0',                             // default = true
       xl('This will use the custom immunizations list rather than the standard CVX immunization list.')
     ),
-      
+
     'amendments' => array (
         xl('Amendments'),
          'bool',                        // data type
@@ -660,11 +660,11 @@ $GLOBALS_METADATA = array(
     ),
 
   ),
-  
+
   // Report Tab
   //
   'Report' => array(
-    
+
     'use_custom_daysheet' => array(
       xl('Use Custom End of Day Report'),
       array(
@@ -676,7 +676,7 @@ $GLOBALS_METADATA = array(
        '1',                             // default = Print End of Day Report 1
       xl('This will allow the use of the custom End of Day report and indicate which report to use.')
     ),
-     
+
     'daysheet_provider_totals' => array(
       xl('End of Day by Provider or allow Totals Only'),
       array(
@@ -717,7 +717,7 @@ $GLOBALS_METADATA = array(
       ),
        '2',                             // default = 2
       xl('This will Display the Invoice Number in the Sales Report or the Patient Name and ID or Patient Name and Invoice Number.')
-    ), 
+    ),
 
     'cash_receipts_report_invoice' => array(
       xl('Display Invoice Number or Patient Name in the Cash Receipt Report'),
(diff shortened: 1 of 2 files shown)


teryhill:
@azwreith can you put a check in to see if the country code for the phone is set to US (Telephone Country Code in globals) so it will still work for everyone else

azwreith:
@teryhill It does work for other countries as well. The thing is that only 3 3 4 type formats can be made using this. Other countries might have formats like XXX XXXX XXX which is a 3 4 3 format and can't be applied using this code. I can put the country code check in but then it'll ignore the setting although for other countries. If that's fine, I'll put it in c:

teryhill:
I am sorry I misunderstood, as long as it still works for other countries. The country code check is not needed then.

azwreith:
Haha sorry I should have been clear about what I meant in the first place >_<

teryhill:
Not your problem I am just way old 8^)

aethelwulffe:
Let us know when you are done so that we can go over, test and approve these changes.

library/globals.inc.php
```diff
@@ -826,9 +826,22 @@
...
+      array(
+        '($1) $2-$3' => '(XXX) XXX-XXXX',   // $1: 3 digits, $2: 3 digits, $3: 4 digits
+        '($1) $2 $3' => '(XXX) XXX XXXX',
+        '$1-$2-$3' => 'XXX-XXX-XXXX',
+        '$1 $2 $3' => 'XXX XXX XXXX',
+        '$1$2$3' => 'XXXXXXXXXX'
+      ),
+      '($1) $2-$3',                    //default
+      xl('The format to display phone numbers in.')
+    )
```
teryhill: this needs to be ),

teryhill:
@azwreith  I added your code , made sure there was a "P" in the layouts. The phone number will not show on the main screen, You can enter it and it stays but is not displayed.

image

library/globals.inc.php
azwreith: Um it's the last one in the list so nuh :sweat_smile:

azwreith:
@teryhill You have to go once in the Globals settings menu, set a phone number format, and save. As the default value for phone number isn't in the database. I encountered the same issue while coding it :smiley:

teryhill:
Yep I forgot about that. It is going to be a big pain when clients start upgrading, We are discussing a way to do the upgrades that is easy to code and maintain for us and include a rollback mechanism for the clients in case there is a problem. Good work it test well. Can you come up with a way of using it for the other phone numbers in the system like Insurance, facility, address book , users ( i don't remember if that's all or not)

azwreith:
One more thing, when you click on edit on demographics page, you go to the page where you can change demographics settings. There, the phone numbers are still there in the xxx-xxx-xxxx format. I did that intentionally because it'll be easier for the person to edit their number. Here's a SS for the same:
screenshot from 2017-03-03 07-53-52
vs
screenshot from 2017-03-03 07-56-48

teryhill:
@tmccormi  @aethelwulffe I up vote this

azwreith:
Well, all I can think right now is to manually code it in for each of them. I'll brainstorm more after my morning lab! :sob:

azwreith:
Well, we could have a common global formatting function (maybe in `formatting.inc.php`) and call upon that wherever needed. And we could control that using a common global variable.

teryhill:
@azwreith  can you check the S.S.N field and see why it is updating with the phone number formatting?

azwreith:
@teryhill Ohh I just noticed that :O Lemme see why..

teryhill:
The mobile phone is not taking the change also.

azwreith:
@teryhill It is, phone numbers needs to have the standard 10 digits or else they format can't be applied to them.

teryhill:
You are correct, I had only 9 digits, some of that old guy stuff

azwreith:
@teryhill Hehe, I'm sure you're quite young and energetic :smile:

library/globals.inc.php
aethelwulffe: He got you there...Line 830.

aethelwulffe:
This stomps on earlier unmerged commits, and they will have to be cleaned up (again), but this is simple, clean, and on the issues list.  Merging.  Shoot me later if you like. :)

Issue #455, opened 2017-03-02
Calendar does not respect the Globals Locale setting for 24hr clock.


tmccormi:
See screen shot:
![calendar-locale-times](https://cloud.githubusercontent.com/assets/285835/23526519/2b4be0f6-ff47-11e6-8b04-0b44de6675eb.png)

teryhill:
That did not work in 4.1.2(7) either

aethelwulffe:
Looks like we need a new calendar, hunh? 🤕

teryhill:
I got it just need to put it on the switch

![image](https://cloud.githubusercontent.com/assets/4655366/23530821/63583dda-ff71-11e6-8ebb-86a73a9a9adf.png)

teryhill:
You closed this too soon I have a fix for add_edit_event to follow the time global also

teryhill:
here it is

![image](https://cloud.githubusercontent.com/assets/4655366/23532852/fb7387c8-ff7b-11e6-883a-62cb219979f5.png)

PR #456, opened 2017-03-02
Fix for calendar to follow time format in globals

teryhill:
Fix for calendar to follow time format in globals #455 

I need to look at add_edit_event and see what changes might be needed there


aethelwulffe:
I like when I can test the meat of something by pasting some code on top of a running instance with the globals already set right!

tmccormi:
This looks good and might be a candidate for a patch to rel-101

Issue #460, opened 2017-03-03
New patient screen not saving with datepicker changes


teryhill:
@nileshprasad137  pointed out that the patient add was not saving in the latest development code.

pri2si17-1997:
@teryhill it is getting saved. Here is the screenshot of the same:
![screenshot from 2017-03-03 19-28-10](https://cloud.githubusercontent.com/assets/12952169/23553645/c04bc45e-0047-11e7-95e7-ed8aad5ee991.png)

I have first created the sample patient and then searched it using the name.

teryhill:
@pri2si17-1997 That is great. I know there are a couple of versions of the code out there, one is from the XAMPP package download (Windows) and the other is from a github download (Linux) .

Screen shot from the development code

![image](https://cloud.githubusercontent.com/assets/4655366/23554075/8fc60110-fff1-11e6-8dce-daebd7ab48f7.png)

nileshprasad137:
@teryhill You can close this issue. It works fine now

PR #461, opened 2017-03-03
Fix for screen not saving

teryhill:
Fix for screen not saving #460 I had the wrong version of jquery being loaded, I had 1.9.1 and it should be 1.7.2


teryhill:
@aethelwulffe  @tmccormi Look at this please

aethelwulffe:
What function is it that causes it to have an issue with the (slightly) newer version?  We do have a deprecation list available.

teryhill:
Don't know figured that was on the jquery list 8^)

aethelwulffe:
OK...make it work, then we will deal with the JQUERY list. :)

aethelwulffe:
Review issue deferred to the jquery cleanup project that will likely be a GSOC project.  This is a one-line fix for something that is totally broken at the moment.

PR #463, opened 2017-03-03
Database cloning Turbo Mode

aethelwulffe:
" --opt --quote-names -r $backup_file " .
//" --opt --skip-extended-insert --quote-names -r $backup_file " .
Comment out above line and enable this to have really slow DB
duplication.
This change speeds up cloning of the database for backup or launching a
new site from a template DB by a factor of about 600.
  This is used in production, and has launched ~45 production sites dependably.

### DIFF
--- a/library/classes/Installer.class.php
+++ b/library/classes/Installer.class.php
@@ -567,7 +567,8 @@ $config = 1; /////////////
     $backup_file = $this->get_backup_filename();
     $cmd = "mysqldump -u " . escapeshellarg($login) .
       " -p" . escapeshellarg($pass) .
-      " --opt --skip-extended-insert --quote-names -r $backup_file " .
+      " --opt --quote-names -r $backup_file " .
+      //" --opt --skip-extended-insert --quote-names -r $backup_file " .  Comment out above line and enable this to have really slow DB duplication.
       escapeshellarg($dbase);
     
     $tmp0 = exec($cmd, $tmp1=array(), $tmp2);


teryhill:
This is what y'all did for suncoast PQRS right?

aethelwulffe:
Yes.  This is essentially a squashed @leebc commit.

azwreith:
How do we select the option for cloning the DB? I don't see it while installing :sob:

aethelwulffe:
Unfortunately, we removed the "multi-site" feature (admin.php) from the system (it was a security risk as it existed).  Right now, unless this is called from database backups, I don't think there is an interface that allows you to use it without some hoops to jump through.  The only way currently to create a second site is to duplicate a blank "default" directory into the sites directory, then attempt to log into that with site="newsite" in the URL...which then (I hope) will give you an install prompt to "duplicate existing database".  This is where this bit comes in.

tmccormi:
Setup allows you to duplicate / clone a database.  You just have to run setup.php against the URL without the ?site=default tag and you get this:

setup 1

Change the site name from default to something else [new-site]  and proceed with the setup per normal until you get to the step 2 page.  On the page there is a selector and check box for what to clone.

setup-2

aethelwulffe:
@tmccormi Should we just include this in the documentation (as a link) in a special segment named "Adding Sites an Additional Databases" or something, or should we pull in an admin.php replacement to allow an "Installer" level admin access to a sites list and give them the ability to do this with a UI?

tmccormi:
Yes, that would be good.  Also I plan on pulling a fixed up version of the AutoInstaller.php command line tool back into the project.

aethelwulffe:
@tony  We really ought to get leebc involved in that discussion as well.  He may have some BASH that he could share around for multi-site upgrades and the like.  He scripts almost everything he does.  I think he has a bash version of the Quake 3D game engine around somewhere... 😈

tmccormi:
Ok with me, though I have no idea who leebc is ..

aethelwulffe:
That is my Bryan...and a GitHUB user on repos where you have a seat...

aethelwulffe:
So, do I drop this pull request, is it expected that I add something else to it first, or should we look at it for a few more weeks? 🙊

leebc:
Hi, I'm leebc.   ((I'm not a database guy))

When you use the flag **--skip-extended-insert**, **mysqldump** generates a .sql file with EACH insert on a separate line.  

It was my experience that cloning a new database, that is to say one that has the database structure and some code tables but no customer data in it, was taking about 45 minutes.

When I removed **--skip-extended-insert**, the time dropped to 45 SECONDS.  That's a speedup of 60 times!

There is a risk here:
If your database contains a LOT of data, these inserts could become too large to process.  We've run into that problem when importing masses of billing data generated by an external program we use.  (But not through **mysqldump**.)
There are some other comments on forums about having problems specifically when importing large blobs of data.

So there's a trade-off:
"Faster performance" vs. "risk a failure because a customer is cloning a large, populated database".


For what Art and I are doing, (cloning a minimal site default configuration dozens of times and we know how to fix it,) there is no question, WE want it faster.

-----
**mysqldump** options note:

According to the documentation, the **--opt** option includes **--extended-insert**  and it can be disabled by **--skip-extended-insert**, which is what happens in the OpenEMR code.

leebc:
I do have a *bash* script that will read multiple .sql files and apply them to each database used by an entity in the sites/ directory.

I think I played Quake or Doom on a text console once. ;-)

PR #465, opened 2017-03-03
Turn NationNotes into a module

aethelwulffe:
Commit 1 Tested without turning the module on.  Passes all checks, but
need to implement $GLOBALS['modules_dir'] vs relative paths and
$GLOBALS['mod_nn'] as an IS_SET and for value.


aethelwulffe:
Layout engine works, everything that calls changed files with the module OFF has identical log errors and warnings that it did before...

modules/nation_notes/nn_acl_upgrade.php
```diff
@@ -0,0 +1,56 @@
+<?php
+/**
+* Add GACL sections for Nation Notes.
+
+* @package   LibreEHR
+* @author    Art Eaton <art@suncoastconnection.com>
```
teryhill: LibreHealth EHR Is what I think we are calling it now

modules/nation_notes/nn_acl_upgrade.php
```diff
@@ -0,0 +1,56 @@
...
+* @link      http://LibreHealth.io
+*/
+
+$ignoreAuth = true; 
+
+require_once('../../interface/globals.php');
```
teryhill: Since this is used all over the code base should ve come up with a variable like $srcdir?

modules/nation_notes/nn_lbf_new.inc
```diff
@@ -0,0 +1,635 @@
...
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://opensource.org/licenses/gpl-license.php>.
+ *
+ * @package LibreEHR
```
teryhill: LibreHealth EHR

modules/nation_notes/nn_super_edit_layout.inc
```diff
@@ -0,0 +1,1718 @@
...
+ * as published by the Free Software Foundation; either version 2
+ * of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://opensource.org/licenses/gpl-license.php>.
+ *
+ * @package LibreEHR
```
teryhill: LibreHealth EHR

modules/nation_notes/nn_acl_upgrade.php
aethelwulffe: I added the GL value and a variable into globals.inc.  I also cleaned up the ZEND modules stuff, and re-organized them to a location in the code that makes more sense (like where we are making path globals).
Next round will be me removing these relative paths and using the global variables.

modules/nation_notes/nn_acl_upgrade.php
aethelwulffe: Gotcha on all these.

interface/forms/LBF/new.php
```diff
@@ -235,21 +240,16 @@ function end_group() {
...
-<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/jquery.1.3.2.js"></script>
-<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/common.js"></script>
-<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/fancybox/jquery.fancybox-1.2.6.js"></script>
-<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/jquery-ui.js"></script>
-<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/jquery.easydrag.handler.beta2.js"></script>
 <script type="text/javascript" src="../../../library/textformat.js"></script>
 <script type="text/javascript" src="../../../library/dynarch_calendar.js"></script>
 <?php include_once("{$GLOBALS['srcdir']}/dynarch_calendar_en.inc.php"); ?>
 <script type="text/javascript" src="../../../library/dynarch_calendar_setup.js"></script>
+<script type="text/javascript" src="../../../library/js/jquery.js"></script>
```
aethelwulffe: is jquery.js 1.2.1?  I think you are right.  In any case, these should go ahead and be moved to the assets directory (as duplicates I suppose) right?

aethelwulffe:
Basically, all the Nation Notes bulk includes that were simply dumped whereever were moved (as a lump) to the nation notes module directory.  No effort was made to actually maintain the nation notes code at all, just get it into position so that it can all be tracked down, and easily removed without harm if someone wishes to do so.  It can also be left out of the main package, in which case the hooks and globals amount to just a few added lines in each file that are still less than the original code contained.  In total, about 500 files are removed completely without affecting function, and several hundred more can be removed with some careful experimentation.

aethelwulffe:
Requesting additional reviewers.

tmccormi:
This branch has conflicts and can't be merged yet ...

aethelwulffe:
Yeah, I remarked that the other merge would create a conflict with datatype 34....Fixed it, and it is good to go now.

Issue #466, opened 2017-03-03
Deceased Date not working with date picker


teryhill:
Deceased Date and Financial review date not working with new date picker

aethelwulffe:
This fix is in, simple and clean.  Merging #467 and closing this issue.

PR #467, opened 2017-03-03
Fix for deceased date and financial review date

teryhill:
Fix for deceased date and financial review date #466


teryhill:
@tmccormi @aethelwulffe  Review requested

Issue #457, opened 2017-03-02
iFrames need to be resizable


aethelwulffe:
We need to be able to re-size the iframe divs, and the tab labels need to follow them:
A resource for using css or other methods for doing this (some of these solutions will not work for our application:
http://stackoverflow.com/questions/8117761/how-can-i-make-an-iframe-resizable

azwreith:
I can give this a shot!

aethelwulffe:
You rock man.  GO!GO!GO!

teryhill:
I have tested this and it looks and works food with 3 or less tabs open. 4 tabs it is no longer re-sizable on my screen (laptop) Good work @azwreith

aethelwulffe:
Terry, how about doing the review on this and approve it.  I have not tested it.  I imagine that 4 tabs hits a minimum size or something anyway...
I will go for whatever you think...I just want to see it merged!

azwreith:
Yes, it's not re-sizable with 4 tabs because I set the minimum width of each tab to be 20%. That can be changed though

azwreith:
This can be closed now, unless we want to keep it open for future fixes :smile:

aethelwulffe:
Darn good job.  Closing this issue.

teryhill:
@azwreith  fixes can be on there own issue. This is a great addition to the code base. @pri2si17-1997  @nileshprasad137  and @azwreith  I sincerely hope y'all stick around after GSOC we can use all the help we can get.

nileshprasad137:
@teryhill Surely, I will!

pri2si17-1997:
@teryhill Sure...

pri2si17-1997:
@teryhill, i'm not able to access chat forum. It is showing '502 Bad Gateway'. Please help me.

aethelwulffe:
Chat is down at the moment...for all of us.

teryhill:
It is down I have sent a request for assistance.
On Monday, March 6, 2017, 1:03:06 PM EST, pri2si17-1997 <notifications@github.com> wrote:
@teryhill, i'm not able to access chat forum. It is showing '502 Bad Gateway'. Please help me.

aethelwulffe:
Ditto.

azwreith:
I sure will, @teryhill :blush: 

Good night y'all!

teryhill:
Good night
[...]

Issue #420, opened 2017-02-11
Glory Road UI


aethelwulffe:
Goals:
1.  Optimize vertical space.
2.  Make all toolbar elements fully interactive, minimizing the need for navigation or context labels.
3.  Organize menu cascades and toolbar layout to minimize navigation steps.
4.  Group related toolbar elements for readability and to minimize navigation errors.

![gloryroad](https://cloud.githubusercontent.com/assets/553751/22858255/a9869d74-f085-11e6-9507-6454616183c0.png)

Functions (Left, Right, or Centered display):
[CLEAR] -Clear Active Patient.  Removes patient from the session data.
[RESET VIEW] -Reset to default user view.
[PATIENT NAME and DEMOGRAPHICS FIELD(s)] -Display configured patient name values.  Clicking on any element in this acts as the active patient search, and will display a selection drop-down, obviating the need to go to the full patient search tab.
[LATEST ENCOUNTER] -This field lists the most recent encounter for the active patient.  This field's format is configurable to remove the label text.
[CURRENT ENCOUNTER] -This field (when interfaced) provides the Encounter History dropdown.  If the currently selected encounter is NOT the most recent, the field displays an alert (colorization, "OLD", or whatever is configured as the user preference).
[STATUS] -This is the flowboard status.  It is interactive to allow status updates.  The full status, in the form of a series of small icons or the like could take place of this, changing from red to green as the steps are progressed through.
[TAB MENU ITEMS] -JSON menu items.  The behavior of how many tabs, possibly ordering, and any number of other things related to this should be placed in a user configuration file.
[USER] -Displays user session data.  This also produces the "Preferences, Logout" dropdown.

[TABS] -iframes are resizable, and the tabs follow the frame.

teryhill:
I like it. You want me to add Dr Ray's role based code to it?

aethelwulffe:
WEll, There ain't no code as yet, just this picture, Copyright Art Eaton 2017

tmccormi:
Other that the patient name/data being the "search" button it's OK, 
Thoughts:
Latest should be ACTIVE ENCOUNTER, latest implies most recent but not necessarily the working encounter.

I think CURRENT ENCOUNTER is just confusing, Maybe button called "Enc History"?  what does it open, same as now a single past encounters? or does it open the View History Tab?  
Should we differentiate between finding a single encounter to see/edit/sign for the past on a pick list and choosing to do that from broader view of the encounter history?

[...]

PR #470, opened 2017-03-04
Add frame resizing capability

azwreith:
Hey guys! Check out this trial version. I was initially having problems due to mouse movements being blocked by iframes but I solved it using a transparent layer whenever we are resizing

It goes a little bit crazy when 3 frames are open so I'll have to modify the formula a bit for that.

I didn't use JQueryUI because importing that whole thing for such small use seems like a waste of resources.

Thanks!

### DIFF
--- a/interface/main/tabs/css/tabs.css
+++ b/interface/main/tabs/css/tabs.css
@@ -3,7 +3,7 @@ To change this license header, choose License Headers in Project Properties.
 To change this template file, choose Tools | Templates
 and open the template in the editor.
 */
-/* 
+/*
     Created on : Jan 22, 2016, 2:13:40 AM
     Author     : yehster
 */
@@ -60,7 +60,7 @@ body
     flex-direction: row;
 }
 
-#framesDisplay > div
+.frameDisplay
 {
     flex: 1 0 auto;
     align-items: stretch;
@@ -76,6 +76,25 @@ body
     flex: 1 0 auto;
     border: 1px solid black;
 }
+
+#frameBarrier {
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
+    bottom: 0;
+    z-index: 10;
+    visibility: hidden;
+}
+
+.handle {
+    width: 5px;
+    text-align: center;
+    z-index: 11;
+    transition: all ease-in 0.1s;
+}
+
+.draggable { background: #BBCCFF; }
 
 .tabControls
 {
@@ -114,7 +133,7 @@ body
 
     width: 99%;
     height:99%;
-    
+
     background-color: gray;
     z-index:1000;
     opacity: 0.5;
@@ -249,4 +268,3 @@ div.closeDlgIframe {
 {
     color:red;
 }
-
--- a/interface/main/tabs/css/tabs.css
+++ b/interface/main/tabs/css/tabs.css
@@ -62,7 +62,7 @@ body
 
 .frameDisplay
 {
-    flex: 1 0 auto;
+    flex: 100 0 auto;
     align-items: stretch;
     align-content:stretch;
     display: flex;
--- a/interface/main/tabs/main.php
+++ b/interface/main/tabs/main.php
@@ -1,6 +1,6 @@
 <?php
 Use Esign\Api;
-/* 
+/*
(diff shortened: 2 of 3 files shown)


azwreith:
Addresses #457

teryhill:
There are versions of UI on the system. What version do you want to use? This is what I get for the inital screen using chrome

image

When I try to resize it I get this with one click and it will not return to the previous screen size.

image

azwreith:
Well, on the main page we are using JQuery 2.2.0. A compatible JQueryUI would be 1.12.1 I guess. The latest one we have, 1.8.21, was giving errors with the JQuery version.
Anyway, I think this is a better approach, though the code is bigger.

Hm..that's strange, can you check whever you get the `handle` div in the center? It should look something like this:
image

teryhill:
@azwreith  Does this help?

image

azwreith:
Yes it does! Ahh, it seems like you still have the old tabs.css file

teryhill:
The original tabs.css is on the left the change is on the right

image

azwreith:
Oh I got it! Chrome is loading the css from the cache so you gotta open dev tools (Ctrl+Shift+i) and then right click the refresh button (next to search bar) and click "Empty Cache and Hard Reload"

teryhill:
sometimes I hate cache. That was it sorry I usually do a refresh just been putting in some extra hours trying to stay up with you young guns.

teryhill:
I like that change. Can you set it up on a global to turn it off or on?

azwreith:
Yes yes, cache has destroyed all our lives xD
Well, you deserve a nice relaxing weekend.

So how's it looking?

On Mar 5, 2017 4:41 AM, "Terry Hill" <notifications@github.com> wrote:

sometimes I hate cache. That was it sorry I usually do a refresh just been
putting in some extra hours trying to stay up with you young guns.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/LibreHealthIO/LibreEHR/pull/470#issuecomment-284191720>,
or mute the thread

teryhill:
Look at #420 I think you can handle that one also 8^)

teryhill:
Thanks but I still have code to do for my customers so the weekend is still very busy. I do take Sunday off.

teryhill:
can you look at opening 3 and 4 frames. I don't think the screens size is good with more than 4 open, What do you think?

teryhill:
@azwreith  here's another thing, if the all tabs are unlocked the active tab only take's up the original size, can you make it go full frame when only one frame is active? Try it with 3 frames unlocked and you will see what I am talking about.

image

aethelwulffe:
Hint:  Browser options: "delete all history when closing firefox."

aethelwulffe:
Terry, give this a review or approval please.  This should be merged before some of the other pieces, and I don't want to see it stomped on.

aethelwulffe:
This still has the issue of needing to "spring" back to take up the unused space...Is that a usability issue?

azwreith:
Hey guys! I was out all day yesterday. The jumping with 3 tabs and the tab not taking up the either frame issues are caused because of the formula, which needs to be modified. I'll push a commit with the changes today :3

aethelwulffe:
Glad to hear it Ujjwal.  This is a LONG awaited improvement.  It is probably the most important piece of the tabs system that was missing.

aethelwulffe:
Oh heck yeah.  I am testing NOW!

azwreith:
I updated the formula to accommodate tabs springing back when alone. It's still a bit jittery but works :smile:

aethelwulffe:
It does have a bit of a jitter, but not real bad.  That behaviour is complex though, and unless we slow down the change with some sort of smooth transition, It will likely be that way.  It looks good to me, and I am hitting my thumbs up.

teryhill:
Looks good in testing. Thanks @azwreith

aethelwulffe:
@teryhill if you have given this a final test, I would like to merge it.

aethelwulffe:
Oh, I see you did.  OK, I will merge.

teryhill:
@aethelwulffe Merge it and we will give it a shake down over the course of development. Also merge @nileshprasad137  stuff and get it in

PR #471, opened 2017-03-05
misc billing date and encounter date

teryhill:
misc billing date and encounter date


aethelwulffe:
I merged this so as to not hold up Terry's ongoing work.

Issue #428, opened 2017-02-22
Customer requested that ordering, referring and a user defined field be added to fee sheet


teryhill:
Customer requested that ordering, referring and a user defined field be added to fee sheet. The user wants the user defined field to be a changeable label that they can make their own. They have interns and contractors that they use. The label changing is for different sites one which has the interns and the others use contractors. This field just stays in the encounter form it does not go to HCFA or 837.

aethelwulffe:
Did you get the commit for this in?  I saw the code screenshots, and I would like to add to it

teryhill:
Not yet I have it working and just need to test and make sure I got all the HCFA files done.

teryhill:
@aethelwulffe  just sent a pull request

PR #473, opened 2017-03-06
Adding Ordering, referring and contract physician to fee sheet

teryhill:
Adding Ordering, referring and contract physician to fee sheet #428.

In the HCFA 1500 Print this only changes the referring provider. If a box 17 provider is supplied on the misc billing form it is used. If there is no misc billing box 17 information, the fee sheet is checked for referring then the patient data field.

They are separate fields in the electronic form so you can have an ordering and a referring.

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -8,7 +8,7 @@
 // of the License, or (at your option) any later version.
 // Modified by dozens of contributors
 
-/* @package LibreEHR
+/* @package LibreHealth EHR
  * @author Rod Roark <rod@sunsetsystems.com>
  * @author Tony McCormick <tony@mi-squared.com>
  * @author Terry Hill <terryhill@librehealth.io>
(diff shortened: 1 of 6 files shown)


library/globals.inc.php
```diff
@@ -961,6 +961,48 @@
...
+
+    'contract_physician_in_feesheet_name' => array(
+      xl('Label Title for Contract Physician on Fee Sheet'),
+      'text',                           // data type
+      'Contractor',
+      xl('Label Title for Contract Physician on Fee Sheet')
+    ),
+    
+    'allow_appointments_in_feesheet' => array(
+      xl('Allow Schedualing Appointments from Fee Sheet'),
```
tmccormi: Spelling issue -  Sched - Dualing sounds dangerous

library/globals.inc.php
```diff
@@ -961,6 +961,48 @@
...
+      'bool',                           // data type
+      '0',                              // default = false
+      xl('Allow Schedualing Appointments from the Fee Sheet.')
```
tmccormi: Same here

library/globals.inc.php
teryhill: Fixed

library/globals.inc.php
teryhill: Fixed

PR #476, opened 2017-03-07
Fix for nation notes change missing curly brace in options.inc

teryhill:
Fix for nation notes change missing curly brace in options.inc good catch @pri2si17-1997


aethelwulffe:
Good catch.  Looks like it crept in with the conflict resolution.

aethelwulffe:
This is tiny, and is obvious.  Merging.

Issue #196, opened 2016-09-20
CSS error list


aethelwulffe:
Expected color but found 'none'.  Error in parsing value for 'background-color'.  Declaration dropped.style_light.css:1007:20
Unknown property 'speak'.  Declaration dropped.typicons.min.css:1:368
Unknown property 'font-smoothing'.  Declaration dropped.typicons.min.css:1:521
Error in parsing value for 'cursor'.  Declaration dropped.ajax_calendar.css:17:29
Error in parsing value for 'cursor'.  Declaration dropped.ajax_calendar.css:138:29
Unknown property 'zoom'.  Declaration dropped.demo_table.css:36:5
Expected declaration but found '_'.  Skipped to next declaration.demo_table.css:82:1
Expected declaration but found '_'.  Skipped to next declaration.demo_table.css:157:1
Expected declaration but found '_'.  Skipped to next declaration.demo_table.css:267:1
Unknown property '-moz-border-radius'.  Declaration dropped.demo_table.css:332:19
Expected declaration but found '_'.  Skipped to next declaration.demo_table.css:336:1

pri2si17-1997:
Hello @aethelwulffe, can you please tell the steps to reproduce this?

aethelwulffe:
I think you got them.

PR #482, opened 2017-03-09
Fixed css property.

pri2si17-1997:

### DIFF
--- a/interface/themes/ajax_calendar.css
+++ b/interface/themes/ajax_calendar.css
@@ -14,7 +14,8 @@ div.tiny { width:1px; height:1px; font-size:1px; }
     cursor: default;
 }
 #printview, #dayview, #weekview, #monthview {
-    cursor: pointer; cursor: hand;
+    cursor: pointer; 
+    cursor: hand;
 }
 #functions {
     width: 13.5%; /* can't make this 25% for some reason */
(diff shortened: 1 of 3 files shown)


pri2si17-1997:
#196 

Hello @teryhill @aethelwulffe I have made some changes in the css  file mentioned in above issue. Please check it let me know if anything needed.

teryhill:
@pri2si17-1997  What does this change exactly? You are working with Old Guy's 8^)

pri2si17-1997:
@teryhill I am not sure what this change will do. But when I looked at issue #196, I browsed to those files. There I found syntax error(`*` before css property). Also i found  `zoom`  was used which is a non-standard css property, so i changed it to `transform : scale()`. So i thought to change that. 

http://stackoverflow.com/questions/4563651/what-does-an-asterisk-do-in-a-css-property-name
https://developer.mozilla.org/en-US/docs/Web/CSS/zoom

pri2si17-1997:
@aethelwulffe Yes `css/demo_table.css` called in following pages:
`/interface/tags_filters/views/layouts/patients_tags_layout.php`
`/interface/tags_filters/views/layouts/tags_layout.php`
`/interface/tags_filters/views/layouts/filters_layout.php`
`/interface/themes/rtl.css`
`/interface/main/finder/dynamic_finder.php`

aethelwulffe:
It's just an odd name, and...well...this project has a history of including unneeded files in packages.  Thank you for taking care of these irritating css errors.  There was an issue that listed every error in my logs, but that is long gone I think...

aethelwulffe:
Getting rid of asterisks is good.  The non-standard Firefox bit...so long as it is currently supported is fine I suppose, but should be commented as non-standard blah, blah, blah, if you deem it might be a future issue.

PR #484, opened 2017-03-09
General clean up Eclipse found most of these

teryhill:
General clean up Eclipse found most of these, html tags, missing commas, some syntax errors in old edi code, errors in patient portal code etc just general stuff, @aethelwulffe  I turned off the trailing blank check in my editor, but I think they are going to be a problem in the future. Still killing the tab's and replacing with spaces.


aethelwulffe:
So long as they are intentional, and not just random, it is all good by me.  Just do a codebase-wise find and replace for tabs, and replace with four (or five) spaces I suppose....

library/classes/rulesets/Amc/reports/AMC_314g_1_2_14_STG2/Numerator.php
```diff
@@ -39,5 +39,5 @@ public function test( AmcPatient $patient, $beginDate, $endDate )
 
         }
     }
-}
```
aethelwulffe: This stuff is useless at this point anyway...but...I guess fixing it doesn't hurt!

PR #493, opened 2017-03-10
Add option to disallow printing for deceased

### LINKED ISSUE #440: Change statement to allow selection of no printing if patient is deceased.
Change statement to allow selection of no printing if patient is deceased. This should be user definable using a global variable.

azwreith:
Resolves #440 

Lines 1127-1133 of `globals.inc.php` & 107,451 of `statement.inc.php`
Couldn't really test since I was having difficulty adding payments, but I think it should work :pray: 

Thanks!

### DIFF
--- a/sites/default/statement.inc.php
+++ b/sites/default/statement.inc.php
@@ -32,7 +32,7 @@ $STMT_PRINT_CMD = $GLOBALS['print_command'];
  *
  *  2.  Branded Statement, whose core is build from 1., the original statement, using HTML2PDF.
  *
- *      To customize 2., add your practice location/images/practice_logo.gif 
+ *      To customize 2., add your practice location/images/practice_logo.gif
  *      In the base/default install this is located at '/openemr/sites/default/images/practice_logo.gif',
  *      Adjust directory paths per your installation.
  *      Further customize 2. manually in functions report_2() and create_HTML_statement(), below.
@@ -99,8 +99,13 @@ function report_header_2($stmt,$direction='',$providerID='1') {
 }
 function create_HTML_statement($stmt) {
   if (! $stmt['pid']) return ""; // get out if no data
+
   #minimum_amount_due_to _print
   if ($stmt['amount'] <= ($GLOBALS['minimum_amount_to_print']) && $GLOBALS['use_statement_print_exclusion']) return "";
+
+  // Don't print if decreased
+  if ($GLOBALS['disallow_print_deceased']) return "";
+
   // Facility (service location)
   $atres = sqlStatement("select f.name,f.street,f.city,f.state,f.postal_code,f.attn,f.phone from facility f " .
     " left join users u on f.id=u.facility_id " .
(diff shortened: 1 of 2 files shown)


sites/default/statement.inc.php
```diff
@@ -32,7 +32,7 @@
  *
  *  2.  Branded Statement, whose core is build from 1., the original statement, using HTML2PDF.
  *
- *      To customize 2., add your practice location/images/practice_logo.gif
```
aethelwulffe: This should be practice_logo.png...I think.

aethelwulffe: White space is making review very difficult...

sites/default/statement.inc.php
```diff
@@ -442,6 +447,9 @@ function create_statement($stmt) {
  #minimum_amount_to _print
  if ($stmt[amount] <= ($GLOBALS['minimum_amount_to_print']) && $GLOBALS['use_statement_print_exclusion']) return "";
 
+ // Don't print if decreased
+ if ($GLOBALS['disallow_print_deceased']) return "";
```
aethelwulffe: This should probably give a message as to WHY it is not printing...."Your patient is DEAD.  You must start a new game or load a saved one!"

teryhill: if you are printing 100 statements do you want a statement printed with "Your patient is DEAD. You must start a new game or load a saved one!" stuck some where in it?

aethelwulffe:
Log output window, interface warning, print the statement with "Patient Deceased: Printing of statements for Deceased Patients disabled in ADMINISTRATION->Globals." or something other than simply not printing the statement.  I can just see the pissed off office worker trying over and over to print the same statement...not understanding what the issues is.  "I have been printing statements for this patient for 50 years!  Why doesn't it work now!!!!"
  If you guys are OK with the behavior as-is, I am cool with that.  If I am describing an edge case you don't think is worth addressing and all that.  Just giving my input.

sites/default/statement.inc.php
azwreith: This was there from before. I'll push a commit changing that!

azwreith: Sorry about the whitespace removal thingy, my editor (Atom) does it automatically. I shall look into disabling that for this project. Or maybe we can create an issue out of this and removal redundant whitespace from every file in a single PR

azwreith:
@aethelwulffe Sure, I can do that! And it makes sense too yep. Though if we are doing for this, we might as due it for other cases such as it not printing a statement when below the minimum amount, etc. What do you think?

aethelwulffe:
https://github.com/atom/whitespace/issues/10
Terry, as the originator of this issue, should have final say in the needed implementation.  He is, after all, dealing with the issue as a specific requirement from an actual user.

azwreith:
Thanks for the link! I'll disable it for now. I shall figure out how to set it on a per project basis later on ^^"
Though I think we really should run the project files though a whitespace stripper script once if possible, but that is a seperately issue on its own

aethelwulffe:
I am withdrawing the change request for a notice that you are printing for deceased.  This has two reviews, and has been here for weeks, so I am going to merge it.
Good job @azwreith

Issue #494, opened 2017-03-10
No datepicker to choose "start date" and "end date" for medical problems of patients


nileshprasad137:
Steps to reproduce this issue:
1. Choose a patient.
2. Enter the medical problems or  medications sections and click on  edit .
3. Try to enter begin date or end date. Unexpectedly, no date picker appears there.

I am working on this issue, and i'll send a commit soon to fix this issue.

PR #495, opened 2017-03-10
added datepicker for end and start date for medications of patient

nileshprasad137:
Fixes https://github.com/LibreHealthIO/LibreEHR/issues/494

screenshot 109

Please tell , if any more modifications are required!

### DIFF
--- a/interface/patient_file/summary/add_edit_issue.php
+++ b/interface/patient_file/summary/add_edit_issue.php
@@ -832,5 +832,17 @@ function divclick(cb, divid) {
 <script language='JavaScript'>
  newtype(<?php echo $type_index ?>);
 </script>
+
+<script>
+    $(function() {
+        $("#form_begin, #form_end").datetimepicker({
+            timepicker: false,
+            format: "<?= $DateFormat; ?>"
+        });
+        $.datetimepicker.setLocale('<?= $DateLocale;?>');
+    });
+</script>
+
+
 </body>
 </html>


aethelwulffe:
@tmccormi  Can we get clinician or eRX expert on this to determine if this can or should be a date-time field?  Some, many in fact, medications are administered as procedures in an office, and have time factors.

nileshprasad137:
@aethelwulffe  Do you mean along with datepicker , there should be a timepicker as well? Or there should be no begin date or end date at all?
Its possible to put a timepicker there.

aethelwulffe:
@nileshprasad137 What I am asking is if there is a need for a date-time field.  I am not sure at the moment if the field accepts datetime or not in any case.  If it is not a datetime field, then your code should simply stand as-is, because to add time would be a different feature.  In whichever case, that was only a question, and I am going to do a review as-is right now...since we got no input on that.

teryhill:
If time is needed just change this from false to true. @nileshprasad137 @aethelwulffe 

timepicker: false,

tmccormi:
The medications table is predominately for history data.  So a time would not be available.   Prescriptions are stored in the prescriptions table by the eRX and by the built in RX

aethelwulffe:
But they push to the meds list...which is actually the "issues" table, which is a bit of a mess anyway...just checking that the formats across both are compatible, and that there isn't any need for time-based therapy records (which do happen, as I have confirmed).  This should be merged up in any case.

PR #498, opened 2017-03-11
data base fix for ordering,referring and contact Physicians on fee sheet

teryhill:
data base fix for ordering,referring and contact Physicians on fee sheet


sql/database.sql
```diff
@@ -1281,10 +1281,14 @@ CREATE TABLE `form_encounter` (
   `stmt_count`        int  NOT NULL DEFAULT 0,
   `provider_id` INT(11) DEFAULT '0' COMMENT 'default and main provider for this visit',
   `supervisor_id` INT(11) DEFAULT '0' COMMENT 'supervising provider, if any, for this visit',
+  `ordering_physician` INT(11) DEFAULT '0' COMMENT 'ordering provider , if any, for this visit',
+  `referring_physician` INT(11) DEFAULT '0' COMMENT 'referring provider, if any, for this visit',
+  `contract_physician` INT(11) DEFAULT '0' COMMENT 'contract provider, if any, for this visit',
```
aethelwulffe: Did we decide to call this field x_physician vs. x_clinician?
Otherwise, this looks good.

sql/database.sql
```diff
@@ -1281,10 +1281,14 @@ CREATE TABLE `form_encounter` (
...
   `provider_id` INT(11) DEFAULT '0' COMMENT 'default and main provider for this visit',
   `supervisor_id` INT(11) DEFAULT '0' COMMENT 'supervising provider, if any, for this visit',
+  `ordering_physician` INT(11) DEFAULT '0' COMMENT 'ordering provider , if any, for this visit',
+  `referring_physician` INT(11) DEFAULT '0' COMMENT 'referring provider, if any, for this visit',
+  `contract_physician` INT(11) DEFAULT '0' COMMENT 'contract provider, if any, for this visit',  
   `invoice_refno` varchar(31) NOT NULL DEFAULT '',
   `referral_source` varchar(31) NOT NULL DEFAULT '',
   `billing_facility` INT(11) NOT NULL DEFAULT 0,
   `external_id` VARCHAR(20) DEFAULT NULL,
+  `pos_code` tinyint(4) default NULL,
```
aethelwulffe: The facility code should come from "service_facility" and not a separate POS code.  That means if you have an error in the facility info, you can change that, and all the encounters are fixed.

aethelwulffe: Make that "facility_id" and not "service_facility".  The two of these should be clearly marked in the encounter form to indicate that the facility_id field is specifically "Place of Service" and the "Billing Facility" is also clearly marked.  I believe each already has the code to list only the correct list of facilities for each dropdown.

teryhill: removed offending field. Not in the LibreHealth code anyway

sql/database.sql
teryhill: This is just a fix to get the fields for the code that has already been placed into the code base into the table

aethelwulffe:
I suppose there was no issue with you needing the POS code in the encounter.  Since the other fields are already there...and they won't kill anyone, I will approve this.

PR #505, opened 2017-03-12
Added commas in array in claim status codes

nileshprasad137:
There were no commas on multiple lines in these files which might be causing problems.
I wrote a PHP script to do so  :wink:

I Copied these lines to a text file and then ran this PHP script.
`$lines = file('test.txt');
foreach ( $lines as & $line ) {
	$line = substr($line, 0, -4);	
	$line = trim($line).'\',';
	var_dump($line);	
}
file_put_contents('filenew.txt', implode(PHP_EOL, $lines));`


nileshprasad137:
Oh !! Why this Conflict??

teryhill:
conflicts because it was fixed already. I fixed it on commit #484 on 3/10/17

aethelwulffe:
I hate conflict.
Good catch though Nilesh!  (better duplication than nothing getting done I say!)

PR #506, opened 2017-03-13
Fixed DOB of insurance subscriber not to accept future dates

nileshprasad137:
DOB of insurance provider was accepting future dates. This has been fixed now.

### DIFF
--- a/interface/new/new_comprehensive.php
+++ b/interface/new/new_comprehensive.php
@@ -658,6 +658,7 @@ if (! $GLOBALS['simplified_demographics']) {
       $(function() {
           $("#i<?php echo $i?>subscriber_DOB").datetimepicker({
               timepicker: false,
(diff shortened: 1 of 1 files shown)


aethelwulffe:
I actually recommend a js function be put in a library with arguments for date, time, future date allowed, past date allowed and whatever else you can think of.
Date, time, datetime could be a DEFAULT (0), 1, 2.  temporal attribute could be default (anything), current+, Current-, also as default(0), 1, 2 or whatever floats your boat.

nileshprasad137:
Yes , that would be much better.

aethelwulffe:
The datepicker needs to live in a function in a library, but apparently refactoring that will be another project as the implementation has been ad-hoc.  This is approved and can be merged the same as all the other date stuff...so here we go.  On to the next one.

Issue #504, opened 2017-03-12
Empty data is getting imported


pri2si17-1997:
Steps to reproduce it:
1- Create a patient
2- Then select `Popups->Import`. Click on `import patient`. It will create a alert, but also data is getting imported. (To verify this click on `Patient/Client -> Finder` you will see screen like this:
![screenshot from 2017-03-12 23-02-59](https://cloud.githubusercontent.com/assets/12952169/23834180/16435570-0778-11e7-9711-664a42bc446c.png)
)

This will also address one more issue that even after the data is replaced, still the patient selected is not replaced until we refresh it.

pri2si17-1997:
@aethelwulffe @teryhill 

Some help needed. Do we have to close the iframe (tab) after successful import or at cancel. As this is not working.

pri2si17-1997:
hello @aethelwulffe, this can be closed now.

PR #509, opened 2017-03-14
Fixed invalid import data.

pri2si17-1997:

### DIFF
--- a/custom/import_xml.php
+++ b/custom/import_xml.php
@@ -121,71 +121,77 @@
   if ($olddata['squad'] && ! acl_check('squads', $olddata['squad']))
    die("You are not authorized to access this squad.");
 
-  newPatientData(
-   $olddata['id'],
-   $apatient['title'],
-   $apatient['fname'],
-   $apatient['lname'],
-   $apatient['mname'],
-   $apatient['sex'],
-   $apatient['dob'],
-   $apatient['street'],
-   $apatient['zip'],
-   $apatient['city'],
-   $apatient['state'],
-   $apatient['country'],
-   $apatient['ss'],
-   $apatient['occupation'],
-   $apatient['phone_home'],
-   $apatient['phone_biz'],
-   $apatient['phone_contact'],
-   $apatient['status'],
-   $apatient['contact_relationship'],
-   $apatient['referrer'],
-   $apatient['referrerID'],
-   $apatient['email'],
-   $apatient['language'],
-   $apatient['ethnoracial'],
-   $apatient['interpreter'],
-   $apatient['migrantseasonal'],
-   $apatient['family_size'],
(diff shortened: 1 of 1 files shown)


pri2si17-1997:
Hello @teryhill @aethelwulffe 

This commit addresses #504 

I have tested it on my machine. If any empty data or invalid data is imported then it will popup an alert and nothing will happen. Else data will be saved in database.

Please check it and let me know if anything i'm missing.

custom/import_xml.php
```diff
@@ -121,71 +121,77 @@ function setInsurance($pid, $ainsurance, $asubscriber, $seq) {
...
-   $apatient['interpreter'],
-   $apatient['migrantseasonal'],
-   $apatient['family_size'],
-   $apatient['monthly_income'],
-   $apatient['homeless'],
-   fixDate($apatient['financial_review']),
-   $apatient['pubpid'],
-   $pid,
-   $olddata['providerID'],
-   $apatient['genericname1'],
```
aethelwulffe: genericname, genericval stuff should be taken out.  That is not used.  More importantly, the actual table structure should be queried (select *) for what the fields are.  Matching can be done after that.

custom/import_xml.php
```diff
@@ -121,71 +121,77 @@ function setInsurance($pid, $ainsurance, $asubscriber, $seq) {
...
+   		$apatient['migrantseasonal'],
+   		$apatient['family_size'],
+   		$apatient['monthly_income'],
+   		$apatient['homeless'],
+   		fixDate($apatient['financial_review']),
+   		$apatient['pubpid'],
+   		$pid,
+   		$olddata['providerID'],
+   		$apatient['genericname1'],
+   		$apatient['genericval1'],
```
aethelwulffe: "generic" stuff was a pre- "billing note" field usage, and no longer has a place.  Other fields like social security number have no place in usage as an importer (or anything else in the medical industry).  SSN's are only useful in the medical industry for social security claims, and that is handled in a different way.  I suggest avoiding the use of it in a standard distribution in any case (not relevant to this commit I suppose).

custom/import_xml.php
```diff
@@ -121,71 +121,77 @@ function setInsurance($pid, $ainsurance, $asubscriber, $seq) {
...
+
+  	setInsurance($pid, $ainsurance, $asubscriber, '1');
+  	setInsurance($pid, $ainsurance, $asubscriber, '2');
+  	setInsurance($pid, $ainsurance, $asubscriber, '3');
+  }
+  else
+  {
+  	echo "<html>\n<body>\n<script language='JavaScript'>\n";
+  	//if ($alertmsg) 
+	echo " alert('$alertmsg');\n";
```
aethelwulffe: Looks good.

PR #510, opened 2017-03-14
Fixed JQuery exceptions.

pri2si17-1997:

### DIFF
--- a/interface/usergroup/addrbook_edit.php
+++ b/interface/usergroup/addrbook_edit.php
@@ -286,7 +286,8 @@ td { font-size:10pt; }
  $(document).ready(function() {
   // customize the form via the type options
   typeSelect("<?php echo attr($row['abook_type']); ?>");
-  if(abook_type == 'ord_lab') {
+  //making changes to reflect appropriate value of type selcted.
+  if("<?php echo attr($row['abook_type']); ?>" === 'ord_lab') {
     $('#cpoe_span').css('display','inline');
    }
  });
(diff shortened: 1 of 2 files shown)


pri2si17-1997:
Hello @teryhill @aethelwulffe ,

There was JQuery exception in `Administration->Addr Book` as follows : 
screenshot from 2017-03-15 01-04-20

After that clicking on `Add New` produces following exception:
screenshot from 2017-03-15 01-04-44

Now its working fine on my machine.

Please do look at it and let me know if any other changes required.

teryhill:
I think the issue is that abook_type is undefined so it is throwing an exception. The check you are using does not make sense to me. One thing I do is hard code the variable and see if my test , if statement, works.

pri2si17-1997:
@teryhill . 
Exception shown in first screenshot is due to missing jquery plugin. So I added it along with necessary fancybox js.

In second screenshot `abook_type` was not defined. In the code it was mentioned to display according to type selected. If we look at the line `typeSelect("<?php echo attr($row['abook_type']); ?>");` it gets the value selected. It can be checked in console. So I compared that value in if condition, as `abook_type` is not a variable declared but is one of the column of `users` table. 

This was what my understanding was while making this change. Please let me know if i'm wrong.

pri2si17-1997:
Hello @teryhill 
Now you please check it and let me know if anything else needed.

teryhill:
Looks Very good  @pri2si17-1997

PR #511, opened 2017-03-15
Tabs follow iframes, floating unlocked tabs

azwreith:
Partially addresses #420 
Grouped tabs is a bit complex to program, so until I can figure that out, this seems okayish.

Thanks!

### DIFF
--- a/interface/main/tabs/css/tabs.css
+++ b/interface/main/tabs/css/tabs.css
@@ -107,12 +107,25 @@ body
 .tabSpan
 {
     margin: 0 0 0 0;
-    padding:  0 0 0 0;
-    border: 2px solid black;
+    padding: 5px;
+    border-left: 1px;
+    border-right: 1px;
+    border-top: 1px;
+    border-bottom: 0;
+    border-style: solid;
+    border-color: black;
     border-radius: 6px 6px 0 0;
     cursor: pointer;
     font-weight: bold;
     flex: 0 1 auto;
+}
+
+.tabFloat
+{
+    border: 1px solid black;
+    border-radius: 6px;
+    margin-bottom: 5px;
+    margin-right: 2px;
 }
 
 .tabSpan:hover
--- a/interface/main/tabs/templates/tabs_template.php
+++ b/interface/main/tabs/templates/tabs_template.php
@@ -1,35 +1,50 @@
 <script type="text/html" id="tabs-controls">
-    <div class="tabControls" data-bind="with: tabs">
-        <!-- ko  foreach: tabsList -->
-            <span class="tabSpan bgcolor2">
+    <!-- All the unlocked tabs -->
+    <div class="tabControls">
+    <!-- ko  foreach: tabs.tabsList -->
+        <!--ko if:!locked() -->
+            <span class="tabSpan tabFloat bgcolor2">
                 <span  data-bind="text: title, click: tabClicked, css: {tabHidden: !visible()}"></span>
-                <span class="typcn typcn-refresh" data-bind="click: tabRefresh"></span>
-                <!--ko if:!locked() -->
-                    <span class="typcn typcn-lock-open"  data-bind="click: tabLockToggle"></span>
-                <!-- /ko -->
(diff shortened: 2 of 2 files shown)


teryhill:
Looks good, would be great if the tabs had a minimize in addition to the close. That way you could minimize an unlocked tab.

azwreith:
I'm not quite sure I understand, I assume on minimizing the tab, the page would close but the tab will stay on top right?

aethelwulffe:
Yes, that is what he is saying.  That is so cyber-hoarders can have 20 tabs open and still work.  Honestly, it is actually a good idea.  If you have run a report, and don't have the ability to pop it into another browser tab or window (most ehr functions get goofy when you try that) then being able to preserve it on a hidden frame would be nice.  It would also be nice if duplicate tabs were marked (1), (2) etc....which is yet more complex...or even given a color scheme or something.  That is all much more elaborate, and this pull request is a good deliverable as-is.  I am fine with merging this and forging ahead vs. making it into a long-term project with no merge.  More commits, smaller size!

azwreith:
@aethelwulffe Yuh I agree that would be nice. I think if we are adding the minimizing functionality, we should do it for all the tabs not just the unlocked ones, since they together take up only one window anyway. I'd like to take up that issue! (though I'll only be able to get to on Tuesday)

aethelwulffe:
@teryhill  or anyone else, please mark a review or give a thumbs-up so we can move this in.  This is a good deliverable on the path to being "slick".  That is my story, and I am sticking to it.

azwreith:
Also, @aethelwulffe, @teryhill please review my GSoC proposal draft! I can't seem to find you guys online on librechat T_T

aethelwulffe:
@azwreith 
I will be proud to...and yes, I am a bit out of pocket.  I am actually trying to avoid responding to too many Libre messages, because my answers might not be well formed, or otherwise terse due to current workload insanity.

tmccormi:
Looks nice and simple to me.

azwreith:
@aethelwulffe https://goo.gl/4rI7RW
Your critique will be most appreciated :grin: 

Thanks!

aethelwulffe:
Critique initiated.  I will be watching for those TODO sections to be completed.

azwreith:
They will be! How's the timeline looking? I think I set the schedule to give adequate time to everything, but some of the sections might prove to be unexpected long while implementing.
Is there anything else in the project you'd like to be included in the timeline?

azwreith:
I'm finally done with completing up all of the TODOs! I'm aiming my to complete the project in 3 phases, briefly:
Phase 1: Replace the old calendar by new one, including porting current functionalities
Phase 2: Add new features to the new calendar, completely remove of SMARTY
Phase 3: Revamp the entire UI, documentation, guides, etc

Please take a look!

tmccormi:
Did the link change, I can't get to it? https://goo.gl/4rI7RW

azwreith:
@tmccormi Sorry! Stupid me forgot to send the new link xD
Here: http://goo.gl/t9PkO4 :)

tmccormi:
I like the proposal in general,  I think it's very likely to be too aggressive.  Might consider just focusing on the Calendar replacement.

The Smarty removal and replacement of the UI that use it should probably be a separate proposal, as well as the database revamp related work.

--Tony

azwreith:
@tmccormi Okay! It's just that I want to do both of the things and I can't decide which >_< I am allowed to submit 2 proposals right? I can just split this one into two different proposals then :D

nileshprasad137:
@aethelwulffe @azwreith @tmccormi Even me.  :laughing:

tmccormi:
Split away!

azwreith:
@tmccormi Hey, I just checked on the web, we are allowed to submit multiple proposals but they must be to different organizations. 

In that case I'd like to submit my proposal for the calendar replacement as part of GSoC and then have a smooth transition over to removing SMARTY completely and revamping the UI after GSoC ends :)

PR #515, opened 2017-03-15
Added the ability for certain appointment status to be overbooked (if…

kchapple:
… appt cancelled, for example.) Find-availble will return those slots as available. This was added in such a way that a plugin can override the status array

### DIFF
--- a/interface/main/calendar/find_appt_popup.php
+++ b/interface/main/calendar/find_appt_popup.php
@@ -40,9 +40,20 @@
 
  $input_catid = $_REQUEST['catid'];
 
+ // These are the default "overbooked" statuses, which means that
+ // Even though there is an event, an appointment can be booked
+ // These can be overridden by a plugin by adding overbook_statuses action
+ function default_overbook_statuses()
+ {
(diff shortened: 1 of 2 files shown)


teryhill:
Looks good but should it be on a global switch and shouldn;'t the user be able to select the statuses they want to over book?

tmccormi:
This is related to #513 - code (like the other hooks) should have no effect on the base application without the plugin.

tmccormi:
It's a plugin: if they want to change the defaults they would need to provide (or acquire) the plugin to do that.   We do that so customer specific requests don't affect the baseline.

teryhill:
Add what you want, I ave yet to see any benefit from this line of code it is putting extra code in the code base that is just there. It is your project @tmccormi

tmccormi:
Once you learn how to use it you will see the benefit ... :-) It's our project, not mine ...

kchapple:
@teryhill that's a better idea. thanks.

tmccormi:
shall we close this a wait for a new one, then @kchapple ?

tmccormi:
Deleting this one.

PR #517, opened 2017-03-15
Added the ability for certain appointment status to be overbooked (if…

kchapple:
… appt cancelled, for example.) Find-availble will return those slots as available. This was added to globals.


tmccormi:
Related to #513

tmccormi:
This one looks good to me..   the merge is against rel_101..  can pull if forward to master as well if that's OK.

aethelwulffe:
Just explain the reasons, and I will try to come around.

tmccormi:
This one is ordinary globals not plugin based..

PR #477, opened 2017-03-07
print calendar closes abruptly in chrome

teryhill:
print calendar closes abruptly in chrome but works in FireFox @pri2si17-1997 thanks for finding this

PR #519, opened 2017-03-18
Fixed calendar print for week view and month view in chrome.

pri2si17-1997:


pri2si17-1997:
Hello @teryhill ,
This is in addition to #477 
In that commit, only date view is fixed. That time we missed to check for month and week view. 
Please do have a look at it and let me know if anything i'm missing.

pri2si17-1997:
Yes @aethelwulffe ,

If we have a new calendar, then surely we can get rid of these small things, as they will be in mind while re-creating it. :)

PR #530, opened 2017-03-27
Refined Double Booking logic, ignore canceled appts and ALL Dr views

tmccormi:
This is a patch/enhancement for a customer against rel-101, I need to port it to master but it's testing well in this version.

### DIFF
--- a/interface/patient_tracker/patient_tracker.php
+++ b/interface/patient_tracker/patient_tracker.php
@@ -416,6 +416,7 @@ function openNewTopWindow(newpid,newencounterid) {
 
 <?php
     $prev_appt_date_time = "";
+    $overbookStatuses = array_map('trim', explode( ',', $GLOBALS['appt_overbook_statuses'] ) );
 	foreach ( $appointments as $appointment ) {
 
                 # Collect appt date and set up squashed date for use below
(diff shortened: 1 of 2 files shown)


interface/patient_tracker/patient_tracker.php
```diff
@@ -483,12 +489,16 @@ function openNewTopWindow(newpid,newencounterid) {
...
+            // flag possible double booked, ignoring canceled appt types
+            //  and skipping of ALL providers are viewed
             $apptflagtd = '<td class="detail" align="center">';
-            if ($GLOBALS['ptkr_flag_dblbook']) {
-                if ($appt_date_time === $prev_appt_date_time) {
-                    $apptflagtd = '<td class="detail" align="center" bgcolor="orange">';
+            if ((count($chk_prov) == 1 ) and $appt_canceled != true) {
+                if ($GLOBALS['ptkr_flag_dblbook']) {
+                    if ($appt_date_time === $prev_appt_date_time) {
+                        $apptflagtd = '<td class="detail" align="center" bgcolor="orange">';  //TODO Make color configurable
```
aethelwulffe: This needs css.

library/globals.inc.php
```diff
@@ -493,7 +493,8 @@
     'gbl_currency_symbol' => array(
       xl('Currency Designator'),
        'text',                          // data type
-       '$',                             // default
```
aethelwulffe: ISO 4217.  Dollar (USD), Euro, Yen and BPS are four most commonly traded currencies, but providing currency code based conversions are a bit outside our scope, despite the folks doing business in the borderlands.

aethelwulffe:
Someone please review the code, and review the workflow effects of this so we have two sets of eyes on things and we can see about pulling it.

PR #533, opened 2017-04-01
some code-cleanup

nileshprasad137:
Minor Code cleanup.

This Pull request intends to clean-up some wrong CSS comment format and extra `<tr>` in tables.

### DIFF
--- a/interface/main/tabs/edit_menu.php
+++ b/interface/main/tabs/edit_menu.php
@@ -44,8 +44,8 @@
  * JSON EDITOR is included from:
  *   https://github.com/josdejong/jsoneditor
  *   Copyright (C) 2011-2015 Jos de Jong
- *   usage docs etc in js/jsoneditor/docs
-
+ *   usage docs etc in js/jsoneditor/docs 
+ */
 
  /* Include our required headers */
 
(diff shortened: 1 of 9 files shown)


interface/main/tabs/edit_menu.php
```diff
@@ -82,8 +82,7 @@
     <link href="js/jsoneditor/jsoneditor.css" rel="stylesheet" type="text/css">
     <script src="js/jsoneditor/jsoneditor.js"></script>
```
nileshprasad137: Why are these lines marked with red? Is it because of the comment `/* Include our required headers */` in this file above ? Please check.

aethelwulffe: No idea at all.  Totally confounded.

aethelwulffe: Yes.  the comments have caused the github syntax check to spaz.  I had no idea it did much syntax checking, as I have only seen such in this case...where it is wrong.

nileshprasad137: I've seen this behaviour first time as well. Last commit fixes this.

tmccormi:
I think a better description of what this is is in order...  I don't seen any issues, but  "some code cleanup" is does not describe the commit adequately

nileshprasad137:
Alright, I've edited the description.

PR #534, opened 2017-04-06
Added "Back to patient" button on immunizations 

nileshprasad137:
There was no ` "Back to patient" ` button on `Immunizations ` tab for patient. This PR adds a back button for going back.

Also, I tried to align some code for better readability. I suggest if anyone is working on any issue, he should try to re-align the code of that particular file .


tmccormi:
Note on cleaning up code formatting.   If the changes are complex, it is better to do the changes with out code cleanup so that the new code can easily be seen.   I suggest code clean-up should be a separate issue and pull request entirely, not a mixed in.

PR #539, opened 2017-04-11
Fixed download pdf report for patient.

pri2si17-1997:
Signed-off-by: Priyanshu Sinha <pksinha217@gmail.com>

PR #541, opened 2017-04-11
Tweaking #539 html2pdf upgrade

aethelwulffe:
This removes the composer files and all that, which may not be wise.  I have not verified if there are any issues associated with fonts removed either.  I suspect you may want to just cherry pick out of this, or use it as a reference.
Moved the directory back to just html2pdf
*included updated files
*removed currently unused files
*checked all references in the original html2pdf mod branch to ensure
everything was cool with the new versions
*updated documentation.
NOTE: if anyone feels that additional fonts are required at this time,
yell now.


aethelwulffe:
@pri2si17-1997 You probably want to keep the composer files and the like that I snuffed in this branch.  This is just my input.

pri2si17-1997:
@aethelwulffe I think they are not required.

aethelwulffe:
Well, I know squat about composer.  I am a "drag and drop" integrator myself.   I figured you might have some different opinions though.

tmccormi:
@aethelwulffe  "tweaking" seems like a weak description for something that effected 279 files.   what do you mean by "might want to cherry-pick this"?

aethelwulffe:
We are discussing this in chat...sorry.
Basically he brought in the full package with updated content, added it as a new directory etc...
This is a branch off of his PR that sort of does the same thing with his newer code that was done with the original move of these files.  Namely, adding docs, removing excess files etc...  279 files indicates that he added 210 files that had been deleted previously, and the remaining 68 were moved back over to the original directory name of /html2pdf.
The "cherry pick" reference was not necessarily a GIT term, but a "use this as a reference for what I am talking about".  To that end, the question was if he should include the composer files and any additional fonts that I dropped in this branch.

Issue #445, opened 2017-03-01
Create a 'on demand' receipt that can be printed for a patient at any time.


teryhill:
Create a 'on demand' receipt that can be printed for a patient at any time. This will be a receipt containing the patient information , visit information and payment information. This will have a patient selection and date range.

nileshprasad137:
@teryhill Is this report needed to be added here?
![rep](https://cloud.githubusercontent.com/assets/16336390/23822667/68e9df62-0677-11e7-8e42-a381599cd598.PNG)

rksahu17:
@teryhill Can you assign this task to me?

nileshprasad137:
@rksahu17 Are you working on this?

teryhill:
@nileshprasad137  The Idea for this is a receipt that the office can run at anytime. It would be an on demand receipt. This can be a cleaner format of the Ledger . If a part of the ledger it would need to need be available in both areas of the ledger print, The patient demographics screen and the reports screen.  May be as just a better print option tan what was originally done on the ledger. Just , in my opinion, needs to be a more professional print out. While you are at it see if you could figure out how to print a statement from the Ledger also.

teryhill:
@nileshprasad137  did that explanation help?

nileshprasad137:
@teryhill  Yes! But what would be the contents of such a report? I have some confusion over this.

nileshprasad137:
@teryhill 

> While you are at it see if you could figure out how to print a statement from the Ledger also.

This hasn't been done yet and i intend to do this as and when I get some time.
It would be great if you can recommend some changes in the format , whenever you get some time.

tmccormi:
Pull reverted., reopened issue

PR #543, opened 2017-04-15
Addresses #445, Adds a On-Demand Receipt

nileshprasad137:
This Pull request addresses #445 . 

receipt

Note: This PR may be currently labelled as `"WORK IN PROGRESS"` , as the receipt  added currently behaves very similar to the ledger itself , having a date range to print the payments. 
@aethelwulffe @teryhill Please have look at the format and recommend some changes if required. We need to figure out a way to print a statement from the ledger itself.

### DIFF
--- a/interface/reports/pat_ledger.php
+++ b/interface/reports/pat_ledger.php
@@ -346,7 +346,7 @@ function sel_patient() {
   #title {
     visibility: visible;
     display: inline;
-  }
+  }  
 }
 </style>
 
@@ -356,7 +356,7 @@ function sel_patient() {
  $(document).ready(function() {
   var win = top.printLogSetup ? top : opener.top;
   win.printLogSetup(document.getElementById('printbutton'));
- });
+ });  
 </script>
 
 </head>
@@ -443,6 +443,8 @@ function sel_patient() {
                     <div id="controls">
                     <a href='#' class='css_button' id='printbutton'>
                          <span><?php echo xlt('Print Ledger'); ?></span></a>
+                    <a href='#' class='css_button' id='printreceipt' onclick="printDiv()">
+                         <span><?php echo xlt('Print Receipt'); ?></span></a>
                     <?php if($type_form == '1') { ?>
                     <a href="../patient_file/summary/demographics.php" class="css_button" onclick="top.restoreSession()">
                          <span><?php echo xlt('Back To Patient');?></span></a>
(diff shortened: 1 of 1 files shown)


nileshprasad137:
Thanks @aethelwulffe 

>  I am also curious as to the potential for such statements to be stored, recreated on demand, or even generated via a patient portal. 

I will surely look for these when I get some time.

teryhill:
There is a version of the patient ledger in the patient portal so adapting this code to it should not be an Issue. I would like to see just one version of the patient ledger report program used for both Portal and the main core.

aethelwulffe:
OK guys, what is the next step here?

teryhill:
I got some changes I will send when I get back to the office.

teryhill:
It would be good to have the facility information on the receipt. Get rid of the UAC Appl and UAC tot information. Add provider name. That ought to be good for now.

nileshprasad137:
I would be fully back once my exams get over on 20th May. Till then, you all may make required changes.

aethelwulffe:
Just 3 more days!  WooHoo!

nileshprasad137:
@aethelwulffe I'll get back to this asap. Also, I'd like to work on the format of other reports during the summer , Posting Screen,etc.

nileshprasad137:
> It would be good to have the facility information on the receipt. Get rid of the UAC Appl and UAC tot information. Add provider name. That ought to be good for now.

I think of wrapping this up by making these changes .

aethelwulffe:
Note:  It is real important for the end users to be able to select what facility goes on a particular statement or receipt.  Obviously, the facilities marked as "billing facilities" (in the practice UI you are going to be re-writing soon anyway) should be the first choices, but the clinic may have a legal name, a couple of hundred (especially for assisted living facilities) different facilities, and several of those may have names that need to go on the output file instead of the actual legal entity...or vice versa.

nileshprasad137:
After latest commit:
capture

Currently, user would be able to select facility if he visits `Reports->Financial->Pat Ledger ` but he is unable to select from the patient screen itself.

tmccormi:
@teryhill is this ready to merge?

teryhill:
Looks good to me asfar as the printout. I have not looked at the code.

aethelwulffe:
Uh...Nilesh is doing this weird thing in the code where he sticks to one scripting language for like more than half of a line at a time...I thought in PHP you had to randomly write <?php every few characters... 😈 .  Yeah, the code looks good to me.

tmccormi:
your funny....

aethelwulffe:
Not as funny as you're.

tmccormi:
Check it online ... fixed that typo, but dang github has not email notice dealy and not update notices...

tmccormi:
And as you may have noticed my type+send is faster than my +edit phase.

teryhill:
The only thing I noticed in my real quick testing is that provider is not the provider in the patient file but it is the provider selected in the report  and comes up all in the patient print.

aethelwulffe:
Ouch.
It needs to be the billing/rendering provider from the billing table (as transposed to form_encounter after a fee sheet is saved) or the per-line-item provider from the billing table, and not the "provider_id" from demographics though...generally speaking...and should be able to do "authorized" status checks on those items if someone is smart enough to use those checkboxes in the fee sheet...assuming the feature is turned on.   However...we can't get everything in one commit...

tmccormi:
should I revert?

nileshprasad137:
Yes! I think you should revert it. I'd look at them today. I think ledger
had the same behaviour, so I did it that way. I'm not fully aware of many
billing concepts,that's what creating problems.

tmccormi:
reverted.

tmccormi:
Do a new pull request, please

Issue #468, opened 2017-03-04
Posting Screen Changes.


teryhill:
Need to add a denial remarks dropdown to the posting screen. This should be based on a list. The remark should show on the collections report as a field and should be searchable. This is the current posting screen.

![image](https://cloud.githubusercontent.com/assets/4655366/23579092/b5157988-00b3-11e7-946f-b9031f06c39e.png)

Would also be good to revamp the screen it self.

aethelwulffe:
Terry, do you have any specific suggestions for a re-design of the lay-out?
  Myself I am thinking a text summary of the status so that it is easy to read vs. looking at the populated radio buttons, then offering controls (with reverts) in the context of that status.
Like:
Name
Providers (all)
Expandable field listing the patient's insurance payer(s) and other details 
Posted to insurance X Primary (indicator if the insurance has been updated as well)
Then "Bill to secondary [Y]" and other more in-depth conditional labels.
This will require care with the translations table and the like... though I now have a good way of pulling out all the items that need translation out of the code base and making them into a table (with the file and line statistics in notes where they are used).

teryhill:
No I do not have any specific suggestions. It just needs to be brought into this century 8^)
There needs to be a group discussion on it but I thought the GSOC folks might have some Ideas.

nileshprasad137:
@teryhill Can you tell me which page page is this , how to reach here?

teryhill:
Fees=>posting then enter a patient

aethelwulffe:
@nileshprasad137 
I will provide a testing database that has billing data and the like ASAP.

nileshprasad137:
Great! I'll like to work on it.

aethelwulffe:
The patient must have billing data (make an encounter, go to form /administration/fee sheet, which you can only do properly if you have added the billing codes and all that.
Feel free to mail me at art AT suncoastconnection  D0t c0m for links and the like I will be able to share from one of my servers.  We will get this sort of resource put into a repo later, but I can't currently create repos here...

nileshprasad137:
@aethelwulffe I have mailed you , asking for sample datasets.

teryhill:
@aethelwulffe Did you ever get the sample data to @nileshprasad137

nileshprasad137:
Hey , i think that I'd created the sample dataset myself. Isn't that same
as the one i used to address the issue of on-demand receipt.?

teryhill:
[...]

PR #545, opened 2017-04-21
Billing changes

teryhill:
This allows the auto write off function to be turned off. It also adds
some information ti the posting screen that will be used in tracking
denials. #468


tmccormi:
We really need to create migration/database upgrade stuff now :-)

Issue #531, opened 2017-03-29
Delete Patient


pri2si17-1997:
Steps to reproduce : 

- Create patient or use existing patient.

- Delete that patient from the DELETE button provided. A popup will come displaying the tables from that patient reference has been deleted(ideally this should not happen). Close that popup.

- The page will not get refreshed after closing the above said popup, and it will still show the patient as it is.
 It should get refreshed as soon as the patient is deleted.

Here is the screenshot : 
![screenshot from 2017-03-29 16-27-43](https://cloud.githubusercontent.com/assets/12952169/24451503/b7b3c896-149c-11e7-8405-c1055a8f48f4.png)

Clearly it can be seen that, no patient is present right now (as patient finder shows no records). And export is showing some default values. Ideally that must get refreshed immediately.

pri2si17-1997:
Hello @teryhill @aethelwulffe , please look at this so that if required then I can proceed further.

pri2si17-1997:
Hi @aethelwulffe 

This is what log file shows : 
[...]

PR #546, opened 2017-04-22
Fix for Delete Patient.

pri2si17-1997:
This addresses #531.

Hello @teryhill @aethelwulffe, this PR fixes delete for patient. As the patient is deleted, user will not see from which tables data is deleted, and page will get refreshed as soon as the patient is deleted.


pri2si17-1997:
Hey @teryhill @aethelwulffe  There is some conflict. I'm closing this PR.

Issue #555, opened 2017-05-29
Patient Billing Encounter by Carrier report


teryhill:
I had a request for this report. It is based off the encounter report but it has more options. The billing person wanted it as a snapshot of what they were working on. It uses the table sort js so each column is sortable. Screen shot here.  Report Name could use some work!!!

![new_report](https://cloud.githubusercontent.com/assets/4655366/26561339/70aa0342-448b-11e7-98d1-20cb5c3a607e.JPG)

I also added the table sorting to the encounters report.

tmccormi:
I've always wondered with that "Status" is based on.  What does "Open/Closed" mean.  Empty seems obvious (missing coding).

teryhill:
Here is the logic from the code.

      if ($billed_count && $unbilled_count) $status = xl('Mixed' );
      else if ($billed_count              ) $status = xl('Closed');
      else if ($unbilled_count            ) $status = xl('Open'  );
      else                                  $status = xl('Empty' );
so if there are items that are billed and unbilled on the encounter it is mixed
if it is billed it is closed
if it is unbilled it is open
if none of these apply (no charge) it is empty.

Clear as MUD right?

PR #556, opened 2017-05-29
Encounter by carrier

teryhill:
#555 Not sure why I have Arts commit for Dan in this.


teryhill:
If we decided on the theme to use we could cut out all the rest of them. It has the options to use arrows to indicate direction but I had issues figuring that out and could not get it to show properly , If some one else can figure it out that would be great. I needed it to be done , you know that under the gun stuff.

aethelwulffe:
Yeah, I am pretty familiar with having an artillery piece laying on top 
of me while people are pinging me with a bb-gun.

If you see a default theme you think would be a good look...otherwise 
maybe I will see if there is a way to integrate the themes with the 
application's flavor of them.

teryhill:
@aethelwulffe see if that's more gooder.

tmccormi:
Is this one ready to merge?

teryhill:
Unless you guys want to change the name or something Yes it is ready

PR #561, opened 2017-05-30
JS header include library function

aethelwulffe:
Call js assets within php tags from anywhere after simply dumping your .js library stuff into /assets/js/
  Includes example of use from within the "tabs" directory...whatever those are!


teryhill:
Can we go ahead and just put everything in the header file and then just include it? Including the css stuff?

<link rel="stylesheet" href="../../library/css/jquery.datetimepicker.css">
<script type="text/javascript" src="../../library/js/jquery-1.7.2.min.js"></script>
<script type="text/javascript" src="../../library/js/jquery.datetimepicker.full.min.js"></script>
<?php include_once("{$GLOBALS['srcdir']}/ajax/payment_ajax_jav.inc.php"); ?>
<script type="text/javascript" src="../../library/js/blink/jquery.modern-blink.js"></script>
<script type="text/javascript" src="../../library/js/common.js"></script>

teryhill:
We would have to get everything to the same level.

aethelwulffe:
The headers.inc.php file is for functions that call headers...so if you have a bunch of common ones you could arrange them in packs and just call them with report_includes(), form_includes() etc... or make a big fat list in a bunch of nested functions and call the ones you want with a single call general_mishmash_header($arg1,$arg2,$arg3)

aethelwulffe:
Part of the idea is that when you move or copy a .js file to /assets/js/my_js/index.js, you can replace the original script tag with this function and never have to worry about the path...assuming you have called globals.php.
I wish the call to that was simply $webroot/globals.php.  Then everything would be easy.

teryhill:
I up vote it also

PR #564, opened 2017-05-31
Migrate jscolor.js picker

aethelwulffe:
This library is not actually needed, as <input type="color"
value="#ff0000"> takes care of the issue with HTML5.  As two versions
with a minor release number between them existed, and each was only used
once in the application, I merged them together.
Ref #416  #561


aethelwulffe:
<input type="color"
value="#ff0000">

dpnashsh:
How can we make the jscolor panel stable? I dont want it to enable on focus or click. I want it to be visible all th time.

teryhill:
@dpnashsh you are posting to a closed issue. Could you start new issue for this or go to the forums and post it there. The link to chat is chat.librehealth.io/channel/support the link to the forums is forums.librehealth.io

PR #565, opened 2017-05-31
Fix for silent fail

teryhill:
Fix for silent fail This needs to go in asap


teryhill:
It was Mine.

aethelwulffe:
Blame as in "date it was created and what installs it might affect"

teryhill:
PR545 1 month ago 306e3bc

PR #566, opened 2017-05-31
openflashchart as module

aethelwulffe:
Moved openflashchart to the modules directory...even though it contains
"graphs.php" which is a native piece.  This tool is a mixed-script
module of applications in php as well as self-contained assets.


teryhill:
Is there a newer version of this out there?

teryhill:
Code looks ok to me also . I don't know how to test it 8^)

tmccormi:
Use it in the Vitals to view graphs of weight, Blood pressure etc ...

aethelwulffe:
Please test before merging...

aethelwulffe:
There is a mid-2009 version...only 8 years old or so.  Hard to tell.  I ain't got flash...

aethelwulffe:
After downloading the latest version and pulling out all the excess junk, diffing between the two resulted in one comment being different and a "security" addition that was added that basically exited the script.  The version we have is the latest and last (circa 2009).

tmccormi:
There a project ... replace openflash with something from this decade.  PHP or JS ...

aethelwulffe:
Well, it would be one of the HTML5 canvas thingies with svg data I imagine.  That is a job for another day...at least openflash does not require a plugin.

PR #568, opened 2017-06-01
Made some changes to consolidate the jquery tools

teryhill:
Made some changes to consolidate the jquery tools #416 

When complete thru the code base the added items will be removed from /library/js.


interface/billing/edit_payment.php
```diff
@@ -477,30 +478,17 @@
...
-<script type="text/javascript" src="../../library/dialog.js"></script>
-<script type="text/javascript" src="../../library/js/jquery.1.3.2.js"></script>
-<script type="text/javascript" src="../../library/js/fancybox/jquery.fancybox-1.2.6.js"></script>
-<script language='JavaScript'>
- var mypcc = '1';
-</script>
 <?php include_once("{$GLOBALS['srcdir']}/payment_jav.inc.php"); ?>
 <?php include_once("{$GLOBALS['srcdir']}/ajax/payment_ajax_jav.inc.php"); ?>
-<script type="text/javascript" src="../../library/js/common.js"></script>
+<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/common.js"></script>
```
aethelwulffe: Pretty sure common.js and some others could be moved already,  but if you are sure that this list of includes are everything that is required and intend on migrating other bits in other commits (good idea) then this is fine.

interface/globals.php
```diff
@@ -191,13 +191,13 @@ function undoMagicQuotes($array, $topLevel=true) {
 //module config TODO:  module and global registry for same.
 
 // images directory
-$GLOBALS['images_path'] = "$web_root/assets/images";
+$GLOBALS['images_path'] = "$web_root/assets/images/";
 
 // css directory
-$GLOBALS['css_path'] = "$webserver_root/assets/css";
```
aethelwulffe: do we have $webserver_root and $web_root both still?

interface/orders/procedure_provider_edit.php
```diff
@@ -44,7 +44,7 @@ function invalue($name) {
 <script type="text/javascript" src="<?php echo $webroot ?>/interface/main/tabs/js/include_opener.js"></script>    
 <title><?php echo $ppid ? xlt('Edit') : xlt('Add New') ?> <?php echo xlt('Procedure Provider'); ?></title>
 <link rel="stylesheet" href='<?php echo $css_header ?>' type='text/css'>
-<script type="text/javascript" src="../../library/js/jquery.1.3.2.js"></script>
+<script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>/jquery-min-3-1-1/index.js"></script>
```
aethelwulffe: Assuming 3.1.1 is tested...

interface/reports/encounters_report.php
```diff
@@ -53,8 +54,8 @@ function show_doc_total($lastdocname, $doc_encounters) {
   }
 }
 
-$form_from_date = fixDate($_POST['form_from_date'], date($DateFormat));
-$form_to_date = fixDate($_POST['form_to_date'], date($DateFormat));
+$form_from_date = fixDate($_POST['form_from_date'], date('Y-m-d'));
```
aethelwulffe: is there a problem with the configurable date format here that we need a static Y-m-d?

interface/billing/edit_payment.php
teryhill: Tried to move the ones to header.inc.php that were common to all files. Not sure what opening the unused ones would benefit or hinder.

interface/globals.php
teryhill: Nope we do not use $webserver_root that I can find

interface/orders/procedure_provider_edit.php
teryhill: Yep have it running and testing locally.

interface/reports/encounters_report.php
teryhill: I was getting a goofy looking date when I was in MM/DD/YYYY format so I switched it back till I could spend some time later looking into it.

aethelwulffe:
It is all better now then, as soon as someone merges this commit @tmccormi 
I am with ken on most of this stuff.
Being able to just make viewable objects available to anything in the application would be the way to go with (at very least) the super stuff.
  It would be almost as good as having a real programming language.
...anyway, I would still be interested in figuring out why superglobals act as superglobals on my machine, and not with Terry's set-up.  Probably that I have something turned on in my server variables that is a no-no for production.

PR #575, opened 2017-06-02
Hcfa form check

aethelwulffe:
Check for existing hcfa form options before allowing new one


teryhill:
Should CMS 1500 be changed to Miscellaneous?. It looks and works good.

aethelwulffe:
Recommend you push a change to this that has your preferred warning label.  This was your request after all!

teryhill:
I will if the billing ladies complain. @tmccormi commit this Please!

Issue #571, opened 2017-06-01
Charges Panel


teryhill:
Do we want to keep it? Yes or no. It is not setup in the menus. It is not normally used since the feesheet was written.

tmccormi:
drop it.

aethelwulffe:
We have never used the charges panel, 'cause we only started electronic billing 14 years ago.
Drop it.

PR #578, opened 2017-06-02
Removal of charges panel

teryhill:
I think this gets them all #571


interface/patient_file/encounter/diagnosis.php
```diff
@@ -1,312 +0,0 @@
-<?php
```
aethelwulffe: LibreEHR\interface\super\rules\library\RuleCriteriaDiagnosis.php may be calling this file.

library/globals.inc.php
```diff
@@ -2635,7 +2629,7 @@
         'ms' => xl('Malay'),
         'ml' => xl('Malayalam'),
         'mt' => xl('Maltese'),
-        'mi' => xl('Māori'),
+        'mi' => xl('Maori'),
         'mr' => xl('Marathi (Marāṭhī)'),
```
aethelwulffe: If you are removing non-US keyboard characters, maybe you should do all of them before they get translated here?  I think this stuff should only be included when translations are on (period).

teryhill: These could be causing some of the collation errors

interface/patient_file/encounter/diagnosis.php
teryhill: He has a return with that file name. I am not sure what he is doing, but it is not pointing to the interface/patient_file/encounter/ directory. 

Tell me do y'all want me to put it back?  @tmccormi @aethelwulffe

library/globals.inc.php
aethelwulffe: Well, there are a lot more non-US keyboard characters in all that.  Any suggestions?

aethelwulffe:
Looks like you got to replace the globals.inc.php

teryhill:
Resolved

aethelwulffe:
Mr. Tony,
Please look this over and give it a push if you would sir.

aethelwulffe:
@tmccormi  I realize you are correct, but I think @teryhill intention was to fix a silent-fail issue.  There is some issue there that needs to be addressed, so we need to get that issue documented and hopefully replicated.

tmccormi:
I do not think that this fixes the silent fail.  That would be database.sql not globals.inc.php.

aethelwulffe:
I figured that the silent fail was when this stuff was dumped into the database, and I thought is wasn't included in database.sql...rather processed to do the inserts the first time it is called vs. within the actual schema file.  I forget how this particular mess works sometimes though...as does everyone else.

tmccormi:
Yes, it does not relate to this issue.  It should be a separate issue and tested, not mixed in with a demolition commit.

tmccormi:
Done.  thanks for pandering to my OCD about mixing issues ...

aethelwulffe:
Sure thing, Ling-Ling!

Issue #581, opened 2017-06-04
Dr Ray's reminder and recall code


teryhill:
Dr Ray's reminder and recall code. If you want to test it as a tester let me know.

aethelwulffe:
I am still trying to figure out how to get an actual test set up.
Where is some documentation?

teryhill:
no documentation yet. 

Once you install the software (the commit) then you will notice that messages now has some options that weren't there before. go under File and then setup MedEx. supply an e-mail and password agree to the items listed and this will setup your account on the medexbank servers. 

The changes for the postcards , SMS messages, E-mail, and AVM messages are taken care of on the server. Just be patient we will have them in a little while.

teryhill:
Did I mention that the testing is FREE for right now

teryhill:
also some major changes are being made to the medexbank server side software and so there is an issue with windows that is actively being worked on

teryhill:
Another change to the display on the flow board with the reminder code

![image](https://user-images.githubusercontent.com/4655366/26909795-382025cc-4bd1-11e7-9e4a-8ae4a5978f20.png)


The blue Icon under arrive time is extra text that was sent to the clinic by the patient from the received SMS. Have some Ideas on expansion of that.

aethelwulffe:
OK, that is pretty cool.

aethelwulffe:
How are those changes on the Medex side coming along?  Were is a link so I can learn more about the company, the service, and the API (without doing the setup)?

teryhill:
medexbank.com

aethelwulffe:
Thank you.

teryhill:
SMS Bot is functional. If you click on the message Icon in tracker this is what you will see.

![image](https://user-images.githubusercontent.com/4655366/28143426-704ada14-6733-11e7-8621-83df898c95fc.png)

PR #582, opened 2017-06-04
Reminder and recall

teryhill:
reminder and recall #581


interface/main/messages/print_postcards.php
```diff
@@ -0,0 +1,96 @@
...
+  "p.fname, p.mname, p.lname, p.pubpid, p.DOB, " .
+  "p.street, p.city, p.state, p.postal_code, p.pid " .
+  "FROM patient_data AS p " .
+  "WHERE p.pid = ? LIMIT 1", array($pid));
+
+# sprintf to print data
+$text = sprintf("  %s %s\n  %s\n  %s %s %s\n ", $patdata['fname'], $patdata['lname'], $patdata['street'], $patdata['city'], $patdata['state'], $patdata['postal_code']);
+
+$postcard_message = "It's time to get your EYES checked!
+Please call our office to schedule\nyour eye exam  at (413) 276-4543.
```
aethelwulffe: This hard-coded stuff needs to come from the facility data or a custom template and configuration.

interface/main/messages/save.php
```diff
@@ -0,0 +1,279 @@
...
+        if ($logged_in) {
+            $token      = $logged_in['token'];
+            $practice   = $hb->practice->sync($token);
+            $token      = $logged_in['token'];
+
+            $response   = $hb->practice->sync($token);
+            $campaigns  = $hb->campaign->events($token);
+            $response   = $hb->events->generate($token,$campaigns['events']);
+
+            $response['success'] = "OK BABY!";
```
aethelwulffe: That is the required response?

interface/main/messages/print_postcards.php
teryhill: We need to work on the postcards. They will be configurable from the MedExbank server like the SMS, Email and AVM messages. We were working on these first and then get to the other features.

aethelwulffe:
While this stuff could probably make it to a module, it is currently wrapped up in the general mashup of the communication tools.  If we push it to interface and libraries, we should perhaps do just a bit more work on it first.  The post-cards and other bits should be customizable pretty early on (lest we forget), or perhaps we can just move a piece or two into the already existing directories under /sites as "Examples" that don't have a real clinic's info in them.  I will see if I can help with this a little.

teryhill:
I will be cutting down the assets by removing the items that we are not using.

teryhill:
@aethelwulffe I am working on the post cards.  What do you think will be the best way of getting the information in the body of the card. The facility information is there.

postcard

teryhill:
@aethelwulffe medexbank,com is the service that sends the SMS, AVM and emails.

aethelwulffe:
I have been getting the test SMS and email just fine.  I assume my responses have shown up.

teryhill:
Yes the confirm by SMS and the confirm by e-mail you sent showed up just fine. I have a commit I need to make for some small fixes. Right now working on a way to stop the automatic refresh on tracker so notes can be entered.

aethelwulffe:
I am pretty sure this piece is going to take a lot of security review etc... and I am gonna need to put some real work into the reviews...  Do we have a clinic that is going to be using this as a test/production in the immediate future?

teryhill:
I will bringing up a customer in the next week or so. It is currently running on Dr Ray's system.

aethelwulffe:
Cool.
Speaking (out of context) about Dr. Ray, could you push the eye form to the new forms repo?

teryhill:
That will take some work. I can put it back on the list, If it is wanted?

aethelwulffe:
IfNot upgrade changes look correct.

teryhill:
conflicts resolved. I tested the script on my system which already had the medex options installed.

interface/main/messages/save.php
```diff
@@ -0,0 +1,264 @@
...
+        $data['sender_name']    = $user_data['fname']. " " .$user_data['lname'];
+        $data['sender_email']   = $facility['email'];
+        $data['callerid']       = $facility['phone'];
+        $data['MedEx']          = "1";
+        $data['ipaddress']      = $_SERVER['REMOTE_ADDR'];
+
+        $prefix ='http://';
+        if ($_SERVER["SSL_TLS_SNI"]) $prefix = "https://";
+        $data['website_url']    = $prefix.$_SERVER['HTTP_HOST'].$web_root;
+        $practice_logo = "$OE_SITE_DIR/images/practice_logo.gif";
```
teryhill: What is the valid value?

interface/main/messages/save.php
```diff
@@ -0,0 +1,264 @@
...
+                        <li> ".xlt('Activate your subscription plan')."</li>
+                        <li> ".xlt('Contact support if your server\'s IP address is dynamic')."</li>
+                    </ul>
+                </div>
+                ";
+            echo json_encode($response_prob);
+        }
+                    //then redirect user to preferences with a success message!
+    } else {
+        echo xlt("Sorry you are not privileged enough...");
```
teryhill: Changed

interface/main/messages/js/reminder_appts.js
```diff
@@ -0,0 +1,464 @@
...
+ * @link http:Librehealth.io
+ */
+
+
+var labels=[];
+var postcards=[];
+var show_just;
+
+$(function() {
+  $( "#datepicker1" ).datepicker({ dateFormat: "yy-mm-dd" });
```
teryhill: no it is not the same, also the one in patient portal is the same they both will need to be changes at some point in time

interface/main/messages/save.php
```diff
@@ -109,10 +111,10 @@
 
             $runQuery ="select * from facility order by name";
             $fetch = sqlStatement($runQuery);
-            while ($frow = sqlFetchArray($fetch)) { $facilities[] = $frow; }
+            while ( $frow = sqlFetchArray( $fetch ) ) { $facilities[] = $frow['id']; }
```
aethelwulffe: Ouch.  Guess this doesn't happen much.  Is there a way to do full end-to-end validation via their API to make sure all the pieces work?

library/MedEx/API.php
```diff
@@ -2058,7 +2069,8 @@ public function login() {
         $this->curl->setData(array(
```
aethelwulffe: any particular special cURL configuration stuff needed server-side?
I suppose the install needs to check for:
extension=php_curl.dll in phpinfo

teryhill: Did not have to do anything to my standard windows set for my test box or the user that is getting the code

interface/main/messages/save.php
teryhill: It is working on my test system and has been for a while

library/MedEx/API.php
aethelwulffe: Finally scrolled down past the giant word-pressy things on their site to find out how much it costs.  Something tells me that our outfit isn't going to be able to afford the $62,400 a year subscription fees for text messaging...

teryhill:
not up to you. the cost is 99.00 per month per provider

On Tuesday, July 11, 2017, 3:42:40 PM EDT, Art Eaton <notifications@github.com> wrote:


@aethelwulffe commented on this pull request.

In library/MedEx/API.php:
> @@ -2058,7 +2069,8 @@ public function login() {
         $this->curl->setData(array(

Finally scrolled down past the giant word-pressy things on their site to find out how much it costs. Something tells me that our outfit isn't going to be able to afford the $62,400 a year subscription fees for text messaging...

aethelwulffe:
Well, 45 counselors x12x100.00 makes it basically an impossible dream.

teryhill:
start with the busiest ones or just a couple and see if it improves your customer communication. I am sure we can work some kind of a deal for you 8^)

aethelwulffe:
They all busy, and most share clients with a different program, and there is plenty of turnover that would make things crazy.  60K is two salaries, and $100 a month is more than they pay for internet and two phones, so unless it comes out to being something more like the cost of a phone line, this is waaaay out of reach, even at .  HIPAA fax service is only 20 a month for the whole business.  eRX is 50.

teryhill:
Sorry Art. Most of my practices are 1 to 5 providers and the cost is manageable.

aethelwulffe:
Why close this?  We can't afford it, but I am sure that there are plenty of 90210's that may want it.  It does need to be an add-on, but this is an operable connector.

tmccormi:
Agreed.  Make it an add-on module.  It a good option for many practices

teryhill:
Sorry the close was an oops

teryhill:
I will add some controls to turn it off but the code is very heavily embedded in Patient Tracker beyond me to make it a module I do not understand hoe to accomplish it.

tmccormi:
Good reason for a hack-a-thon ...

teryhill:
I will pull the code until that happens it that is what y'all want. I have it working for the version my clients are on and since it appears to not be a fit with LibreHealth EHR. it is just taking up space.

teryhill:
Just an FYI for you guys

	https://www.remindercall.com/pricing-2/  
	Example: 20pts/day x 20d/month = 400 events.  
With a 60 second phone call, and one SMS with a reply, you are at 1600 units which is $127.50/month.

aethelwulffe:
Heh, Terry...I guess we are just waaaaay too darn poor.  Do they take 
bottlecaps?

Honestly, I will look at stuff and see just how much they would actually 
be sending.  On a volume rate, we are OK.  On a per-provider 
rate...hahahahahahahahaha...:-D

aethelwulffe:
Any volunteers to take this over and create a "module" out of it?

PR #585, opened 2017-06-12
Fancybox 1.3.4 migrate

aethelwulffe:
Interface/forms/patient_encounter/common.php
interface/patient_file/encounter/forms.php
interface/reports/patient_list_creation.php
~Are the affected files.  At least two other versions of fancybox
remain.  One all have been migrated, we can decide if we are to split
the js asset packages up by css vs. code, or preserve the asset package
as-is, then get rid of the older or less useful versions.


aethelwulffe:
There be two upvotes here...

PR #586, opened 2017-06-12
added alternative deidentification script

growlingflea:


contrib/util/deidentification/deidentification.php
```diff
@@ -0,0 +1,414 @@
+<?php
+
+/*   This script is for de-identifying an OpenEMR database for the purpose of creating a live demo or for
```
aethelwulffe: OpenEMR is spelled "LibreHealthEHR"

contrib/util/deidentification/deidentification.php
```diff
@@ -0,0 +1,414 @@
...
+    removes data from the encounter, removes personal information from the patient_data, notes, and form_encounter
+    tables. This should be the first function called.
+
+    INPUT: connection, last name array, firstname arrays, debug
+    OUTPUT: Number of patients DEID.
+*/
+function deIdPatientData($con, $lnames, $male, $female, $DEBUG=false){
+
+    removeColumn($con, "patient_data", "ss", "0000-00-00");
+    removeColumn($con, "form_encounter", "reason", "reason goes here, bucko");
```
aethelwulffe: Spelled "Buck-O"

contrib/util/deidentification/deidentification.php
```diff
@@ -0,0 +1,414 @@
...
+    removeColumn($con, "form_encounter", "facility", "Service Facility");
+    removeColumn($con, "notes", "note", "notes goes here, Captian");
```
aethelwulffe: Spelled "Captain".

contrib/util/deidentification/deidentification.php
```diff
@@ -0,0 +1,414 @@
...
+
+    $i = 0;
+
+    $removeLname = ("Select lname, pid, id, ss, street, sex from patient_data ");
```
aethelwulffe: ss, zip+4, city, state should be removed and replaced.  ssn should simply be removed with an "if exists".  A structure query would help a lot here, or use "SELECT *" then see what you get.

contrib/util/deidentification/deidentification.php
```diff
@@ -0,0 +1,414 @@
...
+
+        //Give the user a new first name
+        $first_name_male = $male[rand(0,32)];
+        $first_name_female = $female[rand(0,74)];
+
+        //Change the street address patient_Data.street
+        $street = rand(1,9999)." ".rand(0,200)." Avenue ";
+
+        //remove the drivers license
+        $drivers_license = rand(2,999).rand(0,999).rand(0,99);
```
aethelwulffe: DL is not standard, and should also be simply dropped.  It doesn't contribute to a de-identified database usage any more than ssn.

contrib/util/deidentification/deidentification.php
```diff
@@ -0,0 +1,414 @@
...
+
+        $string .= " phone_biz = '$phone_home', ";
+
+        $string .= " contact_relationship = '', ";
+
+        $string .= " email = '', ";
+
+        $string .= " street = '$street' ";
+
+        $string .= "where pid = "."'".$result['pid']."' ; " ;
```
aethelwulffe: You should probably just preserve PID, or re-order the ID fields, then set PID =ID.

contrib/util/deidentification/deidentification.php
```diff
@@ -0,0 +1,414 @@
...
+
+
+        mysqli_query($con, $string) or die("Failed Patient Replacement");
+        $string = '';
+        deIdInsuranceDataTable($con, $result['pid']);
```
aethelwulffe: better to create fake insco data out of whole cloth.  Depends on the usage of the database after this point.

contrib/util/deidentification/deidentification.php
```diff
@@ -0,0 +1,414 @@
...
+
+    }
+
+    echo "successfuly altered user table \n ";
+}
+
+//Clears most forms.  User must verify that this function handles all text fields that might hold personal identifying information
+function deIdForms($con){
+
+    removeColumn($con, "form_physical_exam", "comments", "no comment, talk to my lawyer");
```
aethelwulffe: Nicer might be to look for the replaced names from patient_Data, which could leave you with realistic seeming data that is (mostly) de-identified.  Unfortunately, few people use the soap form or any of the others we have.  If those forms don't exist in the database, this breaks.

teryhill:
@aethelwulffe didn't the de-identification directory get removed. I was under the impression that we did not need this since there was going to be an alternate means of generating de-identified data.

teryhill:
@tmccormi and @aethelwulffe does this mean that ya'll want to add back all the deidentification stuff that was removed?

aethelwulffe:
Uh, my opinion was that these scripts were...not complete...and had a 
bunch of stuff from the other project.  I also though it should not be 
in the base code, unless it was in there as a separate module.

Issue #584, opened 2017-06-08
Role based menus. 


teryhill:
I have code from Dr Ray for role base menus that he uses in his practice If it would be a good fit for LibreHealth EHR with the UI changes let me Know and I will send it. @aethelwulffe @tmccormi @nileshprasad137

teryhill:
No one interested in role based menus!!!

aethelwulffe:
I thought I commented on this!  Yes, hell yes I want to see it.

PR #587, opened 2017-06-12
Role Based Menus

teryhill:
Role Based Menus #584


aethelwulffe:
Hey terry,
is this testable as-is with LH?

teryhill:
running it on my system now. It has a few quirks , like you need to be in front desk to add a patient. but it works. There needs to be a menu editor for the database, since all this is in the tables.

aethelwulffe:
What would you say I could do to make this a final thing?  Should we make a DB-based menu editor form, or a JSON schema editor?  I would be willing to help out with either one.

teryhill:
It runs against the tables and ignores the JSON file. A editor for the Tables is all that is needed. I know @yehster did a menu builder for the menus. Don't know if it would work with this or not, but the menu editor that @tmccormi did will need changing to work with the tables.

aethelwulffe:
Which one do you think serves us better?  In a DB or in a JSON file?  
Either one is better than left_nav, and it has always confused me why it 
was never a thing handled by layouts... Probably best that it wasn't, 
but it always seemed short-sighted to me.

teryhill:
It could be handled thru layouts with the DB and that may be a good place for it. Just needs a good editor that the customer can understand.

aethelwulffe:
So...do it in some way that people will understand?  What's the value in 
that? :)

-OK, sounds good to me.

teryhill:
Make it easy enough to use that unhum will want it.

yehster:
The menu items ultimately end up as JSON, because that's what knockout in the browser needs in order to build up the menu.  Serving different JSON depending on role either hard coded or from the database should be relatively straightforward.

Designing an editor to modify the menus for each of the different roles would be the trickiest part.  One could simply hand modify the baseline JSON as desired.  (Someone could make sample menu options...)  There is code which uses/constructs the menu tree as DB rows which can be tweaked manually as well. 

Translating the convoluted logic in left_nav, handling all sorts of conditionals (concurrent layout, sports team, IPPF, acl, etc...) across so many items is what made creating a data driven representation hard.  My approach was to scan the menu's HTML and build up the JSON from each of the HTML elements, in combination with inspecting the PHP code for the additional logic.

teryhill:
So since we don't have a left_nav any more does that mean the menu builder won't benefit us.

aethelwulffe:
Kevin, looking at the JSON in an editor/validator, it seems like there is little complexity difference in making an editor that is served off of a table vs. saving a json file.    Turning up short trying to find a complete hierarchical JSON editor that is fairly closely attuned to the sort of project we are looking at, I am honestly thinking about building a knockout based tool to do this from the ground up.
  Do you guys see any value in allowing both table stored as well as JSON string files?
Also, when doing either one, would an option to allow totally custom menus per-user be desirable?  Instead of restricting the menu based on role and an ACL check, you create the user menu config file (or set of table rows) when the user is created initially base on their role, then those are the menu items that they can show, re-arrange or hide.  Each user then gets their own menu, or can shift back to the user-role menu default.
  I am just looking at where to demark between files in /sites vs DB storage.

yehster:
The menu analysis code probably not, but the code which takes JSON menu and builds the database rows and vice versa might.  

https://github.com/LibreHealthIO/LibreEHR/tree/master/contrib/util/menu

Last that I remember working with the code, if the rows are loaded in the DB, it uses the DB, if not, it uses the JSON in the PHP file.  Tony made some changes though so I'm not sure of the current state.

teryhill:
If the tables are there it does use them

teryhill:
@aethelwulffe if the tables are used the JSON is not. I say just store the info in the tables.

yehster:
Since a DB representation needs to get converted into JSON before it can be used, if you are using the DB, you essentially have both.  One could store JSON strings for individual users, which if they existed would override the DB records. 

If someone wanted me to manage the menus, I personally wouldn't bother with building an editor.  I'd tweak rows in the current database model from the dataview/spreadsheet of MySQL Workbench to build the different versions and export the results.

Realistically, how frequently is the menu going to be tweaked?

aethelwulffe:
I think that the ultimate goal is for the menu to be responsive.

Scenarios:

   I enable a report, or I create a new report view with (as yet 
non-existent) report builder.

I turn on a module or feature.

I get rid of a module.

I change someone's access.
I redefine an access level.

In any of the above, I think we want the menu to be be easy to handle 
for the dev or the user.  When you use an editor that operates within 
schema constraints, it is pretty much impossible for the user to mess up 
the results too!

I honestly think I can build a drop and drag builder that would not be 
heavy.  I did something like this a long time ago, where I used the 
client's file manager to create a directory structure (similar 
hierarchy), then parsed the directory (and file contents that had the 
URLs in them) to build a configuration file...then got rid of the 
directory.  That was kind of whack-o (worked well) but it used an 
amazingly small amount of new code, just the user's file manager.  I 
won't do anything that weird if I take this on though...I think...
   I am pretty sure that using a JSON editor is no less of a hassle than 
any "utility" quality tool.  Either it is slick and extensible, or we 
shouldn't bother.

aethelwulffe:
I am scrounging for time at the moment...I am still working this though.

teryhill:
@aethelwulffe have you had any thoughts on the menu editor for this?

aethelwulffe:
Well, there are several things in the whole role-based dealio I am trying to sync with the FacilityACL stuff (that is really problematic to begin with) as well as the bukakke code in the TAGS stuff.  Needless to say, the going is slow...but that is not to say nothing is done yet! 😸 
image
Each menu item is a "node".  The forms directory, the reports directory, and a few other places are also set up to parse the directory for content, and make that available to the menu editor in the drop-downs.  You can also configure those so you can add modules etc... or anything else you need to automatically scan to bring up lists of files in those directories to select for the target URL.  This means that (at the moment) it is creating relative paths and not paths that use stuff like $GLOBALS[] helpers...but I am working on that too.

sql/database.sql
```diff
@@ -7926,3 +7926,579 @@ CREATE TABLE `form_clinical_instructions` (
...
+('Administrators', 'Forms|/interface/forms_admin/forms_admin.php', '', '', 'Other:', 100, NULL),
+('Administrators', 'Calendar|/interface/main/calendar/index.php?module=PostCalendar&type=admin&func=modifyconfig', '', '', 'Other:', 200, NULL),
+('Administrators', 'Logs|/interface/logview/logview.php', '', '', 'Other:', 300, NULL),
+('Administrators', 'Database|/phpmyadmin/index.php', '', '', 'Other:', 400, NULL),
+('Administrators', 'Certificates|/interface/usergroup/ssl_certificates_admin.php', '', '', 'Other:', 500, NULL),
+('Administrators', 'Native Data Loads|/interface/super/load_codes.php', '', '', 'Other:', 600, NULL),
+('Administrators', 'External Data Loads|/interface/code_systems/dataloads_ajax.php', '', '', 'Other:', 700, NULL),
+('Administrators', 'Merge Patients|/interface/patient_file/merge_patients.php', '', '', 'Other:', 800, NULL),
+('Administrators', 'Patients|/interface/main/finder/dynamic_finder.php', '', '', 'Patient/Client:patimg', 0, NULL),
+('Administrators', 'New/Search|/interface/new/new.php', '', '', 'Patient/Client:patimg', 100, NULL),
```
aethelwulffe: Geez.  Is this still called /new/ ? *facepalm*

aethelwulffe:
OK, just noticed that if I am not in a specific role, I have squat options for doing certain things, like creating a patient.  I suppose that is ok on some level, though an admin override option would be pretty slick, you know, like opening up a terminal window that allows you to run >LEHR -u root -p -createpatient;
...just kidding.
  I though you said that only frontoffice could make a new patient?  It is in the admin role as well.

sql/database.sql
teryhill: That is from the doc's system. What did y'all call it?

teryhill:
Yes that what I was saying earlier there needs to be a more even spread of the options, just not sure what is needed where.

aethelwulffe:
Yeah, I think we should just merge it, then get wider input.  So long as it can be turned on and off, or a superuser has a full menu, there is no harm.

teryhill:
@tmccormi  @aethelwulffe  I did not include the tables upgrade in this commit. So right now the role based items are not working it is just as it always was.

PR #589, opened 2017-06-15
Added an ORDER BY name to the provider search

teryhill:
Added an ORDER BY name to the reffering provider and provider search.


teryhill:
Have this at a customer site and it is being used with no issues. They have 28 providers and 49 referring providers.

aethelwulffe:
Tested and simple.  I am pretty sure that the two of us are the most familiar with actually doing billing, so let's just move this on.

PR #590, opened 2017-06-16
Added back the old upgrade scripts

teryhill:
Added back the old upgrade scripts until the new ones can be done by some one familiar with frame work #91


teryhill:
I did test this as it is in this pr and it updated the added fields. Tony is out of town si You will have to make the decision on whether to merge or not.

aethelwulffe:
Well, we have a number of things (changes) to document for migration pretty much immediately (I sure do, and the edu project probably will), so we need *something* to do that with.  I am merging up.

PR #592, opened 2017-06-20
Removed calendar reference that was causing the error

teryhill:
Removed calendar reference that was causing the program save error


teryhill:
Nothing else. It will save the forms and the date picker works. Merge it

aethelwulffe:
Merging as this is priority bug fix, and our team is a bit short right now...

PR #593, opened 2017-06-20
Bug fix in SOAP and ROS.  en…

yehster:
Move closing HTML tag out of {/php} segement so properly interpreted as HTML


teryhill:
@aethelwulffe I can work with this now since it is a PR. Approving and merging

PR #594, opened 2017-06-21
Add missing includes for calendar to Immunization Page

yehster:
Javascript was failing because of missing dynarch_calendar libraries.


teryhill:
Not sure why you are adding these back? Where are they needed?

teryhill:
@yehster #592 fixed the issue.

yehster:
Didn't realize you were working on the bug too.

PR #600, opened 2017-06-28
UI changes in Patient Finder and Global Settings

nileshprasad137:
Decided to go with Bootstrap 3.3.4 as its already there in repo and there are a very few minor changes in 3.3.7.  Added the link to Bootstrap assets in `header.inc.php`.

finder
global

### DIFF
--- a/interface/super/edit_globals.php
+++ b/interface/super/edit_globals.php
@@ -300,22 +300,20 @@ input     { font-size:10pt; }
 <?php } else { ?>
   <p><b><?php echo xlt('Edit Global Settings'); ?></b>
 <?php } ?>
-
+<!--tabNav-->
 <ul class="tabNav">
 <?php
 $i = 0;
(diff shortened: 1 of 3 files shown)


interface/super/edit_globals.php
```diff
@@ -300,30 +300,28 @@ function checkBackgroundServices(){
...
+    echo " <li" . ($i ? "" : " class='current'") ."><a href='/play/javascript-tabbed-navigation/'>" . xlt($grpname) . "</a></li>\n";
     ++$i;
   }
 }
 ?>
-</ul>
+</ul> 
 
-<div class="tabContainer">
+<div class="tabContainer well">
```
aethelwulffe: Is "tabContainer well" a valid class name?

aethelwulffe: Uh, oh, nevermind.

PR #610, opened 2017-07-03
fixed adress label to work on envelopes.  User can format envelope se…

growlingflea:
…ttings as well

### DIFF
--- a/interface/patient_file/addr_label.php
+++ b/interface/patient_file/addr_label.php
@@ -5,7 +5,7 @@
 * Program for displaying Address Labels 
 * via the popups on the left nav screen
 * 
-* Copyright (C) 2014-2017 Terry Hill <teryhill@librehealth.io> 
+* Copyright (C) 2014 Terry Hill <terry@lillysystems.com> 
 * 
 * LICENSE: This program is free software; you can redistribute it and/or 
 * modify it under the terms of the GNU General Public License 
@@ -18,13 +18,19 @@
 * You should have received a copy of the GNU General Public License 
 * along with this program. If not, see <http://opensource.org/licenses/gpl-license.php>;. 
 * 
-* @package LibreEHR 
-* @author Terry Hill <teryhill@librehealth.io>
-* @link http://www.libreehr.org 
+* @package OpenEMR 
+* @author Terry Hill <terry@lillysystems.com>
+* @link http://www.open-emr.org 
 *
 * this is from the barcode-coder and FPDF website I used the examples and code snippets listed on the sites
 * to create this program
+ *
+ *
+ * Ability to format printed addresses on envelopes improved / modified by Daniel Pflieger
+ * daniel@mi-squared or growlingflea@gmail.com
 */
+
+
 
 $fake_register_globals=false;
 $sanitize_all_escapes=true;
@@ -34,11 +40,12 @@ require_once("$srcdir/classes/PDF_Label.php");
 require_once("$srcdir/formatting.inc.php");
 require_once("$srcdir/classes/php-barcode.php");
 
+
 //Get the data to place on labels
 //
 
 $patdata = sqlQuery("SELECT " .
-  "p.fname, p.mname, p.lname, p.pid, p.DOB, " .
+  "p.fname, p.mname, p.lname, p.pubpid, p.DOB, " .
   "p.street, p.city, p.state, p.postal_code, p.pid " .
   "FROM patient_data AS p " .
   "WHERE p.pid = ? LIMIT 1", array($pid));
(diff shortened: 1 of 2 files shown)


growlingflea:
The address label for envelopes was broken just like in OpenEMR5.  I made some changes.  The only difference is that these settings are in Miscellanous (the label settings were put here instead of PDF for reasons unbeknown to me).  I kept it that way and added the envelope formatting settings in Miscelanous.  Feel free to take a look.

interface/patient_file/addr_label.php
```diff
@@ -5,7 +5,7 @@
 * Program for displaying Address Labels 
 * via the popups on the left nav screen
 * 
-* Copyright (C) 2014-2017 Terry Hill <teryhill@librehealth.io> 
+* Copyright (C) 2014 Terry Hill <terry@lillysystems.com>
```
teryhill: Needs to go back

interface/patient_file/addr_label.php
```diff
@@ -18,14 +18,20 @@
 * You should have received a copy of the GNU General Public License 
 * along with this program. If not, see <http://opensource.org/licenses/gpl-license.php>;. 
 * 
-* @package LibreEHR 
-* @author Terry Hill <teryhill@librehealth.io>
-* @link http://www.libreehr.org 
+* @package OpenEMR 
+* @author Terry Hill <terry@lillysystems.com>
+* @link http://www.open-emr.org 
 *
```
teryhill: These 3 lines need to go back the way they were

interface/patient_file/addr_label.php
```diff
@@ -34,11 +40,12 @@
 require_once("$srcdir/formatting.inc.php");
 require_once("$srcdir/classes/php-barcode.php");
 
+
 //Get the data to place on labels
 //
 
 $patdata = sqlQuery("SELECT " .
-  "p.fname, p.mname, p.lname, p.pid, p.DOB, " .
+  "p.fname, p.mname, p.lname, p.pubpid, p.DOB, " .
```
teryhill: we don't use pubpid

interface/patient_file/addr_label.php
```diff
@@ -48,37 +55,51 @@
...
-$pdf = new PDF_Label('5160'); // used this to get the basic info to the class
-$pdf = new eFPDF('P', 'mm',array(102,252)); // set the orentation, unit of measure and size of the page
-$pdf->AddPage();
-$pdf->SetFont('Arial','',50);
 
+//Format envelope settings.  Changes made by Daniel Pflieger
+//Pull the settings from globals.  If the Global settings have not been determined in Globals,
+//the default is to print on a 4 1/8 by 9 1/2 envelope.
+//todo: put the option in settings to set envelope size.
+
```
teryhill: Did you do away with just the labels? What if they are going on something besides an envelope? I know there was a change to the appointment report to allow labels.

teryhill:
Not sure what was broken. The address label appears to work on my local test system and the demo system including Nhanes.

growlingflea:
So the address label printed in one size only.  Really BIG.  There was weird code like the following:

image

The object $pdf gets created then overwritten the very next line.  The formatting and the envelope size were hard coded so the end user wasn't able to print on more than one size envelope.

interface/patient_file/addr_label.php
growlingflea: No.  I modified just the address labels.

aethelwulffe:
I noted the "gallon size" effect that had before.  I think the office here just used print scaling.

interface/patient_file/addr_label.php
aethelwulffe: pubpid bad.

teryhill:
Prints on a dynmo label printer.

aethelwulffe:
@teryhill   Brother, I ain't got no printer...and print to file isn't where the money is.  Have you got one of these 'dynamo' label printers, or can you tell me how I should be testing this?  It is just addresses/envelopes, right?  Not things like pee bottles or pill bottles?

teryhill:
Yep I develop with one. Don't buy it if you can't load , don't sell it if you can't test it.

aethelwulffe:
Well, I can't test it fully.  Sorry Dan.  This one is up to you guys.

tmccormi:
I don't have a printer like that either ... let me know when/if I can merge this.

teryhill:
I think @growlingflea is just using a regular printer for his changes.

teryhill:
@tmccormi there are some review Items that need to be fixed

aethelwulffe:
...and I don't even know what the big deal with that printer even is...or why it should be preferred to any other printer.  Never seen anything in the docs that talk about any special printer.

teryhill:
A dymo printer is a printer that prints labels, just labels. It is really a neat printer for what it does. We used them for specimen and tubes in the lab. This is the one I have. http://www.dymo.com/en-US/labelwriter-450-turbo-label-printer

aethelwulffe:
Ah, it is a speci label printer then.  That is important.  This isn't just for address labels.  Sounds like we need options that handle that printer as well as your typical MFP.

teryhill:
It does not matter to me if this gets changed. The customers that use the existing program can keep the program they have and use it.

aethelwulffe:
@teryhill  does this still work with the bottle label printer?  If it does not, then this is just silly.  We will need both options at very least.

aethelwulffe:
Also, do we want a printing tab for globals...because it seems to me we should have one.

teryhill:
@growlingflea needs to weigh in on this.  I have not tested but it looks like it is a replacement for the labels,Which is fine with me as I have already stated.

aethelwulffe:
We just need confirmation from Dan, then we will merge.

growlingflea:
Sorry for being out of the loop.  The issue that my customer was running into was that they were unable to print on size 10 envelopes.  Just to confirm, the previous version of the label function only printed on one size label, unless you have a special printer (which I think is what Terry was saying).

The Globals allows you to specify the size of the paper, and the starting point on the x and y axis

image

tmccormi:
x and y axis are in what kind of measurement?

growlingflea:
mm.  I went with the terminology that was in the code.

growlingflea:
Correction:  I accidentally posted an image from OpenEMR5 admin->globals.  In LibreEHR its under Miscellaneous because that is where the "Print Patient Address Label" was

tmccormi:
Is this ready guys? @teryhill @aethelwulffe

aethelwulffe:
Go for it.

Issue #554, opened 2017-05-29
Calendar replacement discussion


azwreith:
## Current Functionality:

**View calendar:** 
Date/Week/Month: already handled by FullCalendar
According to provider, facility

**Print calendar.**

**Add/Edit Appointments:**
Book for Patient, Schedule for provider
Edit/Rescheduling
Find Available time slots
Find Patient
Recurring appointments

**Find appointments:**
Using keywords, categories, date ranges, providers , facilities.

**Patient Tracker/Flow board:**
Status update and by whom (in demographics)

**Some of the same stuff in patient portal.**

## New features: 
**Wait list**
**Reordering of columns**

## Issues to keep in mind:
**Globals**
**Auto refresh**
**Responsiveness**
**Form validation**
**Work with all date formats**

## Schema stuff:
`libreehr_modules` and `libreehr_modules_vars`: Delete and make a new table for calendar globals.
`categories`, `events`: Reuse/Rename.
`patient_tracker`, `patient_tracker_event`: Keep as is.
`limits` and `topics`: Remove as it's not currently being used.

@aethelwulffe @teryhill Please look this over and see if there's anything missing or incorrect. :sweat_smile:

teryhill:
@azwreith Looks like you have most all of the bases covered. I cannot think of anything that is obviously   missing, I will reserve the right to say "oops did not see that one" at some later date

yehster:
Future Appointments and Appointment History for individual patients on the demographics screen is a significant feature that should probably be explicitly mentioned as existing functionality.

teryhill:
Yes we are looking at an appoint log/audit feature that will show those items plus give more information as to what happens to the appointments.

tmccormi:
- Schedule templates for providers, ie reserved and preferred times for appointment types.
- Appointment reports, there are several and the the fact that appoints can be specifically linked to the 'encounter/visit' so you can report on that.
- Given time we should consider group appointments, ie: multiple provider and multiple patients (behavioral and integrative medicine)
- Updates to ehr-api to support the fhir interface we have for the calendar.

tmccormi:
Aso, consideration on how to install the new appt module and migrate old appointment data.

aethelwulffe:
Schedule templates as a background css, then overlay the other stuff was an idea at one point.  You don't want a "template" actually interfering in any totalitarian way.
  Once we have a new data model, we can throw plenty of "real" data into a migrate script to see how messed up it is! :)

[...]

PR #611, opened 2017-07-04
Calendar module replacement

azwreith:
Intermediate PR for calendar changes for reviews and other stuff. I'll keep it updated with changes as they come~

EDIT: Addresses #554


aethelwulffe:
Nice big chunk here dude!  Looking forward to breaking the code! 😸

azwreith:
A few thoughts:
I made the day view as default in place of the agenda view. However, the slots being really tiny hides away the text for appointments which take up 1 slot (though the text can be seen via quickly appearing tooltips that I added). If I lengthen the slots, the person using the calendar would have to scroll a lot in order to use it so that's a dilemma. 
screenshot from 2017-07-05 22-14-20

So in case of a few number of doctors the clinic should just switch to agenda view as it displays everything nicely for that situation.

The `disable_calendar` global is used in demographics and not actually used to disable the calendar right?

Also, I remembered I have to move this entire thing to modules! Easy since it's almost entirely independent of everything else but it might complicate the reviewing process if I move the entire thing around :sob: 

Thanks!

teryhill:
@azwreith would it help the display if you did the appointment times vertical for the hour slot. 
right now you have 
8am 8:15 8:30 8:45 9am etc what would you gain it you left every thing the same but made you slots go

8am                         9am                          10am                       11am                       12 am  etc
8:15                          9:15
8:30                          9:30

teryhill:
Sorry but the "Disable Calendar" does just that, Personally I don't see a reason for disabling it. But @aethelwulffe and @tmccormi should make that call.

azwreith:
@teryhill I'm not sure I understood what you mean. Do you mean we have this kind of arrangement

> 8am 9am 10am 11am 12 am etc
> 8:15 9:15
> 8:30 9:30

replicate below each other for every resource (provider)? I don't think something as flexible as that is supported by the library :/

aethelwulffe:
I encourage you to use 24 hour time format as your default in any case.

Ujj, you would make me happy enough to do a bear dance if you made this a module pack.

I was not aware that GLOBALS['disable_calendar'] was used in any way in demographics.  I thought that just disabled the calendar view in the old left_nav menu (long gone) and perhaps made it not show up in the menu system.  That was it's whole function...hiding the feature.  I also do not see a real need for that.  We only really need default tab menu items, and the role-based menu configuration stuff (when complete) should help there.

tmccormi:
I have had one person that used the disable calendar in production and it was me.  No longer using that version of the EHR anyway.  I don't think it needs to be a global option.  Particularly if it's a separate module anyway.

teryhill:
I say we add the option to the removal list. I put one in to disable flowboard that should be removed also

aethelwulffe:
Removing access by menu options is one thing...a "disable" feature is just silly unless there is a logic option that just should not happen (like data inserts) from some other tool...but I think even if those were left in place (if they exist) it would harm nothing and allow someone to start using it in the future rather easily.
  I just can't see where having a scheduling tool is ever a real "no-no".  Let's drop the exception logic.

azwreith:
Oh so you mean I should rename /modules/calendar as /modules/full_calendar?

I don't really understand what ehrapi and ehrfhirapi are for. Can you elaborate please? :sweat_smile:

aethelwulffe:
Wellllll....that repo is sort of a use case of the FHIR api for the old 
calendar system.  Only Ken, the author, and maybe Tony really understand 
it's ways and means, as it is not particularly easy to follow or 
intuitively organized (to me).  The real gist of what I am saying is 
"Look at https://www.hl7.org/fhir/ and see if there are api message 
values and hierarchies that are usable for our own database schema or as 
variable names".  FHIR is considered by a great number of entities as a 
communication standard that actually works.  This means there is will 
and backing to use it as a data transport medium.  Making our internal 
data structures highly compatible with that, or even using it as the 
model for our own data and object structures is a forward looking 
approach that may have several immediate and long-term advantages.

aethelwulffe:
I am saying Re: Modules that it is probably fine having it /calendar/full_calendar.  If you are going to name the directory something, I would have picked the actual name of the project, which would make it /FullCalendar.  full_calendar seems like an alternative to partial_calendar.  FullCalendarJS would probably be instantly recognized by a developer as a project name without further ado.  Your choice entirely.  You got it done, you got it in /modules, and I am happy.

tmccormi:
The two API repos are intended to allow a flexible method to support incoming requests in FHIR API format and to parse the into specific Libreheatlh EHR calls
ehrfhirapi - is the incoming server for FHIR requests
ehrapi - is the API mapping service that FHIR uses to talk to the EHR and respond to the incoming request

There are only a limited number of FHIR requests currently implemented, but they are ALL there to support patient appointments from a mobile application that was developed in Ireland.  Our side is generic and could be used for any UI that wanted to do patient registration and portal appts, for example.

azwreith:
@aethelwulffe The project is actually in `calendar`. `calendar/full_calendar` contains the files of the library. But yep you're right I should name them FullCalendar and FullCalendarScheduler, and probably put them in a `libraries` folder.

Okay, I'll try to understand the APIs and get back to you guys.

Question. Does the "Search" option in the old calendar not work? I tried it in both my machine and ehr.librehealth.io but it always returns no results :confused:

azwreith:
For the latest commit I used js-cookie. I should move it to the main libraries folder of ehr so it can be used anywhere right?

teryhill:
Yes move js-cookie to the assets directory.

I have tried to use the search function and have never figured out hoe to get it to work. @aethelwulffe or @tmccormi got it to work but I am not sure which one did or how they got it to work.

Yes fix the colorpicker2.js, popupwindow.js used in interface/usergroup/facilities_add.php and facilities_admin.php  and calc.php used in appointments.inc.php. 

The calendar replacement is the most important right now If the FHIR stuff is going to slow you down plan it for after GSoC.

aethelwulffe:
We want anything we can for the calendar feature to live under 
libreehr/modules, and only what is totally necessary to link to it under 
interface or libraries.  We really want libraries to contain assets that 
are in common use by many components, or are core functions for things 
that cannot be just "turned off" and have a functioning interface.

Don't let the FHIR api distract you too much, as you have already headed 
a long way down a particular development path.  The main issue there is 
*it will be talking to the calendar eventually* so consult it if you 
need to do any *additional* design work.

The appointment search feature was flakey for some things, yes. There 
were things it would work for (like particular items) but I cannot 
recall what those were.  No telling now, and do we even care?  I want 
the Ujjwal calendar.  I am telling everyone right now that I am not 
doing any upgrades, new installations, or conversions to LHEHR until the 
new calendar is in place, and SMARTY has died an ignoble death.

aethelwulffe:
On js-cookie:
That is an EXTERNALLY supported and MANAGED dependency, so we want it in the LibreEHR/assets/js directory.

azwreith:
Okay, I create my own search thingy from scratch then. 

Where should I move the colorpicker and the rest of the files? Obviously I won't move them to modules/calendar because they aren't being used there. The colorpicker and popupwindow can go in assets but what about Calc.php?

Ujjwal calendar coming up :smile:

aethelwulffe:
Wasn't Calc.php part of the old PostNuke calendar?
  Are you still using it with your calendar?  
Does anything else use it?  If so, then re-write it a bit to remove the PN references (except for attribution) and pop it into a proper place for something we have to maintain, like /libraries/.

azwreith:
@aethelwulffe Calc.php is used in appointments.inc.php. Okay I'll do just that!

@aethelwulffe @teryhill Please test the search stuff I just added and let me know if any modifications or additions are needed ^^"

modules/calendar/full_calendar/fullcalendar.css
```diff
@@ -0,0 +1,1413 @@
+/*!
```
teryhill: May not since this is a Module but a means to get the color changes to it will need to be done @nileshprasad137

modules/calendar/search.php
```diff
@@ -0,0 +1,173 @@
...
+    return preg_match("/$keyword/i", $fmlName) || preg_match("/$keyword/i", $flName) || $keyword == $arr['pc_pid'];
+  }
+  else return true;
+}
+?>
+<html>
+<head>
+  <link href="<?php echo $GLOBALS['standard_js_path'] ?>bootstrap-3-3-4/dist/css/bootstrap.min.css" rel="stylesheet" />
+  <link href="<?php echo $GLOBALS['standard_js_path'] ?>jquery-datetimepicker-2-5-4/build/jquery.datetimepicker.min.css" rel="stylesheet" />
+  <link href="css/search.css" rel="stylesheet" />
```
teryhill: Look at nileshprasad137  work on the header file which can be used to pull in these items.

modules/calendar/search.php
```diff
@@ -0,0 +1,173 @@
...
+      // filter the appointments using the keyword
+      $filteredAppointments = array_filter($appointments, "keywordFilter");
+    ?>
+    <table class="table table-hover">
+      <tr>
+        <th>Date</th>
+        <th>Time</th>
+        <th>Provider</th>
+        <th>Patient</th>
+        <th>Category</th>
```
teryhill: use this on the above information to ensure translation.
`<th><?php echo xlt('Category')?></th>`
Do it for date,time,provider and patient also

modules/calendar/search.php
```diff
@@ -0,0 +1,173 @@
...
+        echo "<td>" . oeFormatShortDate($appointment['pc_eventDate']) . "</td>";
+        echo "<td>" . oeFormatTime($appointment['pc_startTime']) . " - " . oeFormatTime($appointment['pc_endTime']) . "</td>";
+        echo "<td>" . $appointment['ulname'] . ", " . $appointment['ufname'] . "</td>";
+        echo "<td onclick='goPid($pid)' title='Go to Patient'><a href='javascript:;'>" . $appointment['lname'] . ", " . $appointment['fname'] . "</a></td>";
+        echo "<td>" . $appointment['pc_title'] . "</td>";
+        echo "</tr>";
+      }
+      ?>
+    </table>
+    <!-- end if -->
```
teryhill: It would help us old guys later if you would name the loops at the also start as well as the end. Trust me once the code gets large that little information will be helpful something like filter appointments loop start and then the end of the loop

modules/calendar/search.php
nileshprasad137: Yep. Just include `headers.inc.php` from library and look at `call_required_libraries` function.

modules/calendar/full_calendar/fullcalendar.css
nileshprasad137: I'll look at this later. I haven't yet started with color picker but it is on my priority list.

aethelwulffe:
@azwreith Please remove the commented out code, unless you think it should stay, just tell us.  This needs to get merged...finally.

azwreith:
@aethelwulffe Hi! Sorry for not being online, I was travelling back to India for college ^^"
The commented out code is for a global (Providers see entire calendar), the only one which is pending implementation. I'll look at it and all the other stuff today c:

modules/calendar/search.php
azwreith: Okay, thanks!

aethelwulffe:
Waiting on the go-ahead for final review @azwreith .  Really anxious to see this!

azwreith:
@aethelwulffe Yes, ***ALMOST*** done! Just a couple things: 

1. I don't get how docs_see_entire_calendar global works. I tried it with authorized users and without in the old calendar and there's no difference :confused: 
2. I left add_edit_event.php and its sister files in the old calendar. Some places, including my own calendar uses it. Should I copy it to modules/calendar or library?

aethelwulffe:
@azwreith the docs bit is where you have to be an ACL Physician.
I would put everything you can with the module (copy) to consolidate it as a modular feature.  Other things should call it if they want to work with appointments.

aethelwulffe:
@azwreith the important thing is not that you replicate every paradigm from the old...but satisfy the use-plan according to your design.  The "docs see whole calendar" was intended to check the acl for the Physician access level.  If ==0, then the default calendar view was for it to show only the user's schedule.  Otherwise ==1, then show all providers on the schedule.  This may or may not need be a distinction at the global level anymore.  If user preferences make it simple, or if you could *click* on a user name and have the calendar shift to only display that provider's appointments, that would satisfy a lot of need.

azwreith:
@aethelwulffe Okay, go it! I'll make the changes today~

Sorry for being so unresponsive the past couple days, I have campus job fair going in college and it's really hectic right now and probably will be till the end of weekend. I'll try to devote as much time as I can though!

tmccormi:
@azwreith Are we dropping (or have you already dropped) all the postnuke calendar code from this pull?

teryhill:
@tmccormi not yet they are currently co-existing

teryhill:
@azwreith  I like it . It works great . one small change add the day of the week to the screen like this  Saturday, August 5, 2017.

azwreith:
How'd it looking now? :smirk:

aethelwulffe:
Better yet, 2017, August 5.

aethelwulffe:
It would be nice to have a datetime-format standard global option for both short and verbose dates that works everywhere.
I would choose  YYYYmmdd 24-hour, and Year, Month Day in verbose as well.  Others might like any other sort of confusing mess, but standardizing formats to the international and federal standard of year-month-day as is required in most output formats.  To choose anything else other than ISO-8601, we should inform the user that they are using a "wrong" format and ask if them if they "really really really want to do this stupid thing?"
https://ushik.ahrq.gov/ViewItemDetails?system=mdr&itemKey=81200000

tmccormi:
Guys - display of the date needs to be controlled to support INTERNATIONAL use.   EU uses a different expanded format that the US.   So that has to be configurable not fixed in the code

teryhill:
If they are entered thru the code they should be ok. These are because of the nhanes conversion. The date itself is run thru the right stuff for international use.

aethelwulffe:
International is very often YYYYmmdd, and this is still the *default* config we are talking about...not a hard-code.

tmccormi:
I'm talking about the Human readable display, not the storage format.  That differs by site and culture and even occupation (military -vs- others)  Ireland is particularly confused  (as a culture) about what to call time that happens post meridian ...

aethelwulffe:
ISO dude.  That is the standard you code by...for professional usage.  Everything else is configurable option, and generally not a good idea.  It isn't all that big of a mystery.

tmccormi:
No objections to ISO standards as a base line, just saying the user could care less about that and want what they want in order to use the system comfortably and with out errors based on format assumptions,
12/01/1961 or 01/12/1961 both of these are my birthday ... depending on who reads them

aethelwulffe:
Currently, testing seems all straight from my end.
What is the status on merge again @azwreith ?  I can see there are issues having the old calendar with the new, but it would be nice to merge before moving on here.

tmccormi:
Adding or changes items using the add_appt dialog leaves artifacts on screen when used with Chrome from Linux.

aftter-adding-in-office

tmccormi:
Also, is the Admin->Calendar going to be replaced something to setup Visit types in a way that is not so icky?   Does Full Calendar have that somewhere or should we move it to a list_options table entry or something?

tmccormi:
In "Day View" it would be excellent if the view automatically started at the current time instead of having to scroll to the right to see the afternoon everytime.

Agenda View should be the default in globals in my opinion, however.

aethelwulffe:
That is actually a pretty good idea...Sounds like a *round two* goal. :)

aethelwulffe:
@tmccormi  @azwreith @teryhill  as per conversation with Terry, it seems reasonable to me, in the interest of testing and lowering development complexity, to go ahead and get this strong beta merged with master.  I am creating a release branch.  If voted down, we can trash that, but I would like to get this particular PR to the "small commit" stage, as it should be there right now.  We have way too many lines of code hanging in the .git files...

tmccormi:
Can we drop the old calendar code first @aethelwulffe ?   Creating a Release Candidate branch is not a problem.

aethelwulffe:
That would be up to Ujjwal to do I suppose, and if you really are not comfy with merging up with the old stuff still there.

azwreith:
Hi hi!
Sure I'll delete the old files if that's what you guys really want ^^"
Should we keep Agenda view as default then?

aethelwulffe:
Yeah, let's rip and go.  Getting away from Smarty was a stated goal.  No need to have cold feet at this point.  Nilesh is doing his cleanup now, and if you dump yours, then we are rocking the new world.

aethelwulffe:
@azwreith How are things progressing for final cleanup deletions?  Terry (and I) are only waiting on that (as per Tony's request) to merge the module, and are sitting here biting our nails in hopes of being able to do so. 😬 
Of course, that would very nicely wrap up GSOC....

teryhill:
Quick testing looks great. Created an appointment from the calendar and from the patient's demographics.

azwreith:
It's done! A couple things though:
pntempates/default/images included some small images that are used elsewhere (demographics.php and rtl.css)
The Administration->Other->Calendar (for modifying categories and all) doesn't exist right now. I was planing to replace it later anyway 😮

teryhill:
We will need the ability to modify categories as quickly as possible.

aethelwulffe:
Guys, if we are operable, and we can get this PR done and use another _smaller_ one to do the admin UI, please?
@azwreith can we merge this branch and then do the UI?

azwreith:
By all means, go ahead!

On Aug 19, 2017 1:19 AM, "Art Eaton" <notifications@github.com> wrote:

Guys, if we are operable, and we can get this PR done and use another
*smaller* one to do the admin UI, please?
@azwreith <https://github.com/azwreith> can we merge this branch and then
do the UI?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/LibreHealthIO/LibreEHR/pull/611#issuecomment-323445601>,
or mute the thread

aethelwulffe:
Merging the new calendar...and on to the next PR!

PR #613, opened 2017-07-06
added the search that MDSupport did.

teryhill:
added the search that MDSupport with OpenEMR did. also did a little cleanup to some
small things. The search searches the titles and descriotion. Makes it
easier to find tags you are looking for.

### DIFF
--- a/interface/super/edit_globals.php
+++ b/interface/super/edit_globals.php
@@ -18,6 +18,10 @@
  * You should have received a copy of the GNU General Public License
  * along with this program. If not, see <http://opensource.org/licenses/gpl-license.php>;.
  *
+ * LICENSE: This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0
+ * See the Mozilla Public License for more details.
+ * If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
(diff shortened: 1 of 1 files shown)


interface/super/edit_globals.php
```diff
@@ -50,7 +54,7 @@ function checkCreateCDB(){
       $GLOBALS[$globalsrow['gl_name']] = $globalsrow['gl_value'];
     }
     $directory_created = false;
-  if( !empty($GLOBALS['document_storage_method']) ) {
+  if($GLOBALS['document_storage_method'] != 0){
```
aethelwulffe: keeping couchDB you think?

teryhill: Just have not gotten around to addressing that, but I think it is on Pri's list.

aethelwulffe: Well, it's all good by me.  I wuz just axin'.  This is ready to go as far as I can see. @tmccormi

aethelwulffe: Someone, -ANYONE else can (please) review this, then it can be merged.  We don't like globals stuff in PR's hanging out to dry very long.

aethelwulffe: Any takers?  Please someone do a review?  It's a single-file review, not hard...

PR #614, opened 2017-07-10
Remove Smarty from Prescription Controllers&Templates

nileshprasad137:
Removed Smarty Engine from prescription and also fixed typo in `headers.inc.php` . Had to include jquery there as Bootstrap JS required that.

### DIFF
--- a/library/headers.inc.php
+++ b/library/headers.inc.php
@@ -19,7 +19,6 @@
  */
 ?>
 <link rel="stylesheet" href="<?php echo $css_header;?>" type="text/css">
-<link rel="stylesheet" href="<?php echo $GLOBALS['standard_js_path'] ?>bootstrap-3-3-4/dist/css/bootstrap.min.css" type="text/css">
 <?php
 
 function include_js_library($path)
@@ -33,9 +32,38 @@ function include_js_library($path)
 <link rel="stylesheet" type="text/css" href="<?php echo $GLOBALS['css_path'].$path;?>" media="screen" />
 <?php
 }
-
 ?>
+
+<?php 
+/*  
+    This function can be used to call various frequently used libraries.
+    Parameters for this function are boolean. Use true or false for including the required libraries.   
+*/
+function call_required_libraries($bootstrap,$fancybox,$knockout,$datepicker){?>
+    <!-- All these libraries require jQuery to be loaded initially for best performance -->
+    <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>jquery-min-3-1-1/index.js"></script>
+    <?php 
+    if($bootstrap===true){   ?>
+        <link rel="stylesheet" href="<?php echo $GLOBALS['standard_js_path']; ?>bootstrap-3-3-4/dist/css/bootstrap.min.css" type="text/css">
+        <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>bootstrap-3-3-4/dist/js/bootstrap.min.js"></script>
+    <?php
+    }
+    if($fancybox===true){   ?>
+        <link rel="stylesheet" href="<?php echo $GLOBALS['css_path']; ?>fancybox/jquery.fancybox-1.2.6.css" media="screen" />
+        <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>fancybox/jquery.fancybox-1.2.6.js"></script>
+    <?php
+    }
+    if($knockout===true){   ?>
+        <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>knockout/knockout-3.4.0.js"></script>
+    <?php
+    }
+    if($datepicker===true){ ?>
+        <link rel="stylesheet" href="<?php echo $GLOBALS['css_path']; ?>jquery-datetimepicker/jquery.datetimepicker.css" media="screen" />
+        <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>jquery-datetimepicker/jquery.datetimepicker.full.min.js"></script>
+    <?php
+    }
+}
+?>
+
 <script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/textformat.js"></script>
 <script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/dialog.js"></script>
-<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/jquery-1.9.1.min.js"></script>
-<script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path'];?>/bootstrap-3-3-4/dist/js/bootstrap.min.js"></script>
--- a/library/headers.inc.php
+++ b/library/headers.inc.php
@@ -37,4 +37,5 @@ function include_js_library($path)
 ?>
 <script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/textformat.js"></script>
 <script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/dialog.js"></script>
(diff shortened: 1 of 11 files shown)


nileshprasad137:
There are so many jquery versions out there. Included 1.9.1 as its used at most places.

library/headers.inc.php
```diff
@@ -37,4 +37,5 @@ function include_js_library($path)
 ?>
 <script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/textformat.js"></script>
```
aethelwulffe: This header file was not intended to include all the js files every time it was called.  The intention was to use it as a js script calling function.  Basically with all these scripts included with this file, it has added all of them to everywhere this function include is used!
  If we want a "include every js library" file (which we might in a great number of places), that should probably be something else, or written as a function within this file, but we should not be automatically including all of them in everything.

nileshprasad137: Alright, so I think I should remove both BootstrapJS and jquery out of headers.inc.php, for now. I'll include them in the files I'll be working on using functions defined here as and when upgrading UI.

teryhill: Hey guys would it be best to come up with a list of always used tools and in the future look at a way of including them if they are needed. Not sure how to do it yet , haven't thought it thru but if there is a call for bootstrap them make bootstrap available, this is computer programming there has to be a way. It is just like "There would have never been a fin on a fishes back if there was not a water for him to swim in"

teryhill: That way we would not have to go thru all the code to make the changes for tools that change we just change it in one place.

aethelwulffe: Agreed.  Could we not just make additional functions in the headers file that allow you to load a list?

`function include_allthe_bootstrap_stuff_library($arg0,$arg1,$arg2,$arg3,$arg4,$arg5,$arg6,$arg7,$arg8,$arg9) `

teryhill: @nileshprasad137  we also need to get rid of the old version of things when we can instead of 1.9.1 try using 3.1.1

nileshprasad137: Yes, its good. We can do this. But alternatively , one more thing can be done .Whenever I need to use Bootstrap over a page , I'll need 3 files, Bootstrap.min.css, Boostrap.min.js, jquery. So , one function to include boostrap can be written here to include these files. Then , later when working over specific pages in UI, I'll remove the jquery version used there as it would be already in the function defined in this file. What about this?

teryhill: Test with the encounter code when you do this and see if the function of moving from one encounter to another still works, this seams to be a place where things break so it is a good place to test. I think loading bootstrap was causing some issues , not sure though. What ever the means we just need a good simple way of doing this and loading only those things that are needed.

nileshprasad137: Or we can go with above function suggested by @aethelwulffe and then have proper documentation for the arguments , similar to the one we have for datatypes. I'm fine with both.

nileshprasad137: oh , didn't see your last comment @teryhill . OK, I will test that asap and then reply.

teryhill: @nileshprasad137  you are the lead dog on this. Come up with a good way of doing it. A word of wisdom (I think) "Don't paint yourself in to a corner" Think thru it and don't settle for just getting it done.

nileshprasad137: @teryhill I tested it by making a function
```
<?php
function include_boostrap()
{?>
    <link rel="stylesheet" href="<?php echo $GLOBALS['standard_js_path'] ?>bootstrap-3-3-4/dist/css/bootstrap.min.css" type="text/css">
    <script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/jquery-1.9.1.min.js"></script>
    <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path'];?>/bootstrap-3-3-4/dist/js/bootstrap.min.js"></script>

<?php
}?>
```
Included this function in encounter form , tried changing some UI to Boostrap UI. Works fine there.

aethelwulffe: @nileshprasad137 
maybe a nice switch/case instead of IF statements, or anything clear and compact.  I think one function that lets you have one little bit of documentation for the function arguments, then we can use a (large) single-line function to call most any set of headers we need.

nileshprasad137: @aethelwulffe Alright. I'll go with your suggestion then. In this PR itself?

aethelwulffe: Nilesh, you are the formally trained programmer here.  I am just an amateur enthusiast.  Do it that way if you like it, but like Terry said, you are in charge of this.  We are only here to whine if it breaks! 😈

aethelwulffe: Oh, and I think it is called "bootstrap", not booze-trap (function include_boostrap) 😜 

I think I like the "call_any_libraries you want(1,0,0,0,2,1,1,1,0); function idea myself.

aethelwulffe: I ain't got any boot straps on my boots anymore.  I accidentally pulled them off while trying to start a business.

nileshprasad137: We can have parameters for jQuery, Bootstrap, KnockoutJS, datetimepicker, and fancybox. Anything more? We have to basically do this function for files inside `library/js`, right? Except for Bootstrap which is under assets..

aethelwulffe: Actually, we want to move all non-native JS packages (including the version of bootstrap used in the TAGS system which is in /interface and any others we find floating around) _into_ the /assets directory as a migration...so when adding the function to other code, point all of it at the /assets directory.  library/js should ONLY be for NATIVE js stuff that is purely a LibreHealth maintenance item.  This is an ongoing cleanup effort.

In answer to your actual question, yes, awesome, sounds great, make a huge function unless you see a clear grouping that says "these will never be called with those others" sort of case.

Document in the header file how the file should be used and maintained, and any other useful things you come up with.  We want enough in-line documentation to make it clear what will be allowed, and what will not.  The file is for functions that organize and simplify the calling of includes, and for making full migrations or inclusions of new versions as simple as adding the new package and editing the headers.inc.php

nileshprasad137:
```
<?php 
function call_required_libraries($boostrap,$fancybox,$knockout,$datepicker){?>
    <!-- All these libraries require jQuery to be loaded initially for best performance -->
    <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>jquery-min-3-1-1/index.js"></script>
    <?php 
    if($boostrap==1){   ?>
        <link rel="stylesheet" href="<?php echo $GLOBALS['standard_js_path']; ?>bootstrap-3-3-4/dist/css/bootstrap.min.css" type="text/css">
        <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>bootstrap-3-3-4/dist/js/bootstrap.min.js"></script>
    <?php
    }
    if($fancybox==1){   ?>
        <link rel="stylesheet" href="<?php echo $GLOBALS['css_path']; ?>fancybox/jquery.fancybox-1.2.6.css" media="screen" />
        <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>fancybox/jquery.fancybox-1.2.6.js"></script>
    <?php
    }
    if($knockout==1){   ?>
        <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>knockout/knockout-3.4.0.js"></script>
    <?php
    }
    if($datepicker==1){ ?>
        <link rel="stylesheet" href="<?php echo $GLOBALS['css_path']; ?>jquery-datetimepicker/jquery.datetimepicker.css" media="screen" />
        <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>jquery-datetimepicker/jquery.datetimepicker.full.min.js"></script>
    <?php
    }
}
?>
```

I tested the above function with `Edit Globals` screen and `Patient Finder` screen. Works fine.
Should I add any more parameter for any other library?

aethelwulffe:
You might want to add one for bootstrap instead of boostrap.
image

nileshprasad137:
Any more changes reqd. ?

PR #615, opened 2017-07-10
Added file check to forms files. I think it can be taken a step furth…

teryhill:
…er at some point.

Added file check to forms files. I think it can be taken a step further
at some point.  This is basically the same code that is used in other
forms done by ZH. I just modified it to work as a function.

### DIFF
--- a/library/formsoptions.inc.php
+++ b/library/formsoptions.inc.php
@@ -1,0 +1,37 @@
+<?php
+/*
+ * Functions to be used in the forms.
+ *
+ * @copyright Copyright (C) 2017 Terry Hill <teryhill@librehealth.io>
+ *
+ *
+ * LICENSE: This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0.
+ * See the Mozilla Public License for more details.
+ * If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
+ *
+ * @package LibreHealth EHR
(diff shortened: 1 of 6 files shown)


library/formsoptions.inc.php
```diff
@@ -0,0 +1,37 @@
+<?php
+/*
+ * Functions to be used in the forms.
```
aethelwulffe: "clinical/visit/encounter forms?"

aethelwulffe: And include a sentence of how and why.

teryhill: I get the clinical/visit/encounter forms change but not sure about what you want here or do you mean this for the function itself?

aethelwulffe: Noooo.  Nothing that deep.  Just suggesting a bigger explain'n of what the file does.

teryhill: Did that explain it enough?

aethelwulffe: I suppose so.  I was just suggestion of "stuff for encounter forms, just common functions, don't put any BS in here or I will bomb.gzip your butt."
It is fine, and I guess I am just wasting time.  This stuff is ready to be merged...Oh Third set of Eyes! 👁‍🗨 👁

aethelwulffe:
Real good, real clear, and this should get merged pronto IMHO

tmccormi:
ok...

PR #620, opened 2017-07-12
Add Limit to Ins Data sql for sync to New Crop

tmccormi:
NewCrop hates more than 3 insurances and the SQL was sending ALL of them, even if they had been replaced

### DIFF
--- a/interface/eRx_xml.php
+++ b/interface/eRx_xml.php
@@ -850,6 +850,7 @@ function PatientFreeformAllergy($doc,$r,$pid)
     return $allergyId;
 }
 
+//Get insco with Self, LIMIT to first 3 to avoid erX overflow
 function PatientFreeformHealthplans($doc, $r, $pid) {
     $resource = sqlStatement('SELECT
             `ins`.`name`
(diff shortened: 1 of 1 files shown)


teryhill:
There is an inactive flag in the file.

tmccormi:
I'll add that, but that does not exist on the OpenEMR 4.x customers...  so I didn't notice it here.  Thanks

interface/eRx_xml.php
```diff
@@ -784,7 +785,7 @@ function PatientMedication($doc,$r,$pid,$med_limit)
             $b->appendChild( $sig );
             $dispenseNumber = $doc->createElement( "dispenseNumber" );
             $dispenseNumber->appendChild(
-                $doc->createTextNode( $prec['quantity'] )
+                $doc->createTextNode( $prec['quantity'] )  // THIS CAN't BE RIGHT tdm $prec is outside of scope
```
aethelwulffe: That is from the function/query above this one.  Totally out of scope, and was probably meant to be $presc vs. prec.  Very confusing for a while there.  Probably by the same person that used JASON vs. JSON.  :)  Frankly, I am lost as to what this function does, unless it is nested...  $active above starts a query string with AND.  Makes me think it is a nested function, which may mean that $prec is in scope in a sort of slimy, stupid dangerous way.

tmccormi:
Glad to see you agree.  will merge this and address that issue later, as it does not seem to be breaking anything.  eRX needs a revamp, in anycase, but unfunded at this time.

PR #632, opened 2017-07-20
Dr Ray's tan css and compact and full css.

teryhill:
Dr Ray's tan css and compact and full css.


teryhill:
This is what globals looks like on Dr Rays 

image

PR #637, opened 2017-07-22
fixed header and removed pubpid In address label change

teryhill:
fixed header and removed pubpid as requested in the review

### DIFF
--- a/interface/patient_file/addr_label.php
+++ b/interface/patient_file/addr_label.php
@@ -5,7 +5,7 @@
 * Program for displaying Address Labels 
 * via the popups on the left nav screen
 * 
-* Copyright (C) 2014 Terry Hill <terry@lillysystems.com> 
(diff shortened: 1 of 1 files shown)


interface/patient_file/addr_label.php
```diff
@@ -5,7 +5,7 @@
 * Program for displaying Address Labels 
 * via the popups on the left nav screen
 * 
-* Copyright (C) 2014 Terry Hill <terry@lillysystems.com> 
+* Copyright (C) 2014-2017 Terry Hill <teryhill@librehealth.io>
```
aethelwulffe: First publish date is fine by (c) law.  No actual need for expanded, unless you just want to so that dates of re-publish are clear.

aethelwulffe:
I am bypassing the two-factor review for very very minor fixes that are obvious, meaning a single line thing.  This is just in light of expediency, as reviewer time is hit-or-miss, and currently we lack more review input to keep things moving.

PR #639, opened 2017-07-22
Needed for the UB04 claim submission

teryhill:
Needed for the UB04 claim submission. Missed this file in the initial commit.

### DIFF
--- a/library/Claim.class.php
+++ b/library/Claim.class.php
@@ -81,6 +81,7 @@ class Claim {
   var $supervisor_numbers; // row from insurance_numbers table for current payer
   var $patient_data;      // row from patient_data table
   var $billing_options;   // row from form_misc_billing_options table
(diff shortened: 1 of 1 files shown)


library/Claim.class.php
```diff
@@ -265,6 +266,17 @@ function Claim($pid, $encounter_id) {
       "ORDER BY forms.date";
     $this->billing_options = sqlQuery($sql);
 
+    if ($GLOBALS['claim_type'] =='1' || $GLOBALS['claim_type'] =='2') {
```
aethelwulffe: Is there a global that toggles between either professional or institutional billing and prevents you from billing both types?

teryhill:
You can bill 1500,1450 or both. My user uses both.

aethelwulffe:
Question answered.  Change is minor, and all against original author commits.  merging.

PR #644, opened 2017-07-26
RHoyt requested the dates on medications and medical problems

teryhill:
RHoyt requested the dates on medications and medical problems


interface/patient_file/summary/stats.php
```diff
@@ -223,8 +223,12 @@ function load_location( location ) {
                 echo "  <td colspan='$numcols' style='color:red;font-weight:bold;'>&nbsp;&nbsp;" . htmlspecialchars( $row['title'] . $reaction, ENT_NOQUOTES) . "</td>\n";
         }
         else {
+          if ($key == "medical_problem" || $key == "medication") {
```
aethelwulffe: What does it look like with date trailing the name?

teryhill: If you have a long name it staggers the date . I chose the date before name because of this and Immunizations already have it that way.

aethelwulffe: Gotcha.
Getting this in so as not to delay ->EDU demo.

PR #665, opened 2017-08-13
Remove smarty templates

nileshprasad137:
Deleted all smarty templates !


aethelwulffe:
Terry, if you are the second reviewer to pass something, especially minor stuff, ...go ahead and merge.

PR #666, opened 2017-08-19
General cleanup, minor bug fixes and UI improvements in Administration

nileshprasad137:
Just minor changes here and there. Smaller Input Boxes and layout issues.

### DIFF
--- a/interface/forms/fee_sheet/new.php
+++ b/interface/forms/fee_sheet/new.php
@@ -1070,7 +1070,7 @@ echo " </tr>\n";
    <?php echo xlt('for'); ?>&nbsp;
   </td>
   <td>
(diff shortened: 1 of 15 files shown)


aethelwulffe:
I ain't seeing it any different.  Mongo approve...without testing.


At the same time, this PR just adds to my profound belief that we need 
to totally re-do the codes admin stuff...and remove the old fee sheet 
lists, and put all the stuff scattered in lists named "code types" 
Service Categories" and all that stuff somewhere coherent so that 
someone with mere near-genius intellect can figure it all out. That 
means removing the "Fee Sheet" list in favor of just the code categories 
until the whole thing can combine features of both (which is limited to 
being able to group codes/modifiers into packs).

Issue #672, opened 2017-09-21
setup errors for pnTemplates


aethelwulffe:
https://github.com/LibreHealthIO/LibreEHR/blob/6ddcc6b995cb4284431fdb6500c2b1c68948703f/setup.php#L73

setup.php must be updated to reflect removal of old PNcalendar.

aethelwulffe:
Line 73 and 74 at least.

tmccormi:
Are there any directories that need to be writable for this new version, considering that smarty is gone?

tmccormi:
Srry, I missed the fact that there was already a pull request.  All merged now.

PR #673, opened 2017-09-22
setup broken by missing old calendar

aethelwulffe:
discovered by bclee
quote: " Just to clarify WHY this change is being made:
requiredDirectory1 and requiredDirectory2 were setup to make sure
permissions were set correctly for old calendar directories. These
directories no longer exist. The setup script complains and stops
because the directories '' and '' were not writable. Those are null
strings. These variables were commented out, but needed to be removed
from the list of directories that get checked."


aethelwulffe:
#672

Issue #668, opened 2017-08-21
call_required_libraries is badly implemented (not self documenting)


tmccormi:
In my opinion:
  **This is bad:**  call_required_libraries(true,false,false,false);  //impossible to know at a glance what is being included

  **This better:**  call_required_libraries(bootstrap=true,
                                        fancybox=false,
                                        knockout=false,
                                        datepicker=false);
  
  Or maybe use constants
  add this to header.inc.php
  define (BOOTSTRAP,true);
  define (FANCYBOX,true);
  define (KNOCKOUT,true);
  define (DATEPICKER,true);
  
  Modify the function:
  function call_required_libraries($bootstrap=false,$fancybox=false,$knockout=false,$datepicker=false)
  
  Then use the call like this:
  
 call_required_libraries(BOOTSTRAP);

tmccormi:
It would even nicer if the function was not ORDER dependent in the arguments.

aethelwulffe:
Agreed.

My instinct is to not use constants, just arguments to the function, but 
I think either way is good.  I like the all caps constants pretty 
good...certainly simpler to implement.  This is what I imagined it being 
from the beginning.

aethelwulffe:
function F($params) {
   if(!is_array($params)) return false;
   //do something with $params['A'], etc...
}

You could then invoke it like this:

F(array('C' => 3, 'A' => 1, 'B' => 1));

teryhill:
I would suggest lowercase as opposed to upper case. Some times it just acts funny to have them UC.

tmccormi:
The standard for industry constants is uppercase, just saying.   but as long as it's obvious what is being included I'm good.

aethelwulffe:
Uh, which industry.  +this-is-a-constant+, kThisIsOneToo, so_IS_This_One, AND_THIS_ONE_IS_JAVA.

PHP constant that is used in more than one class is all caps.  The singletons are lowercase...by the recommendation guide.

tmccormi:
This one:  PHP constant that is used in more than one class is all caps. The singletons are lowercase...by the recommendation guide.  

I wasn't aware of the singleton recommendation.  Nice to know

nileshprasad137:
Ok, I'll implement this properly in some time. (Mostly next week)

Trodrige:
@nileshprasad137, @tmccormi I would like to implement this. If it's ok by @nileshprasad137

nileshprasad137:
[...]

PR #674, opened 2017-09-24
Fix for #668, call_required_libraries() is badly implemented (not self-documenting).

Trodrige:
Adjusted call_required_libraries() function's parameters and arguments to be explicit.

### DIFF
--- a/sites/default/sqlconf.php
+++ b/sites/default/sqlconf.php
@@ -1,6 +1,10 @@
 <?php
 //  LibreEHR
 //  MySQL Config
+//  Referenced from /library/sqlconf.php.
+
+global $disable_utf8_flag;
+$disable_utf8_flag = false;
 
 $host	= 'localhost';
 $port	= '3306';
@@ -15,6 +19,7 @@ $sqlconf["port"] = $port;
 $sqlconf["login"] = $login;
 $sqlconf["pass"] = $pass;
 $sqlconf["dbase"] = $dbase;
+
 //////////////////////////
 //////////////////////////
 //////////////////////////
--- a/sites/default/sqlconf.php
+++ b/sites/default/sqlconf.php
@@ -1,16 +1,17 @@
 <?php
 //  LibreEHR
 //  MySQL Config
-//  Referenced from /library/sqlconf.php.
-
-global $disable_utf8_flag;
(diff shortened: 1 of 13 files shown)


sites/default/sqlconf.php
```diff
@@ -1,30 +1,30 @@
...
-//  Referenced from /library/sqlconf.php.
-
-global $disable_utf8_flag;
-$disable_utf8_flag = false;
 
 $host	= 'localhost';
 $port	= '3306';
 $login	= 'libreehr';
-$pass	= 'libreehr';
+$pass	= 'libreehr1';
```
nileshprasad137: This file should not be touched. Please use gitignore.

aethelwulffe: What he said...

templates/practice_settings/general_list.php
```diff
@@ -1,7 +1,7 @@
 <html>
 <head>
 <link rel="stylesheet" href="<?php echo $this->css_header;?>" type="text/css">
-<?php call_required_libraries(true,true,false,false);?>
+<?php call_required_libraries(bootstrap=true,fancybox=true,knockout=false,datepicker=false); ?>
```
aethelwulffe: I guess we are still using the /templates directory?

Trodrige: ok,  I'm looking into it @aethelwulffe

nileshprasad137: Yes, but the smarty templates are removed.

aethelwulffe:
This meets with Nilesh's requirements, and Terry approves.  I am merging.

Trodrige:
@aethelwulffe concerning this "look in the docs (after you have fixed the config file issue in this commit) and make sure the function gets documented." 
Please more light on documenting the function, I don't get it clearly.

aethelwulffe:
@Trodrige  I was just talking about looking through the /docs directory for developer documentation (api) related to this include function, then creating it or updating it to list all the implemented named calls (and order of the same, as I think this is still an ordered list of arguments).  I would dig through for you, or will update it myself later if you like.  I was just putting this out there 'cause I am in a nest of spaghetti code at the moment, and trying to dig myself out with my tiny, tiny dried up brain.

PR #677, opened 2017-09-25
Revert "Fix for #668, call_required_libraries() is badly implemented (not self-documenting)."

aethelwulffe:
Reverts LibreHealthIO/LibreEHR#674


aethelwulffe:
I suppose we can rewind head if we don't like this merge and revert in the system...in the meantime, it is mashing things up....and it is my fault.

PR #678, opened 2017-09-25
Facility based Access Control

aethelwulffe:
Restricts users to patients that have facilites matching their
authorized lists.
This is a work in progress, as it needs some help with the new calendar
functions.


aethelwulffe:
I would like someone familiar with the new calendar to check out the function filter_patient_select_pnuserapi implementation, as I can't locate the corollary in the new calendar.
  This was originally part of Ken's plugin system (not "plugins" which is SMARTY I think).  The functions are from his code, and I kind of ripped them out and tried to use them in a simple way, as the source I stole them from was pretty obfuscated and complex.  Essentially it is just a few extended functions using the WordPress style function/class extender thing do_action() add_action().

teryhill:
I have this code loaded on my test system. I enabled it in globals and nothing blow'd up which is good. How should I test this for the desired results @aethelwulffe

aethelwulffe:
@teryhill Right now it apparently does nothing.
I swiped it from Ken's plugin code, an action for which I am pretty sure I obtained MI2's permission.  The source (BPUC) repo it came from has a different version of the plugin system, the files added were in a dozen different sets of subdirectories, and the commit has a lot of find-and-replace instances of replacing "app" for "bootstrap" in comments that makes it even harder.  On top of that, the whole code-base is drastically different.

So... eventually I hunted down where the actual function changes were, and I added them to this.
Currently, I don't think there is even a layout/patient_data field for "facility" which is required for this.  On top of that, I am not sure that these functions will get called properly.  I am also sure they are not implemented everywhere they need to be.

I popped this branch out as a PR to get a working project going, as I am pretty darn sure that this kind of feature is a vital one to a great number of clinic and even hospital installations.

aethelwulffe:
Progress notes:
I added these functions to the tags system start.php (with some function name changes due to the same name being used in a different way there) and was met with success in the form of no longer being able to view any patients no matter what their facility.

aethelwulffe:
Progress notes:
Got it sort of working...
Functions are firing (after dropping the globals.php include that duplicated them and caused errors), but it seems that the users_facility table is not being utilized properly.
Using the user_admin.php interface with the multi-select facility box, you should be able to update the users_facility table.  The entries should look like ('users','1','1').

I have not figured out all the issues regarding user_admin.php or usergroup_admin.php to see why the the items dealing with "schedule facilities" are not firing when that global is set.  I am updating this branch with what I have (As close to working code as I have it).
Required:  Add the field "facility" to patient_data, and the layout for demographics, have more than one facility, set patients to different facilities in demographics, turn on all related globals, manually add a users_facility table entry with ('users',[users.id],[users.id])

aethelwulffe:
Progress notes:
I have added database elements, and updated to the latest tested code.
Something is still wrong with the user_admin.php features somewhere...I think...

aethelwulffe:
Found the issue.  This branch suffers from the call_required_libraries boo-boo.  Seems to all work when applied to the master branch.

teryhill:
Is the file layout in database.sql changed to be in alphabetical order?

aethelwulffe:
Uh, yeah...I sort of did that...uh...you noticed?

...I will put it back if you like...:-[

teryhill:
What would be the main reasoning for the change, lets see make it easier to find items but you would think that to accomplish that completely the fields should be namef, namel and namem. I think is a good idea maybe not just the right time for it on the patient_data file but other files i would vote for.

aethelwulffe:
Otay...I change it back...to the mess...

tmccormi:
This won't be a problem when we split it onto a separate file for each table ...

sql/database.sql
```diff
@@ -5003,7 +5004,6 @@ CREATE TABLE `patient_data` (
   `state` varchar(255) NOT NULL default '',
   `country_code` varchar(255) NOT NULL default '',
   `drivers_license` varchar(255) NOT NULL default '',
-  `ss` varchar(255) NOT NULL default '',
```
teryhill: Will break code base if removed

sql/database.sql
```diff
@@ -5031,12 +5031,7 @@ CREATE TABLE `patient_data` (
   `billing_note` text,
   `homeless` varchar(255) NOT NULL default '',
   `financial_review` datetime default NULL,
-  `pubpid` varchar(255) NOT NULL default '',
```
teryhill: Will break code base if removed

sql/database.sql
```diff
@@ -5046,21 +5041,6 @@ CREATE TABLE `patient_data` (
...
-  `usertext6` varchar(255) NOT NULL DEFAULT '',
-  `usertext7` varchar(255) NOT NULL DEFAULT '',
-  `usertext8` varchar(255) NOT NULL DEFAULT '',
-  `userlist1` varchar(255) NOT NULL DEFAULT '',
-  `userlist2` varchar(255) NOT NULL DEFAULT '',
-  `userlist3` varchar(255) NOT NULL DEFAULT '',
-  `userlist4` varchar(255) NOT NULL DEFAULT '',
-  `userlist5` varchar(255) NOT NULL DEFAULT '',
-  `userlist6` varchar(255) NOT NULL DEFAULT '',
-  `userlist7` varchar(255) NOT NULL DEFAULT '',
```
teryhill: removing user text and user list will break patient portal

sql/database.sql
```diff
@@ -5031,12 +5031,7 @@ CREATE TABLE `patient_data` (
   `billing_note` text,
   `homeless` varchar(255) NOT NULL default '',
   `financial_review` datetime default NULL,
-  `pubpid` varchar(255) NOT NULL default '',
   `pid` bigint(20) NOT NULL default '0',
-  `genericname1` varchar(255) NOT NULL default '',
-  `genericval1` varchar(255) NOT NULL default '',
-  `genericname2` varchar(255) NOT NULL default '',
-  `genericval2` varchar(255) NOT NULL default '',
```
teryhill: removing genericname and genericval will break code base

sql/database.sql
aethelwulffe: Social Security Numbers break the code?  Where?  Shouldn't that be a Localization thing in any case?
No provision in the Medicare manual for gathering SSN...but I will see what is what.

sql/database.sql
aethelwulffe: ...then I want to figure out where and why.  We agreed to get away from pubpid a long time ago, and I have not used it for years.

sql/database.sql
aethelwulffe: If the portal uses it, I will want to find out where/why.  If it needs fields, I will add them.  No undefined elements should be in the patient_data table, and since this table is user-configurable, things that look at it should not expect to use hard-coded demands for a field to exist, or expect any sequence of results from an unordered array.

aethelwulffe:
I will put these back in...then make another commit to clean it up.  I had hoped that tasks like this were going to get done with the related project, but it has turned into a Laravel thing.

tmccormi:
Many of my customers use pubpid extensively for what it was intended for, reference to an external system ID.   Some use it to create "meaningful, manually created patient id" such as MCC20170405 (name part + reg date).   We need that capability still.

tmccormi:
Seem to be mixing purposes in the commit again, ie: cleanup + new functions ...

aethelwulffe:
Sometimes, you can't do work in a manure pile...and you have to clear some space.

sql/database.sql
teryhill: Just a few
image

sql/database.sql
aethelwulffe: Yeah, I knew about the dupecheck.  That was one of the things that always made using the merge patient such a pain in the ass.  People would use the ssn field for all kinds of stuff, then have to open up each patient, edit both, run the merge again etc...

aethelwulffe:
All fields are back in this branch.  attacking it a different way.  This may mess with Pri, but I think some of this needs to move.  Yes, I know I am grumpy today.  Back pain and all.

sql/database.sql
aethelwulffe: Heavy sigh...
Otay.
But it should not be REQUIRED anywhere.  External ID's should be a vert list.

tmccormi: I agree with @aethelwulffe on this, no apps should rely on fields that the user can change and reuse.

aethelwulffe:
...And vital fields should not be modified by users...

tmccormi:
I do not disagree that pubpid is used incorrectly in the application and sometimes mis-used.   Would need a migration to move the current pubpid to an external vertical table (which would allow many external IDs if a clinic needed them...)

aethelwulffe:
NOTE:  I am adding the WIP tag to this.  IT MUST BE HEAVILY TESTED.  The method of assigning users to facilities, and the way access is granted when creating a new facility is very disjointed.  Also, a patient MUST be assigned to a default facility when the code is implemented against an existing database, and all users must be assigned in some similar way.
The user admin modal window does a very poor job of this.  We need a good administrative tool (spreadsheet like) that deals with it.

sql/database.sql
teryhill: They can change and reuse any field in the layout form, not just these

tmccormi:
@kchapple - please look this over as it it mostly based your your work as I recall...

aethelwulffe:
Yeah, I basically pulled the functions out of his work and located them with the existing tags stuff.  It has all been very "experimental".  Translation:  I personally desperately need this to work, so I am trying anything I can....

sql/database.sql
aethelwulffe: Understood.  Layouts is just a bit too powerful in that regard.  Should set some minimal restrictions.  That has been a problem in the past (good old billing note for one) and has resulted in many broken systems.

aethelwulffe:
Use of PubPID:
Still not a great idea for normal use, as mostly folks still wind up putting in values that equate to PHI.  `pid` is not a row index.  If someone really wants to be able to manually set it to something alphanumeric (but unique in the system) that should be allowed, but after training folks away from thinking it is necessary (search features are better), they have never needed the crutch again.  Even now they insist on sending me messages like "Please find the note I never saved for patient JS 35654."  The initials are useless.  They just can't get the idea that you don't need them.  The problem is, they just need to start thinking of people as just numbers!

aethelwulffe:
I suggest we hang on for just a bit with hopes that Ken will chime in.

kchapple:
This might be a stupid question, but what does this have to do with tags?

aethelwulffe:
Nah, facility access control you set up for BPUC a couple of years ago.  It is crammed into the tags stuff with a couple of changes to the functions names (had same names for stuff in tags as in the facility ACL things you did for them.)

tmccormi:
It only relates in that @kchapple did the original work and it's "plugin" style LIKE the Tags stuff ...

kchapple:
@tmccormi but it's actually implemented within the tags plugin, which seems weird if it doesn't related to tags. Maybe it should be implemented directly in the code base, especially if Art is making DB changes and hard-coding globals. So, if the Tags plugin is removed (can be deleted or the reference taken out of the composer.json file) the removal will probably break this new code since he hard-coded a global for it. @aethelwulffe you could add the global using the dynamic method to prevent this.

aethelwulffe:
I gotta figure out what "using the dynamic method" means.  Maybe you are talking about the      $this->user_specific_globals = $args['user_specific_globals']; stuff.
I think I am mostly worried about the saveAsUserSettings() related stuff.
I have puzzled over what, (for most installations), the createMigrationsTable() functions and any of the rest of this does.
I think I need to spend a week or two with a whiteboard figuring out the difference between addAction and add_action and all this other very cool stuff.  It is hard when you don't have any idea of what names like psr-4 are for. :)
While implementing it as a plugin (separate and distinct) would be just great, there was the problem that I couldn't really find out a lot about the plugin system itself...not in noob terms anyway.  I have had to try to track down the action of each item, THEN try to figure out what it does and why.  I don't know what the .  It is some rough reverse engineering, and I have not been inspired enough to break out a white-board.

  The other issue is that the Tags plugin, and the Facility ACL plugin were written with identical function names, yet different functions.  As the calls (add_action, do_action) were mostly all hooked into the right places for the standard top-level menu items, I figured continuing to use those were the smart thing to do.  Unfortunately, I was missing some important bit up the public static function tree, and nothing was firing right.  About the time I figured out that the function names had already been declared elsewhere, I realized the whole mess actually came down to adding a small handful of functions somewhere...so I just piggybacked them into /tags.
  So...I would like to get this done "Right" in terms of the author of this plugin system...and will take any further advice you give.

aethelwulffe:
🐢

aethelwulffe:
Should I go ahead and close this PR?

aethelwulffe:
OK, so no Access controls added to Libre this year.
I am closing this ancient PR by Sunday.  (damn).

PR #694, opened 2017-09-29
Updates part deux

tmccormi:
My adds and tweaks


aethelwulffe:
Fine, but www.bowenmd.com is a bad link.

PR #696, opened 2017-10-03
Remove the entries for role based menus till an acceptable editor is in place

teryhill:
Remove the entries for role based menus


tmccormi:
Any chance these menu items can be added to the JSON file or is that already done? @teryhill

teryhill:
I am not sure what you are asking. The menu items should be in both files since you and Art don't use the Tables.

tmccormi:
The Role based menu items is what I was asking about, specifically

teryhill:
I don't know . Dr Ray did it based on the tables not the Json file. Just put his commit up there.

PR #697, opened 2017-10-04
Some menu and globals cleanup

teryhill:
Some menu and globals cleanup. Took out some left nav specific stuff (tall nav, hide menu etc). Added more checks for existing globals.

### DIFF
--- a/interface/main/main_screen.php
+++ b/interface/main/main_screen.php
@@ -92,8 +92,6 @@ else {
   $frame1url = "TAB_ONE_DEFAULT"; 
 }
 
-$nav_area_width = '130';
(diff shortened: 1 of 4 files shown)


teryhill:
@tmccormi I noticed there was no entries for the new crop code in the json file , is this an over site or is it intentional?

tmccormi:
Kevin created that data from Left Nav, so it should have been there.  It's not supposed to show if the Global is not enabled.  I have not looked at it since I have no production LibreEHR with NewCrop.  I'll open an issue for it.

aethelwulffe:
Well cool  OUt of my hands (woo-hoo!)

aethelwulffe:
Branch still live.

PR #699, opened 2017-10-05
First pass at the menu for new crop

teryhill:
First pass at the menu for new crop


teryhill:
Works now good call @aethelwulffe

PR #701, opened 2017-10-05
Replacement for original mips PR

aethelwulffe:
Updated old branch by replacing it.  This is the new version of the
MIPS/MACRA branch, required due to being 4 months stale, and the review
having about 100 xlt() change requests. :smiling_imp:


tmccormi:
Is there a reason for some many changes to GACL ?  docs/ etc?

tmccormi:
This seems to be a mixed commit that includes removing gacl and zend modules.  Can we get a commit that just has MIPS/MACRA stuff?

aethelwulffe:
It seemed like

1.  It was a huge commit anyway.

2.  I am adding 2700 files...might as well get rid of some trash.

I can make a series of PR's, and eventually maybe my production stuff 
might be able to run off of this repo, but since it was several times 
the amount of work to undo work then redo work, then maintain two repos, 
I figured I would try it this way first.

The impetus to get rid of a lot of it was due to the fact that a. It was 
pure bloat. b. the stuff doesn't work. c.  There were tons of deprecated 
bits and missing dependencies in a lot of it that occasionally clogged 
up the error logs. And d. There were functions that are currently 
considered insecure.

If putting all the stuff back in is the available option, I shall proceed.

tmccormi:
This is fine, just need to know the why... mixing commit reasons is not a very clean or historically searchable model.

aethelwulffe:
I know, and I hate it.

This is an example of actually merging two code-bases vs. developing 
something new.  We should not have desire to do this much in the 
future.  I want to work with ONE repo and ONE code base.  Never again 
should I feel like I need anything not in the master branch and be 
tempted to "tweak" more than adding something to a little hook.

aethelwulffe:
I am running up a set of ACL checks in the menu JSON.  There are pieces of this, namely the patient data saver/loader/importer and the like that should only be enabled if it is a demo system.  The other option is for those pieces to either be pulled out or made a separate download thing.  Probably best to do the config route, especially if we are going to be doing an infomatics/edu thing.

aethelwulffe:
We need some ACL checks in the menu and other spots.  I will get to that immediately.

tmccormi:
@teryhill  please check an approve this for merge ...

PR #703, opened 2017-10-09
Display SQL Errors in Installer

yehster:
When execute SQL fails echo the message.


yehster:
Errors show up during the cloning process.

PR #704, opened 2017-10-10
MyISAM engine declarations

aethelwulffe:
Remove MYISAM engine declarations

### DIFF
--- a/interface/forms/ankleinjury/table.sql
+++ b/interface/forms/ankleinjury/table.sql
@@ -24,4 +24,4 @@ ankle_diagnosis4 varchar(255),
 ankle_plan longtext,
 ankle_additional_diagnisis longtext,
 PRIMARY KEY (id)
-) ;
+) ENGINE=InnoDB;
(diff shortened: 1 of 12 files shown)


teryhill:
Instead of deleting it should you replace it with InnoDB? Just asking, not sure it matters.

tmccormi:
Yes it should have the INNODB declaration, otherwise it will default to whatever the database was created as. Not an issue with "default" setup but could cause issues if someone manually pre-creates the DB and doesn't set the default INNODB.

Also some of the tables were not working in INNODB to to unsupported types, where those issues address? @yehster know the most about that from the Peace Corps project.

aethelwulffe:
https://mariadb.com/kb/en/library/converting-tables-from-myisam-to-innodb/
I use the above bits of advice.
These forms are actually fine either way.  They have nothing special, and have only a single key.  It is listed first, declared specifically, and is otherwise well behaved.
I will specify the engine.

tmccormi:
Nice resource.

aethelwulffe:
Added a commit.

Issue #707, opened 2017-10-10
INNODB server variables setup recommendations.


aethelwulffe:
Recently posted this:  https://mariadb.com/kb/en/library/converting-tables-from-myisam-to-innodb/

There is a lot of good advice on table conversion, and a lot of points made concerning "gotchas" we experienced, as well as how to avoid those.  At this time, the server variable recommendations we make in the setup are not appropriate for running the INNODB engine.  Memory is handled very differently in INNODB vs MyISAM.  Many of the "silent fails" we have seen on setup are due to insufficient settings for a production instance.  Apache has hilariously conservative default settings.
  Of immediate note:
    key_buffer_size — small but non-zero; say, 10M;
    innodb_buffer_pool_size — 70% of available RAM

tmccormi:
I'll update the setup notes.

tmccormi:
added this:
Configuration of PHP...

We recommend making the following changes to your PHP installation, which can normally be done by editing the php.ini configuration file:

Your php.ini file can be found at /etc/php5/apache2/php.ini

To ensure proper functioning of LibreHealth EHR you must make sure that settings in php.ini file include:

- "short_open_tag = On"
- "display_errors = Off"
- "register_globals = Off"
- "max_execution_time" set to at least 60
- "max_input_time" set to at least 90
- "post_max_size" set to at least 30M
- "memory_limit" set to at least "128M"
- "key_buffer_size" set to 10M
- "innodb_buffer_pool_size" set to 70% of available RAM.

In order to take full advantage of the patient documents capability you must make sure that settings in php.ini file include:

- "file_uploads = On"
- "upload_max_filesize" is appropriate for your use (32M seems good)
- "upload_tmp_dir" is set to a correct value that will work on your system.

aethelwulffe:
We should include (in docs) recommendations for TEMPORARY settings and for DEVELOPER settings for things like large database imports, conversions, running super-huge reports and the like.
Personally, I think that system percentages are better recommendations a lot of the time.

Another one of my favorite textbytes for server-side stuff:
http://www.bigeng.io/php-memory-optimization/
These nice little functions figured out (over a couple of weeks) that for the ARSP clinic, we were better off (calculated exacts) with
 max_execution_time =122
max_inpute_time = 160
post_max_size = 52M
memory_limit = 512M 
...that is/was running OEMR 4.2 on a single core (all it could use) with total system memory of 8GB.
[...]

PR #712, opened 2017-10-11
Updates to setup for INNODB and Documentation #707

tmccormi:


setup.php
```diff
@@ -588,9 +605,9 @@ function cloneClicked() {
 echo "<p>Welcome to LibreHealth EHR.  This utility will step you through the installation and configuration of LibreHealth EHR for your practice.</p>\n";
 echo "<ul><li>Before proceeding, be sure that you have a properly installed and configured MySQL server available, and a PHP configured webserver.</li>\n";
 
-echo "<li>Detailed installation instructions can be found in the <a href='INSTALL' target='_blank'><span STYLE='text-decoration: underline;'>'INSTALL'</span></a> manual file.</li>\n";
+echo "<li>Detailed installation instructions can be found in the <a href='Documentation/1_Installing/InstallingLibreEHR.html' target='_blank'><span STYLE='text-decoration: underline;'>'Installation'</span></a> manual file.</li>\n";
```
aethelwulffe: We could just do includes or iframes in hidden divs with all the expanded information in it right there instead of giving links that cause tab-hogging.
`
    <a href="javascript:toggleDiv('main');" style="background-color: #ccc; padding: 5px 10px;">Show Me More</a>

    <div id="main" style="background-color: #F0F0F0; display: none; padding: 5px 10px;"><br>
<?php
include_once "whatever url";
?>
</div>`

teryhill:
Put it on the about page where the documentation is started.

setup.php
tmccormi: I guess that's a matter of opinion, I MUCH prefer to have a new Tab for docs not a split frame.

tmccormi:
@teryhill not sure I understand your comment...

PR #713, opened 2017-10-11
syntax errors in file attribution blocks

aethelwulffe:
Extra character in comments broke a lot of code.

### DIFF
--- a/library/classes/rulesets/PQRS/reports/PQRS_0403/Denominator.php
+++ b/library/classes/rulesets/PQRS/reports/PQRS_0403/Denominator.php
@@ -15,7 +15,7 @@
  * @link    http://librehealth.io
  *
  * Please support this product by sharing your changes with the LibreHealth.io community.
- */
+ *
  * @author  Suncoast Connection
  */
  
--- a/library/classes/rulesets/PQRS/reports/PQRS_0403/Exclusion.php
+++ b/library/classes/rulesets/PQRS/reports/PQRS_0403/Exclusion.php
@@ -15,7 +15,7 @@
  * @link    http://librehealth.io
  *
  * Please support this product by sharing your changes with the LibreHealth.io community.
- */
+ *
  * @author  Suncoast Connection
  */
 
--- a/library/classes/rulesets/PQRS/reports/PQRS_0403/InitialPatientPopulation.php
+++ b/library/classes/rulesets/PQRS/reports/PQRS_0403/InitialPatientPopulation.php
@@ -15,7 +15,7 @@
  * @link    http://librehealth.io
  *
  * Please support this product by sharing your changes with the LibreHealth.io community.
- */
+ *
  * @author  Suncoast Connection
  */
  
--- a/library/classes/rulesets/PQRS/reports/PQRS_0403/Numerator.php
+++ b/library/classes/rulesets/PQRS/reports/PQRS_0403/Numerator.php
@@ -15,7 +15,7 @@
  * @link    http://librehealth.io
  *
  * Please support this product by sharing your changes with the LibreHealth.io community.
- */
+ *
  * @author  Suncoast Connection
  */
  
--- a/library/classes/rulesets/PQRS/reports/PQRS_0403/PopulationCriteria.php
+++ b/library/classes/rulesets/PQRS/reports/PQRS_0403/PopulationCriteria.php
@@ -15,7 +15,7 @@
  * @link    http://librehealth.io
  *
  * Please support this product by sharing your changes with the LibreHealth.io community.
- */
+ *
  * @author  Suncoast Connection
  */
  
--- a/library/classes/rulesets/PQRS/reports/PQRS_0404/Denominator.php
+++ b/library/classes/rulesets/PQRS/reports/PQRS_0404/Denominator.php
@@ -15,7 +15,7 @@
  * @link    http://librehealth.io
  *
  * Please support this product by sharing your changes with the LibreHealth.io community.
- */
+ *
  * @author  Suncoast Connection
  */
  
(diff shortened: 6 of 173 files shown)


aethelwulffe:
Now with premeasure fixes.

teryhill:
@aethelwulffe  does this look right to you?

image

aethelwulffe:
No.  That does not look right.  That is the CQM stuff, which is not even in the MIPS version.

teryhill:
@aethelwulffe it is clobbered on the Demo. You need to look at it and see if you can figure out what is going on.

teryhill:
I am getting this on my system

ERROR: query failed: SELECT * FROM `clinical_rules` WHERE `pid`=0 AND `active` = 1 ORDER BY `id`

Error: Unknown column 'active' in 'where clause'

teryhill:
Should the main menu have changed?

image

aethelwulffe:
1.  The demo doesn't have the MIPS tables installed, and the menu seems to not have the right JSON or is running off of tables.
2.  The CDR report should be turned off, along with AMC and Standard Reports.  They are invalid in any case.  The MIPS stuff is in place of all that.
3.  The MIPS tables are located in LibreEHR\modules\MIPS\SQL\1_INSTALL.  Until I can get an autoconfigure thing put into the system, you have to install the module tables manually.
4.  Without the right menu, and the correct copy of the globals.inc.php code, you can't manage any of the settings right.
Demo is really messed up in a number of ways besides.

aethelwulffe:
I am doing, and documenting a full clean install of everything right now.  It will not include these fixes,  but that should not matter unless I try to run rule-set higher than /# 400

teryhill:
The json file is missing the last 15 lines so this may be causing the menu to be wrong.

image

aethelwulffe:
The demo is not up to date.

teryhill:
The Demo is running off the JSON file

aethelwulffe:
A lot more than 15 lines...
None of the MIPS globals (which are called 'pqrs') seem to be in that file.  If you are looking at the JSON from the web, and those globals are not turned on, then you would not see the entries listed.
Not sure when you are talking about your system, or the demo...

teryhill:
My system which I have the current codebase on.

aethelwulffe:
Do we have any requested changes for this branch?

aethelwulffe:
My new install gets no menu at all.

I replaced it with a different menu.json known to work, and that has no effect whatsoever.

aethelwulffe:
With no menu available, I was able to get to user preferences.  I changed the "role-based menu" type...no change in appearance.

teryhill:
It is running off the JSoN file. I get the no menu with the codebase JSoN file

aethelwulffe:
MIPS_Gateway repo works, and is not that far separated.  Gonna do a diff.

aethelwulffe:
Shouldn't 
` "global_req" : [ "erx_enable", "newcrop_user_role" ]`
be
 "global_req" :  "erx_enable", "newcrop_user_role"

teryhill:
No .... Adding the PQRS stuff is it issue. I am working on fixing it

aethelwulffe:
Must be the ACL stuff implementation then, 'cause everything works on our sites...and there is virtually no difference in the code of the two repos except for eRX, the Facilities bit and a couple of other things...unless you are talking about globals.php

aethelwulffe:
And thank you!

teryhill:
sending a fix now

aethelwulffe:
I checked it out and merged it.
That was the thing in the MIPS branch where I was a little shocked when I pushed a commit before testing, then saw someone merged it immediately after.
Thanks for fixing my screwup.

teryhill:
Need the active field addec to clinical rules table.

aethelwulffe:
Gotcha...I suppose.  I know I intentionally removed it, so I should 
check on all that.

aethelwulffe:
Actually, I just need to put in a stub clinical rules table into database.sql.
It is only needed for these reports, unless you create a custom rule of some kind.
Issue goes away if you install the table...
Structure is the following:
```
CREATE TABLE `clinical_rules` (
  `id` varchar(35) NOT NULL DEFAULT '',
  `pid` bigint(20) NOT NULL DEFAULT '0' COMMENT '0 is default for all patients, while > 0 is id from patient_data table',
  `active_alert_flag` tinyint(1) DEFAULT NULL COMMENT 'Active Alert Widget Module flag - note not yet utilized',
  `passive_alert_flag` tinyint(1) DEFAULT NULL COMMENT 'Passive Alert Widget Module flag',
  `patient_reminder_flag` tinyint(1) DEFAULT NULL COMMENT 'Clinical Reminder Module flag',
  `release_version` varchar(255) NOT NULL DEFAULT '' COMMENT 'Clinical Rule Release Version',
  `web_reference` varchar(255) NOT NULL DEFAULT '' COMMENT 'Clinical Rule Web Reference',
  `access_control` varchar(255) NOT NULL DEFAULT 'patients:med' COMMENT 'ACO link for access control',
  `pqrs_code` varchar(35) DEFAULT NULL COMMENT 'Measure number',
  `pqrs_individual_2016_flag` tinyint(4) DEFAULT NULL COMMENT 'Is MIPS flag',
  `pqrs_group_type` varchar(2) DEFAULT 'X' COMMENT 'XML output scheme type',
  `active` tinyint(4) DEFAULT NULL COMMENT 'Is this measure turned on?');
```

aethelwulffe:
I also need to add in all the web references.  They are available...

teryhill:
@aethelwulffe  I just pushed 2 commits to fix some date stuff in MIPS you need to look at the clinical reports file and fix the checks you are doing on the date.

aethelwulffe:
@teryhill are the fixes that are actually in this PR OK with you Terry?

aethelwulffe:
Yeah, my brain is gonna pop any second now... 🔨

teryhill:
Looks good to me merging

PR #715, opened 2017-10-12
fix the json file missing commas

teryhill:
fix the json file missing commas


tmccormi:
I see you did it :-)

aethelwulffe:
Yeah, sorry...demo might catch this one.

PR #716, opened 2017-10-12
date picker css fix

teryhill:
date picker css fix in Mips

### DIFF
--- a/modules/MIPS/clinical_measures.php
+++ b/modules/MIPS/clinical_measures.php
@@ -117,6 +117,8 @@ if(!empty($report_id)) {
   $page_subtitle = '';
   $dis_text = '';
 }
+$DateFormat = DateFormatRead();
+$DateLocale = getLocaleCodeForDisplayLanguage($GLOBALS['language_default']);
 
 $widthDyn = '470px';  //determine what is needed for pqrs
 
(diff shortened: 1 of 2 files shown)


aethelwulffe:
Terry listed some things that still needed to be fixed.  Now I understand what that should be.

modules/MIPS/clinical_measures.php
```diff
@@ -315,7 +318,28 @@ function Form_Validate() {
                     <td class='label'>
                       <?php echo htmlspecialchars(xl('Begin Date'), ENT_NOQUOTES); ?>:
                     </td>
```
aethelwulffe: Oops.  Two begin dates.

tmccormi: Should I revert this one?

teryhill:
@aethelwulffe and I talked about it I did it in a hurry (had to get to the drying apples) he will fix it and send the fix.

aethelwulffe:
Not totally sure it has to be fixed.  Currently, the case is always that 
the report type = 'pqrs'

The other condition is as to what is done when viewing a completed 
report, and I need to test this to see if it makes any real difference.

I will get that answer done one way or another.  In the meantime, no 
need to revert.

PR #728, opened 2017-10-17
Update xampp install docs

aethelwulffe:
Simple cleanup.  All better now.

### DIFF
--- a/Documentation/1_Installing/InstallingLibreEHR.html
+++ b/Documentation/1_Installing/InstallingLibreEHR.html
@@ -21,7 +21,7 @@ you have run the installer, please follow these steps:
 
 <TR>
 <TD>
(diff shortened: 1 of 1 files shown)


tmccormi:
I think this should be renamed to XAMPP install and/or merged with INSTALL.MD but we can do that in another release...

tmccormi:
note: change merged with libreehr_rel200 branch as well

aethelwulffe:
xampp is meaningless to even the folks that download the xampp package...so it would be "windows" not xampp

  Aaaand....The basic walk-through images (which we can do drop-in replacements of) are really the same for anything...and should be the same for any set of documentation....aaaaand Theeeeeeenn?  Andthenandthenandthen!!!

PR #732, opened 2017-10-19
Configure app title

aethelwulffe:
`<title><?php echo $GLOBALS['libreehr_name'];?></title> `allows user to
identify what system they are in....vs guessing if they have more than
one system.


teryhill:
While you are in there tinkering see what it would take to give the user a notification that they have a new message. 

Why the change ?

aethelwulffe:
Example:

image

tmccormi:
great. I usually make the first name of the admin user the 'system' for that reason ...

PR #734, opened 2017-10-20
Minor MIPS patch

aethelwulffe:
Added a closing html tag  to one file, added linked MIPS documentation.  TODO list will add a global config URL base to link documentation to CMS website or localhost.  Currently hard-coded to the SunCoast Connection website, which is the only clean set of documentation files we know of.

### DIFF
--- a/modules/MIPS/import_data.php
+++ b/modules/MIPS/import_data.php
@@ -34,20 +34,64 @@ $query = file_get_contents($GLOBALS['OE_SITE_DIR']."/filemanager/files/Importer/
 sqlStatementNoLog($query);
 
 sqlStatementNoLog("TRUNCATE TABLE `billing`;");
+if ( file_exists($GLOBALS['OE_SITE_DIR']."/filemanager/files/Importer/billing.sql")) {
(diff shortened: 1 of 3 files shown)


modules/MIPS/import_data.php
```diff
@@ -34,20 +34,64 @@
 sqlStatementNoLog($query);
 
 sqlStatementNoLog("TRUNCATE TABLE `billing`;");
+if ( file_exists($GLOBALS['OE_SITE_DIR']."/filemanager/files/Importer/billing.sql")) {
 $query = file_get_contents($GLOBALS['OE_SITE_DIR']."/filemanager/files/Importer/billing.sql");
 sqlStatementNoLog($query);
+}
+if ( file_exists($GLOBALS['OE_SITE_DIR']."/filemanager/files/Importer/billing1.sql")) {
```
tmccormi: What do this files do?

aethelwulffe:
They allow you (when the PQRS/MIPS 'demo' global is set) to import data 
sets.  Another tool allows you to save and switch between data sets.  
This is done within a single database vs. changing databases.  It can be 
used to save work against a data set, load data from a different 
specialty clinic, whatever, and done at the user level.

PR #740, opened 2017-10-25
Remove 2017 ICD10 so 2018 can load

tmccormi:
In testing I could not load ICD10 due to 2017 and 2018 both being in the list.


aethelwulffe:
Yeah, that is another issue.  Either you need the DB for both, or get 
rid of the old files.

tmccormi:
When upgrading, we really should figure out how to load an "new data set" without removing the old so that claims/billing stay connected to their code descriptions, etc ...

aethelwulffe:
If the files have the correct (old) version files identified, then you can have both versions in there.  Of course we don't want to retain outdated stuff either though.

PR #745, opened 2017-10-26
MIPS Medicare Update

aethelwulffe:
1.  Provides default date global for direct entry encounters when in
reporting mode (to be implemented later).
2.  Removes (temporarily) clinical rules check for Medicare.  This may
be required conditionally for 2018 MIPS standards.

### DIFF
--- a/library/clinical_rules.php
+++ b/library/clinical_rules.php
@@ -697,11 +697,11 @@ function test_rules_clinic($provider='',$type='',$dateTarget='',$mode='',$patien
   }
 
   // Collect applicable patient pids in only medicare
-if (strpos($type, 'pqrs_individual') !== false ) {
-	$onlyMedicarePatients=true;	
-} else {
+//if (strpos($type, 'pqrs_individual') !== false ) {
+//	$onlyMedicarePatients=true;	
+//} else {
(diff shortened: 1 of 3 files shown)


library/clinical_rules.php
```diff
@@ -697,11 +697,11 @@ function test_rules_clinic($provider='',$type='',$dateTarget='',$mode='',$patien
   }
 
   // Collect applicable patient pids in only medicare
-if (strpos($type, 'pqrs_individual') !== false ) {
-	$onlyMedicarePatients=true;	
-} else {
+//if (strpos($type, 'pqrs_individual') !== false ) {
```
tmccormi: Will these need to be added back in?  If not let's just drop the comment out lines...  or make a note about why they are being kept

aethelwulffe: Actually, we will have to revamp an number of these things immediately for ACO accessibility.  Working on it.

aethelwulffe:
It has been a long time since I tested.  If Terry just tested, then it is fine.  Yes, I am working up a large related set of changes to the module.  I am including the "move out of old patient selector" requirement for #860 #866

PR #746, opened 2017-10-27
Addresses #741, fixes layout on encounter screen

nileshprasad137:
Hope this fixes the layout issue.


aethelwulffe:
I hate to see what are essentially reversions, but this has the desired behavior.
I always wanted to revise this whole layout.  It works as is, but there are certain user behaviors (like putting a whole note in the encounter reason) that I always wanted to dissuade.
Looks good.

tmccormi:
Sofry I missed this one.

PR #747, opened 2017-10-27
Updated README.md

SoumyadeepJana:


README.md
```diff
@@ -29,7 +29,7 @@ For detailed and step-by-step Installation Instructions refer [Installation Inst
 
 ## Windows :: 
 
-Firstly make sure that you have [WAMP](http://www.wampserver.com/en/) or [XAMPP](https://www.apachefriends.org/index.html) server installed and the time zone is set correctly.
+Firstly make sure that you have [WAMP](http://www.wampserver.com/en/) or [XAMPP](https://www.apachefriends.org/index.html) server installed with php 7.0.24 or lower and the time zone is set correctly.
```
aethelwulffe: either "and that the time zone is set correctly" or  "and the time zone set correctly."
Also, it is not php 7 or lower.  It is (I think) a range of something like 5.7 - 7.0.24.  We need input on that.  Specifically mentioning that 7.1 is incompatible might be good.

SoumyadeepJana: Does it work with 5.6?Most of the wamp/xampp available to download are shipped with as low as 5.6

aethelwulffe: Yes, 5.6 works...I think!
Brings to mind the fact that we still have some examples of split() in the code base somewhere...

SoumyadeepJana:
@tmccormi I have pushed to my forked repo the updated README as well as some of the UI Enhancements tasks.Should I make another PR with all the changes?

PR #753, opened 2017-11-01
Changed the interval and added a no refresh on the calendar

teryhill:
Made the changes as a result of the check for new appointments being
fired to often

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -128,6 +128,7 @@ $USER_SPECIFIC_GLOBALS = array('default_tab_1',
                                'ledger_begin_date',
                                'print_next_appointment_on_ledger',
                                'calendar_view_type',
+                               'calendar_refresh_freq',
                                'event_color',
                                'pat_trkr_timer',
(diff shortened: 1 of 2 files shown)


teryhill:
@azwreith check out the discussion on the forums https://forums.librehealth.io/t/problems-on-ehr-demo-sites/1186

aethelwulffe:
Awaiting other requested reviews...

azwreith:
Sorry for being gone so long, it's my last semester of college and I have a lot of stuff to get done.

Ohh I see, that constant polling might indeed be an issue. In that case setting the auto refetch rate longer and refetching the events every time a new event is added is a better idea.

aethelwulffe:
I honestly feel that every 3-5 seconds as a floor() is waaaay enough.  This is a web script, thingie, not an FPS.

tmccormi:
No frontdesk runs so fast that 60 secs is too long for the refresh.  A manual refresh if desired is also always possible.

PR #754, opened 2017-11-04
auto logout fix

teryhill:
auto logout fix


tmccormi:
Also applied this fix to the libreehr_rel200 branch

PR #759, opened 2017-11-16
Update calendar scheduler source from 1.62 to 1.90

teryhill:
Update calendar scheduler source from 1.62 to 1.90 and full calendar to
3.70. This keeps the core code of the calendar module up to date.


teryhill:
@tmccormi this just updates the calendar code to keep it updated.

PR #764, opened 2017-11-22
Fix clinical rules error

teryhill:
Fix clinical rules error


robbyoconnor:
Was this tested?

aethelwulffe:
Yes, and functions as advertised.

robbyoconnor:
The issue was that it did not run correctly, hence my question to whether it was tested.

Issue #683, opened 2017-09-25
Tab divider overlaps encounter list.


aethelwulffe:
![image](https://user-images.githubusercontent.com/553751/30791657-fb53609e-a181-11e7-855c-a46cd70e834c.png)

When looking at the full list (hover view, with encounter review) it is very visible.

PR #771, opened 2017-11-28
Fix: tab divider overlays encounter list

Trodrige:
Fix for #683
Gave encounter list div large z-index value

### DIFF
--- a/interface/main/tabs/css/tabs.css
+++ b/interface/main/tabs/css/tabs.css
@@ -202,6 +202,11 @@ div.closeDlgIframe {
     display: block;
 }
 
+.patientDataColumn > div
+{
+    margin-top: -1%;
+}
+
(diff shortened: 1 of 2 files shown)


Trodrige:
selection_018

aethelwulffe:
Not that I am not please to receive...but how about we think about adding all this stuff horizontally?
Conditionals:  
1.  In an encounter?  Print "Now showing Encounter 2017-12-02"
2.  Has past encounters?  Show "Past Encounter List".
After that, have a [Create Encounter] button.

-Otherwise, this is still a much needed PR, and removes a LOT of PAIN when trying to click on a tab!
-Good work man!

Trodrige:
Yup, this will help if the list is really long.

3. Has no encounter? Show "Create Encounter" button

Issue #779, opened 2017-12-01
Add new message icon to the main screen.


teryhill:
Add a notification icon to the main screen. This icon will indicate if the signed in user has a new message in the messaging system. When there is an new message for a user display a notification icon in the upper right side of the main screen.

Changes to /libreehr/interface/main/tabs/main.php to display the notification icon.

![notify](https://user-images.githubusercontent.com/4655366/33466975-7badcdb6-d61f-11e7-909e-a8ac34c0c4cc.PNG)

teryhill:
@robbyoconnor if you can figure out how to get this in as one of Your task it might give you some help while you are making more coding tasks for LibreEHR

teryhill:
Currently assigned to @migdinny

teryhill:
Assigned too @migdinny

robbyoconnor:
@teryhill There's a generic fix an issue -- the issue can be difficult -- so it's okay to use that one.

aethelwulffe:
http://librehealthehr.suncoastconnection.com/refresh
A little demonstrator...

teryhill:
This is done and the PR's are merged.

PR #784, opened 2017-12-02
Unread messages notification created (issue #779)

MigDinny:
If there are unread messages, a notification on upper right corner will appear showing "5 unread messages" for example. If there are no unread messages, no notification will appear.

Messages notification have an absolute position with 50px right. This design isn't responsive and I think it's not a good practice, but the userData button was design this way, so I followed.

This notification ONLY refreshes when the page refreshes (opening, refreshing or closing tabs DO NOT refresh the page, therefore, it's completely predictable that the notification remains the same). There are probably workarounds to fix this.

Suggestion for the future: when there are unread messages, put a number before **Messages** menu option like "(2) Messages" instead of this feature. It might be more appealing but it's not currently and technically possible for me to do it now. Tried a lot of ways but all of them failed.

#779

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,0 @@
-
-sites/default/sqlconf.php
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+
+sites/default/sqlconf.php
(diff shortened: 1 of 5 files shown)


MigDinny:
image

On the upper right corner there is the notification.

teryhill:
@nileshprasad137 @aethelwulffe  can y'all look at this and test it. It looks good to me except there is an alignment issue on my system. Firefox works fine chrome has the issue.

image

PR #795, opened 2017-12-03
Added "If you don't find the sql-mode in my.ini"

alpinus4:
I added what should you do if you don't find "sql-mode" parameter in my.ini file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ In Linux it's typically located in /etc/mysql
 
 Restart WAMPP/XAMPP Server.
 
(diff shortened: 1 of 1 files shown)


aethelwulffe:
@alpinus4 Please confirm which PR's are good...and if these two extras (apparently) are the same as the approved one.  Someone made things a little messy :)

alpinus4:
The PR's #795 and #796 are bad (small language mistakes), and the PR #801 is the same as PR's #795 and #796 (with small corrections) :)

alpinus4:
Please, close PR's #795 and #796

Issue #778, opened 2017-12-01
Add Insurance address information to demographics report


teryhill:
User requested the Insurance address information on the report

Before

![image](https://user-images.githubusercontent.com/4655366/33464758-08933f9e-d612-11e7-94e6-609707cfb635.png)

After

![image](https://user-images.githubusercontent.com/4655366/33464712-bf29800c-d611-11e7-9395-05311ddb239a.png)

Jdew192837:
I plan to work on this as part of GCI

teryhill:
@Jdew192837 this is already done it is just waiting on the core group to vote on whether it is a good thing to add.

htuck:
I don't know if I get a vote but it seems like a good idea to me.  Whoever this report goes to probably would want the real ins co information instead of just what the clinic calls it internally.

aethelwulffe:
Dude.  Add it.
I have added this as custom code in the past.

aethelwulffe:
So long as it has a globals toggle, it is great.

teryhill:
PR Merged Issue closed

PR #810, opened 2017-12-04
Add the insurance company name to the demographics report

teryhill:
Add the insurance company name to the demographics report #778

### DIFF
--- a/library/report.inc
+++ b/library/report.inc
@@ -469,7 +469,14 @@ function getRecInsuranceData ($pid, $ins_type) {
     //column name->dates->values
     //$return{"lname"}[0..n]{"date"}
     //$return{"lname"}[0..n]{"value"}
-    $res = sqlStatement("select *, ic.name as provider_name from insurance_data left join insurance_companies as ic on ic.id = provider where pid=? and type=? order by date", array($pid,$ins_type) );
+
+    $res = sqlStatement(
+    "select *, ic.name as provider_name ".
+     "FROM insurance_data left join insurance_companies as ic on ic.id = provider INNER JOIN addresses ON ic.id = addresses.foreign_id ".    
+     "WHERE pid=? and type=? ".
+     "ORDER BY date", 
+    array($pid,$ins_type) 
+    );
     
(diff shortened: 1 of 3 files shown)


teryhill:
Sorry @aethelwulffe , @Trodrige and @tmccormi  the reports.inc file has a lot of just formatting changes. The only lines that matter is 112-140 and 473-479

teryhill:
There is more than that one. I will fix it as a security fix for the program or maybe give it to one of the GCI folks.

library/report.inc
```diff
@@ -109,23 +109,52 @@ subscriber_employer_state => xl('Subscriber Employer State') . ": ",
...
-			if ($ret[$key]['content'] != $value && $ret[$key]['date'] < $list['date'])
-			{
-				$ret[$key]['title'] = $key;
-				$ret[$key]['content'] = $value;
-				$ret[$key]['date'] = $list['date'];
-			}
-		}
-	}
-	return $ret;
+    $sql = "select * from patient_data where pid='$pid' order by date ASC";
```
aethelwulffe: `$sql = "select * from patient_data where pid='$pid' order by date ASC";`
Change to `$sql = "select * from patient_data where pid = ? order by date ASC;";`
If that doesn't work then...
 `$sql = "select * from patient_data where pid= ".$pid." order by date ASC;";`

aethelwulffe: Ooops.  That assumes a numeric pid...
Make that either the ordinal ? or use '".$pid."' (single, double,dot,var,dot, double, single).

teryhill: That is outside of the scope of this pr. It you look down the file there are a huge number that need to be fixed. I offered to change them in another pr. Tony wants the pr's to be what they are, nothing added or taken away.

aethelwulffe: Otay.

teryhill:
Do y'all want this or should I delete it?

teryhill:
@tmccormi is this wanted?

tmccormi:
Seems fine to me.  I'll merge it in.

PR #811, opened 2017-12-04
Created a confirmation dialog, made changes and organized the code

MigDinny:
This solves: #808 

The code was a mess and I idented it and now it must be fine.

Created a confirm message when clicking on submit to avoid accidental clicks. The message is now red.

The successful message is now an Alert() .

There is an issue on line 57. There must be ONLY one bracket at the beginning ( } ) and the code needs 2 of them to work. If i remove one of them, the tab is going to be Undefined and it just doesn't load. I really don't know why this is happening.

Now:

**Normal state**
image

**Confirmation dialog**
image

There is another 2 dialogs when clicking Ok but this must be enough.

PR #835, opened 2017-12-07
Cleanup (hope so): #811

MigDinny:
this is a cleanup for #811
hope this works


teryhill:
@aethelwulffe when you commit this use squash and merge

image

PR #841, opened 2017-12-08
Reorganized the page and added bootstrap

### LINKED ISSUE #821: Patient Finder main wrapper(div) not responsive
Click on Patient->Finder. At least 3 tabs have to be open in order to observe this:
1. The content isn't responsive. At least the holding div's border should contain the child elements
2. "Open in a new window" should have some space from the border-bottom of the wrapper div.
3. Add bootstrap classes to 'non-bootstrap' fields. 

patient-finder

MigDinny:
Fixes #821 

The border was removed because it was the best solution and it was ugly (I hope the creator of this part is not reading this). I wrote a script that added bootstrap classes onto the fields because there was no other way.

jQuery always here to save the world.

### DIFF
--- a/interface/main/finder/dynamic_finder.php
+++ b/interface/main/finder/dynamic_finder.php
@@ -88,7 +88,7 @@ $(document).ready(function() {
   // language strings are included so we can translate them
   "oLanguage": {
    "sSearch"      : "<?php echo xla('Search all columns'); ?>:",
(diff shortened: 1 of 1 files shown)


MigDinny:
Before: (check Issue's description)

After:
image

_This is what is shown in a Full HD screen but on an HD screen like a laptop it is the same, you can test it (I just used my other screen to this screenshot because on my HD screen the tab had a scroll and I didn't want to screenshot twice)_

PR #844, opened 2017-12-09
Creation of Prism Theme

MigDinny:
This theme has new UI features that the old ones didnt have.

Menu was redesigned with some shadows, new colors, a new font-family was added (Roboto) and minor positioning issues were solved like "Patient:"  div on upper-left corner.

The handler that allows us to resize tabs had z-index: 11 and that meant that it would cover menu dropdowns (fixed).

I had some ideas in my mind that I couldn't put on this theme because of bootstrap integration, messy code and iframes.

Menu hover colors (gray) are impossible to change, at least for me. If someone has an idea on how to change them, I would like to hear.

This theme will (probably) have a system that will offer the option to choose a color set to this theme so it always remains the same but the colors change (any ideas on how this could be achieved are welcome):

- **Grass** _(shades of green)_

- **Capuccino** _(shades of orange/brown/yellow)_

- **Sea** _(shades of blue/cyan, the current color set)_

Any minor (I mean minor, not a whole new theme suggestion) improvements, just comment.


aethelwulffe:
I can't review/view this yet, but I am liking the idea.

MigDinny:
image

here's a screenshot

teryhill:
This need's to be reviewed guys. y'all are holding up things. @aethelwulffe @Trodrige @tmccormi

aethelwulffe:
https://www.w3schools.com/cssref/sel_hover.asp

Here are all the hover options I can find.

New Theme: (which looks great by the way)

Miguel, though it isn't written anywhere (should have a project, not an issue) we actually don't want to add any themes.  We want a single GUI driven set of global and user-override theme options.  This PR could easily go into the contribs repo, or we could flat-out replace an existing theme such as "purple".  The reason we don't have a single theme is due to the fact that not only are colors an issue, but some themes work better with some applications than others.
For instance, under QA Reports>measure selector, the checkboxes that look like toggle buttons become almost unreadable with any theme other than purple, thus I have fought making "TAN" the only canon theme.
   So, we would like a single theme class.  Prism could be it.  Here is what we would need:
1.  Basic settings: Font type, size, and color configurable.  Basic palette selections.
2.  Advanced settings:  Palette, link and font configurations for each CSS property defined.

The methods for doing this are actually fairly recently (to us old people).  DCSS, or dynamic css was only previously possible by generating a stylesheet file and placing that in a directory.  That is fine enough, but that sort of thing does not go well with a model that says "if you copy the database, everything should look the same with the new installation."
  So, all default and user settings go into the database, and are loaded as globals/session stuff.
The methods for doing this sort of thing:  https://css-tricks.com/making-custom-properties-css-variables-dynamic/

Now all of this is a bit outside of the GCI task list.  We know that is not all you are about.  We have also seen that you are getting frustrated with the code indents and generally messy stuff.  As some of this code is older than you are, and originated when php, html, js, mysql and even the whole web environment were very very different, yes, things are messy.  If you had seen the code two years ago when we forked off of OpenEMR, you would understand how fantastically better it is now than when LibreHealthEHR first launched.
  
  This project has never really had a "shut it down, refactor everything, let the users deal with upgrading" period.  That is because we have never been able to pay anyone to do a major refactor.  Across-the-board code beautification is also simple to do, but we have found that trying to do it all at once introduced so many bits of broken stuff (due to ancient functions, mixed languages and the like), it was impossible to manage.  That is what white-space beautifiers typically wind up doing when you have code that is not pure js, pure php etc...  Too much old stuff.  Sometimes, putting in a line break where you *think* it should be fine can break the code, especially with text or sql strings.
-So, though we hate it, we do the "stranglevine" refactor.  When you touch a file, you try to clean it up.  There are problems even with that though.  Some people are really twitchy about changing more than one thing per pull request.  That means cleaning up other code that touches other features than what your "issue" describes means you get yelled at.  I ain't one of those, but it happens.  Some people don't understand that it is too hard to work in a messy shop.  They don't see the problem with clutter.  I can't stand it myself.

  An easy fix for messy code is to not worry about what is in the repo.  When you view it in an IDE, just have it _display_ it all pretty.  Another way is to first put out a PR *just fixing the white space*, then do the real work.  This also stinks, but the PR prudes may accept that.
  
  The final issue with white-space reorganizing is that NOT EVERYONE WILL AGREE with what "clean code" looks like.  I find over-indentation very hard to read.  We don't have neat simple code, and strings get very long...comment tags are every flavor that has ever existed, spelling errors abound in those, variable names are just stupid stupid stupid because the programmer was never able to pass a technical writing class...you name it.
-And so....
We are tolerant (if not accepting) of the mess, and we clean as we go.  We do so while attempting not to upset any of the team members: all of whom, all of us, are high-minded highly opinionated and convinced overachievers.  We can't communicate with baseball bats.  We gotta collaborate to whip this nasty beast.

MigDinny:
@aethelwulffe I agree with you. And when I touch any file, generally I ident it and organize it. I don't use any IDE by the way, I use only Sublime Text 3.

About the GCI task itself, what do I have to do in order to complete it? Those changes in the theme you've talked about (basic settings and advanced) are for me to do?

tmccormi:
I'm the picky one about mixing issues, and not issue related work.   The easy thing to do is to fix the syntax issues and commit that, then fix do the issue work and commit that.    Same Pull Request, is fine.  That way a reviewer can diff (and test) the commits separately if there are issues.

MigDinny:
@tmccormi Yes. I'll do that in the upcoming PR's.

interface/themes/rtl_style_prism.css
```diff
@@ -0,0 +1,2 @@
+@import url("style_prism.css");
+@import url("rtl.css");
```
aethelwulffe: New line at end of file please...

interface/themes/style_prism.css
```diff
@@ -0,0 +1,1507 @@
...
+ * @link http://librehealth.io
+ * 
+ * This style is a replacement and its code (not its design) is based on Tan Theme
+ * This Theme must have some color sets associated (Green, Blue and Orange) to be chosen
+ */
+
+/* Imports area */
+
+/* How to use: font-family: 'Roboto', sans-serif; */
+@import url('https://fonts.googleapis.com/css?family=Roboto');
```
aethelwulffe: This is a medical application (outside calls not good), and may be air-gapped.  You need to add the resource to /assets/fonts.

interface/themes/rtl_style_prism.css
MigDinny: Done

interface/themes/style_prism.css
MigDinny: Had to add the .ttf on theme_assets because there was no other way to do it. Tried assets/fonts but on CSS I can't access it because php is not available there to call the globals.

MigDinny:
Check the new commit guys

interface/themes/style_prism.css
```diff
@@ -11,9 +11,12 @@
 /* Imports area */
 
 /* How to use: font-family: 'Roboto', sans-serif; */
-@import url('https://fonts.googleapis.com/css?family=Roboto');
+@font-face {
+    font-family: "Roboto";
+    src: url("theme_assets/Roboto-Regular.ttf");
```
teryhill: Will that install on a linix computer?

MigDinny: @teryhill i dont know but i cant see why not

aethelwulffe: Unicode ttf fonts should be just fine.

tmccormi: You can use things that require access to the internet for this.   Local assets only.

tmccormi: I do not agree with this.  we can't have imports that reference external sources, live.  This will break systems that are not installed with external internet access for security reasons.   @MigDinny @aethelwulffe @teryhill

MigDinny: @tmccormi You are right and I think that was fixed already... not sure

PR #855, opened 2017-12-13
Update Patch script for rel101 to rel 200, and deploy to demos sites

teryhill:
Files that need to be added to the demo site.


tmccormi:
This seems to have three parts. 
1) The creation of an upgrade script for the master branch
2) A request to patch the rel-200 branch to include this script
3) Update the demo installations with the script and (I presume) run it against the installed demo databases.

@teryhill Is that correct?

teryhill:
@tmccormi The demo sites will need it. Release 200 needs it , I thought this was the master branch and it is just an update to the existing upgrade script.

tmccormi:
OK. I changed the description to be more meaningful.  Will patch and then have Robby update the sites.

tmccormi:
@teryhill  this pull request is against the master branch, It looks like it needs be merged there as well.

PR #857, opened 2017-12-13
Print insurance name on statement

teryhill:
Print insurance company name on statement instead of "Insurance information on file". I am working on the not printing the statements based on insurance company and thought this would be a good addition.


teryhill:
@tmccormi and @aethelwulffe Bump

PR #868, opened 2017-12-17
The Add Patient > Insurance form layout was fixed

### LINKED ISSUE #853: Add new patient Insurance form needs some structure
I just watched this: [video](https://www.youtube.com/watch?v=G9auafGlwQs&feature=push-u&attr_tag=j6VRqPJKYqnl-xqg-6)
I was filled with shame when it reached the 4th minute. Just take a look at the insurance form. The difference is clear with the other categories of forms while adding a new patient. Something has to be done.

MigDinny:
Fixes #853 

The code was idented and organized. I forgot to commit this separately first...

I fixed the layout, created some new tables and stuff like that. The left side wasn't required on the issue but I fixed it too.

Before: 
image

After:
image


aethelwulffe:
Miguel, this is clearly better.
It isn't quite "Beautiful" yet though.  Do you see any way of getting the label>input field spacing to stay consistent at different tab widths?  Should we switch to a resizable modal window independent of a tab iframe for most all inputs that affect session-level variables (like patient, encounter forms etc...)?

MigDinny:
@aethelwulffe What you mean by "consistent at different tab widths"? That space appears to be responsive... Can you give more details?

Didn't understand the second question too.

tmccormi:
mine does not look anything like your example.  I'm using the Tan style.  is this theme specific?

insurancedata

tmccormi:
I see, you did this is search/add but not in EDIT...  I think they need to be consistent.

MigDinny:
@tmccormi Never thought about Edit mode... I'll fix that.

MigDinny:
@tmccormi That is consistent between tabs in EDIT mode. They are different pages (and codes). I think that  it should be organized but when adding bootstrap to EDIT mode, maybe in another pull request...

MigDinny:
@aethelwulffe I found out why it's not consistent. When you add more tabs, one of the inputs of "Subscriber" gets breaklined and that's messing the layout. I'll think about a solution and I'll try to fix that.

aethelwulffe:
@MigDinny Cool.  Standing by with this.
I am checking this out again sort of as it seems like this and the Fancybox thingie might deserve simultaneous consideration in the light of standardization.

No matter what the resolution, we don't want the elements spreading too far apart in columns.  Makes it harder to do input...though I know you understand that, and neither bootstrap nor the layout engine do a very good job of that.

aethelwulffe:
Edit mode in another PR is fine...but I would like the second PR up at the same time before merging.

MigDinny:
@aethelwulffe So, excluding the second PR for the edit mode, what should I do now?

aethelwulffe:
Dunno.
The new version uses a lot more space.  Part of that is unequal column numbers, other is everything in two columns when three is more compact.  In vertical tabs, I would like to test and see how often scrolling is required.  Sometimes, when things are hidden, data simply does not get entered.
On the other hand, the original looks messy too, which causes similar issues.

As to content, SS (Social Security Number) is not actually required anywhere, and is only a regional thing (United States).  Also, the SE section is omitted with a Globals setting, and the layout would need to adjust to that.  That means testing with that global flag off.

MigDinny:
@aethelwulffe I don't know what to do, also... I think that, for now, this version is ok.

MigDinny:
@aethelwulffe what do you think?

aethelwulffe:
I will be testing and reviewing this afternoon (Eastern Standard Time, USA).  After a single approval, you can claim this issue.  If there is anything you think could be better in the meantime, please do.

I will be testing it with the Exclude Employer global flag off, as well as with making other common tweaks.  I will then post the screenshots here.

MigDinny:
@aethelwulffe understood. I'm going to try to solve the "subscriber breaklining" issue meanwhile.

MigDinny:
@aethelwulffe I've made some changes. I've set a fixed width for the divs on Add Patient so it is consistent regardless of the number of tabs opened. That means that nothing gets breaklined as it was happening.

I have put the last row of the right side on the left side so it becomes more consistent, also.

I've realized that different input types make different TR heights. That means that a dropdown can make a TR have 30px and a normal text input can make a TR have 40px. That's what making the rows messy. I made a research on this and we could make a huge "refactor" on the whole program (i'm not sure if it's a good idea but it's an option to keep in mind) which consists of: wrapping all contents of all TD's in divs and those divs could have a fixed height and width so the tables could become consistent. This is probably not a good idea, but can lead to better ones.

image

Look at "relationship" dropdown. It's a bit smaller and that creates that mess. I've searched about way of making TR having a style like:

`table td {
   height: 30px;
}`

But I think that is not possible. Hence the "div" stuff. Wrapping all contents with a div can make this style a possible thing.

Once again: this is probably not a good idea because it would take a huge refactor and can create a lot of bugs just because of this "height issue" i'm talking about.

Anyway, I've made some changes on the layout and added a placeholder on **Subscriber Employer** instead of a span below.

This is what happens when we open several tabs:
image
_(since the div width is 1200px and can't be changed, a scrollbar is created and nothing gets breaklined)_

@aethelwulffe give me some feedback :)

aethelwulffe:
ARgh!  No feedback at the moment.  This is getting quite complex.

Sorry buddy, I gotta look at this in the (my) morning when my brains are on fresh coffee.

MigDinny:
@aethelwulffe no problem! just say something when possible ;)

teryhill:
@MigDinny One suggestion, The "Relationship" field is a field that will fill out most of the subscriber information. Can you move it to the top line where the subscriber name is now?

MigDinny:
@teryhill just exchange relationship with subscriber fields, right?

Yes i can do it

teryhill:
file_change

@MigDinny

MigDinny:
@teryhill done, check it out

aethelwulffe:
@MigDinny 
Put the relationship at the top of the column.
If they pick "self" then all the info below (including name) gets filled in automatically.
I know it seems like the name should be at the top, but functionally speaking "self" is the most common choice, and thus makes a better UI.

aethelwulffe:
-That is my change request awaiting for approval :)  Good work dude!

MigDinny:
@aethelwulffe I did, isn't this what is asked?

image

If not, can you explain me?

teryhill:
@aethelwulffe it is moved. Just needs a screen shot

aethelwulffe:
Yer all done dude.
We have two approvals...this is merging, closing #853

PR #869, opened 2017-12-19
Changed fancybox library code to fix the layout issue

### LINKED ISSUE #852: Add Patient tab freezes when clicking on Search/Add Insurer
Open Add New/Search on Patient/Client >> Scroll down to Insurance >> Click on the anchor "Search/Add Insurer".

This apparently freezes the tab but the problem is on the fancybox. A **fancy_overlay**, **fancy_loading** and **fancy_outer** are created after clicking on that anchor. It is supposed to appear a popup to search for an insurer but that doesn't happen. Only these elements show themselves up with no apparent usefulness. It's possible to see them on Element Inspector after tab "freezes".

There are 2 problems here then:

- **Search for insurer tab doesn't show up**

- **Tab appears to be frozen (impossible to click anywhere) after clicking on the anchor because the elements (described above) fill the space**

_I can solve this, but I need some help by someone who has already coded this system_

MigDinny:
Fixes #852 

I changed some values of "top" properties directly on the library. I've set the popup "position" to "fixed" because I commented the code that was handling the positions when scrolling. That had to be made because when we scrolled, the page would become infinite. Besides, "position: fixed" exists for some purpose.

_There isn't any other solution except this one. The other one would be to create a whole new popup system but that would create a lot of bugs because the code is mixed with fancybox and a lot of features were programmed with fancybox and they would be lost._

Before: _[check [ehr.librehealth.io](ehr.librehealth.io)]_

Now:
image


teryhill:
making changes to a third party item is not recommended It will cause issues for others that are expecting a known result from the library.

aethelwulffe:
We call fancybox in 26 different places, but one of them is in the headers.inc.php, which means a lot of pieces to test.  If we can put these overrides into a second file.
There is another simple fact, in that Fancybox is now at version 3.0.
This version of Fancybox is now 9 years old.  Basically, some of the developers in our project were six years old when it came out.
-So...maybe there IS a different way of addressing this issue.  It seems that since Fancybox now supports touch, and we might also want to use it for displaying images such as patient photos or licenses without other clunky loading happening, I suggest a Fancybox upgrade to start with.  You game for this @MigDinny ?

MigDinny:
@aethelwulffe I didn't realize the version was old... Yes, I could upgrade fancybox. Do I need to test it in EVERY place it is being called?

aethelwulffe:
No, but someone ELSE has to test it at very least!  :)  
Seriously, I totally do not mind testing every instanciation.

aethelwulffe:
Number one reason to upgrade Fancybox is to _get functionality you want right now_.  If it does not do anything new and useful, or take care of any new problems, then reconsider.  Evaluate it, then give us your opinion.

MigDinny:
@aethelwulffe I will upgrade and see if this issue is solved. If not, our best option would be to create a custom file for the changed library only for this issue. I'll bring news in a few hours.

aethelwulffe:
Excellent.

MigDinny:
@aethelwulffe I've tried to upgrade but there were a lot of compatibility bugs and license problems. I tried to use FancyBox 3 but it didn't run as I was expecting. Fancybox2 didn't even run.

I replaced the original fancybox library for the rest of the program but I created a custom version called **fancybox-addpatient** which is **only** called in Add Patient.

I think this is the best solution for now.

Feedback :)

teryhill:
I think the removal of fancy box completely from the system would be better. They are making their licenses more restrictive and other open source projects are removing it.

aethelwulffe:
Yeah, we need to stick with open-source, not projects like that.
http://brutaldesign.github.io/swipebox/ is as close as I have found.

MigDinny:
@aethelwulffe As far as I've seen, that library appears to be easy to use (being easy to use is the first thing I look when choosing a library) and good for us. I vote for its implementation!

However, I don't think it's a priority for now and I have a lot of work to do... is it possible to get this approved to fix the layout issue (described in #852) until we move to swipebox? @aethelwulffe

aethelwulffe:
@MigDinny 
Please add (to the dev docs, and a note in the Fancybox asset directory) describing what you did to Fancybox.  
@teryhill is it ok to do that, or should he make a _mod version of the library files?  I vote we implement what he has here in some fashion (asset wise), and look forward to fancybox replacement as a new project in the near future.

aethelwulffe:
Oops!  my mistake.  Miguel already did a modified version of files.

MigDinny:
@aethelwulffe Thanks! Where are the dev docs?

aethelwulffe:
Miguel, there seems to be no existing document under the Documentation directory for you to add to.  You could put in a readme in your new asset file location to describe it.  My original documentation request was based on you CHANGING the library, not on creating a custom version of it, which is what you did afterwards.

MigDinny:
@aethelwulffe understood. about the merge conflicts, I was aware of them when I made this PR and are, as far as i know, easy to solve

aethelwulffe:
@MigDinny good, because I think you will be one of the people that will suffer from it 😸

MigDinny:
@aethelwulffe fixed some merge conflicts

please take another look just to make sure this is good to go to the codebase without bugs created from this merge conflict commit

aethelwulffe:
Looks okayish.
I shall retest.  Doing a lot of stuff ATM...

aethelwulffe:
There is a lot of Fancybox discussion going on right now, but this BUG has a fix.  It woud be nice to attend to.

aethelwulffe:
I am bumping this again.

This is our oldest mergable PR.

tmccormi:
Has anyone else tested or reviewed this yet?

tmccormi:
I'm going to merge this...

PR #878, opened 2017-12-20
Added Bootstrap to all compatible elements

### LINKED ISSUE #877: Transform Fees' Posting, Batch payments and EDI History pages
We want to integrate Bootstrap in EHR. Change all the components in the Fees->Posting, Fees->Batch Payments and Fees-> EDI History pages to Bootstrap components.

MigDinny:
Fixes #877 

Bootstrap was added to most of the elements of these pages.

Simply adding bootstrap to the required elements would mess up the whole page. To avoid that, several workarounds were made like the addition of styles and breaklines. Adding bootstrap on files like these is not easy. There are a lot of properties that were deprecated in HTML5 that were still there. I wrote an equivalent style for those I've caught.

Couldn't do my best because the way those tables were coded there wasn't much to do.

**My opinion:** I personally don't like bootstrap because it limits us. Several inputs could have _padding: 0px_ and would appear much more stylish but I'm restricted to bootstrap styles. Anyway, I did it but my advice for the whole project is: don't try to insert bootstrap in a platform that wasn't originally designed with bootstrap. It brings a lot of problems. The best way to redesign the whole thing would be to make our own styles.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+
+sites/default/edi/history/edi_history_log.txt
--- a/interface/billing/edih_view.php
+++ b/interface/billing/edih_view.php
@@ -26,6 +26,7 @@
 $sanitize_all_escapes=true; 
 $fake_register_globals=false; 
 require_once(dirname(__FILE__) . "/../globals.php");
+require_once("$srcdir/headers.inc.php");
 //
 if (!acl_check('acct', 'eob')) die(xlt("Access Not Authorized"));
 //
@@ -40,9 +41,11 @@ $DateLocale = getLocaleCodeForDisplayLanguage($GLOBALS['language_default']);
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 
 <head>
-    <title><?php echo xlt("edi history"); ?></title>
+    <title><?php echo xlt("EDI History"); ?></title>
     <meta http-equiv="content-type" content="text/html;charset=utf-8" />
-    
+    <?php
+        call_required_libraries(['bootstrap']);
+    ?>
     <link rel="stylesheet" href="<?php echo $css_header?>" type="text/css" />
 
     <link rel="stylesheet" href="<?php echo $web_root?>/library/css/jquery-ui-1.8.21.custom.css" type="text/css" />
@@ -104,13 +107,19 @@ $DateLocale = getLocaleCodeForDisplayLanguage($GLOBALS['language_default']);
         <tr>
         <td colspan='4'>
         
+        <style>
+            td {
+                padding: 5px;
+            }
+        </style>
+
         <form id="formcsvtables" name="view_csv" action="edi_history_main.php" target="_blank" method="post">
             <fieldset style='float:left'>
                 <legend><?php echo xlt("View CSV tables"); ?>:</legend>
                 <table cols='4'>
                     <tr>
                         <td colspan='4'>
-                            <?php echo xlt("Select a percentage of the rows or or select dates"); ?>
+                            <?php echo xlt("Select a percentage of the rows or select dates"); ?>
                         </td>
                     </tr>
                     <tr>
@@ -118,10 +127,10 @@ $DateLocale = getLocaleCodeForDisplayLanguage($GLOBALS['language_default']);
                             <?php echo xlt("Select CSV table"); ?>:
                         </td>
                         <td align='center'>
-                            <?php echo xlt("Pct (%) of rows"); ?>
+                            <?php echo xlt("% of rows"); ?>
                         </td>
                         <td align='left'>
-                            <?php echo xlt("Start Date"); ?>: &nbsp;&nbsp;&nbsp;&nbsp; <?php echo xlt("End Date"); ?>:
+                            <?php echo xlt("Start Date"); ?> & <?php echo xlt("End Date"); ?>
                         </td>
                         <td align='center'>
                             <?php echo xlt("Submit"); ?>
@@ -129,12 +138,12 @@ $DateLocale = getLocaleCodeForDisplayLanguage($GLOBALS['language_default']);
                     </tr>
                     <tr height='1.5em'>
                         <td align='center'>                 
-                            <select id='csvselect' name="csvtables"> 
+                            <select id='csvselect' name="csvtables" class="form-control form-rounded"> 
                             </select>               
                         </td>                       
                             
                         <td align='center'>
-                            <select id="csvpct" name="csvpctrows">
+                            <select id="csvpct" name="csvpctrows" class="form-control form-rounded">
                                 <option value="5" selected="selected">5%</option>
                                 <option value="10">10%</option>
                                 <option value="25">25%</option>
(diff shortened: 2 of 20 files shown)


Trodrige:
@teryhill test this. I think @MigDinny is right. The pages are better without bootstrap (to me). More eyes, and thoughts.

MigDinny:
@trodrige i can do this kind of bootstrap tasks, would you Open more issues? :)

Trodrige:
Can you make the views presentable. The form elements should be arranged,
not just scattered around the page.

MigDinny:
Well that is the reason inserting bootstrap isnt a good idea. I tried to but i couldnt make it work. I can try again but im not sure it will be presentable .

MigDinny:
@Trodrige 
image

Is it okay if I do this way? All vertical. 49513425245246^2 tables were removed.

(it's not finished, I'll put some paddings and stuff too)

yehster:
Arranging all the elements vertically does not make efficient use of the available space.  
That doesn't leave much room at the bottom for the results, which means if there are more than a couple, the user will have to scroll around a lot.

MigDinny:
@yehster yes, you are right. This isn't the best way.

Any suggestion?

MigDinny:
@yehster Well, it's impossible to have those 3 things:

- Space for the form and results;

- Bootstrap on all of those fields;

- Presentable

We can only choose 2 from those 3. Bootstrap makes the inputs large and that **necessarily** requires space.

There are 3 options:

- Vertical form and results below

- Shrink everything (and turning it into a mess) but having space for results

- Removing bootstrap (but it cannot be inconsistent regarding to the rest of the program; if a field doesn't have bootstrap, the rest of the program must **not** have bootstrap

teryhill:
@MigDinny the vertical screen will not work. The users will complain about the scrolling. Wasted real-estate   on the screen. To understand the impact of the changes you need to use the system to see how it functions.

MigDinny:
@teryhill I did it in a rush, I just had time to screenshot and ask for feedback.

Can you answer my post above? I need an alternative and bootstrap limits me.

aethelwulffe:
Extending static CSS on top of bootstrap is certainly non-optimal.
I find little value in frameworks for exactly this reason...there is no one-size fits all, and while frameworks are great for quickly producing one-offs, it is stupid and sucky to build something specialized and high-value using the "Costco" brand materials.

My suggestion:  https://hacks.mozilla.org/2017/04/replace-bootstrap-layouts-with-css-grid/

aethelwulffe:
Personally I have not been really happy with most of the bootstrap conversions so far.  Some people say it looks "Modern".  I say "Junky toy".
Bootstrap 2x had a lot of interest in compacting this stuff.  It looks like it got ignored despite about 50 thumbs-up votes.  https://github.com/twbs/bootstrap/issues/3737
In whichever case, we have the code we could use to modify the styles.

Bootstrap, as a project, is intended to allow you to download and include only the components you want.  As such, they provide https://getbootstrap.com/docs/3.3/customize/
What did we do?

If the vote is to revert the bootstrap implementation in these particular forms in favor of a nice bit of code that we can use in a number of places for consistency, I am TOTALLY ON BOARD.
  People may want to use Bootstrap, but as a project, we are not Twitter.

MigDinny:
@aethelwulffe What do you mean by "Twitter"? Will bootstrap implementation go on or nah?

aethelwulffe:
Bootstrap was made for Twitter.
We, on the other hand, typically deal with more than 140 characters.  Even more than 280.

As for Bootstrap "going on", I am not the King of Siam.  None of us can say "Let it be written, let it be done!"

Here are the issues:
1.  People (young devs) joining the project hated the diaspora (means "too many") ways things were done.  Their answer was "Make everything Bootstrap".  We like new people, and need them, so it was decided to give them what they wanted.
2.  Bootstrap was already in the project, but added all over the place ad-hoc.  We centralized the asset and started using it.  It was implemented with the intent that it should be used everywhere.
3.  Reality raised its head, and the predictions of the old folks here came true in that Bootstrap is not a one-size-fits-all solution for every type of application.
4.  We still want responsive (meaning scalable layout) utility.  This can naturally be achieved without using the Bootstrap library, but we also want standardization and to limit the different ways things are done.

So:  No, it appears the implementation of Bootstrap everywhere simply for its own sake is foolish.
  -Keeping in mind the re-usability of code, do this, and everything else you think needs improvement in the way _you feel it is best_.  Don't waste time trying to make other people's stupid ideas work 😸 .  You bring more value by saying "I protest this bootstrap stuff" than by following your idiot leaders.  Don't ever assume that someone writing a GCI task (or any other issue, feature request et cetera) has considered every aspect of the problem.

MigDinny:
@aethelwulffe Just to state my opinion, I find bootstrap beautiful but I'm against its implementation. Half of the program with bootstrap and half without bootstrap isn't consistent. Besides, bootstrap brings a lot of problems (mainly layout problems) in a project of this dimension. I mean that inserting bootstrap in EHR will bring more cons than pros, in my opinion.

aethelwulffe:
@MigDinny 
As far as GCI is concerned, if you think you will be finishing up with less than Maxima Cum Laude, you have another think coming.
  I will take measures to make sure you get credit for doing this, as well as seeing about getting a task "Participate in Project Management" claimed as well.  "Do no harm" is better than "Did it anyway".
@Trodrige @teryhill @yehster @tmccormi @robbyoconnor  How do you folks feel about this?  Telling us that a task is dumb scores pretty high in my book.

MigDinny:
@aethelwulffe understood. waiting for new instructions

tmccormi:
I agree, sometimes the best answer is "That's really not a good idea, now that I've done the research"...   In fact I frequently have to tell my clients that, and yet, I get paid for the advice :-)

aethelwulffe:
@tmccormi Glad you agree, but typically I shoot holes in the idea right from the outset, which doesn't get me paid at all.   Of course, even finishing the work usually doesn't get me paid either.

MigDinny:
@aethelwulffe I've made some changes. I've set a fixed width for the divs on Add Patient so it is consistent regardless of the number of tabs opened. That means that nothing gets breaklined as it was happening.

I have put the last row of the right side on the left side so it becomes more consistent, also.

I've realized that different input types make different TR heights. That means that a dropdown can make a TR have 30px and a normal text input can make a TR have 40px. That's what making the rows messy. I made a research on this and we could make a huge "refactor" on the whole program (i'm not sure if it's a good idea but it's an option to keep in mind) which consists of: wrapping all contents of all TD's in divs and those divs could have a fixed height and width so the tables could become consistent. This is probably not a good idea, but can lead to better ones.

image

Look at "relationship" dropdown. It's a bit smaller and that creates that mess. I've searched about way of making TR having a style like:

`table td {
   height: 30px;
}`

But I think that is not possible. Hence the "div" stuff. Wrapping all contents with a div can make this style a possible thing.

Once again: this is probably not a good idea because it would take a huge refactor and can create a lot of bugs just because of this "height issue" i'm talking about.

Anyway, I've made some changes on the layout and added a placeholder on **Subscriber Employer** instead of a span below.

This is what happens when we open several tabs:
image
_(since the div width is 1200px and can't be changed, a scrollbar is created and nothing gets breaklined)_

@aethelwulffe give me some feedback :)

MigDinny:
@aethelwulffe what can I do in order to complete this PR? Is it ready to be approved in GCI ? Do i need to do something more? The form probably needs an adjustment. What is the best way to do that? I think that inserting more frameworks and stuff like that will worsen.  What if I build a table with only 2 columns and with fixed input widths? It will shrink the total height of the table giving enough space for the results. What do you think? Is it a good idea?

aethelwulffe:
If you actually want to make an improvement on this form, go ahead by any method, but this PR is APPROVED _as a task signoff_.  Please link a task claim to this so we may sign off on it.

MigDinny:
@aethelwulffe Appreciate that! I'll into it when I have time. I'll claim this PR in GCI when possible

PR #889, opened 2017-12-26
Organized & Idented code and bootstrap addition

### LINKED ISSUE #887: Bootstrap on User Preferences & Change Pass Phrase
There are some inputs lacking bootstrap.

image
image

Can I work on this? @teryhill @Trodrige

MigDinny:
Fixes #887 

1 detail: the search button now has a float:right because bootstrap was creating a breakline between the input and submit button for some unexpected and unknown reason and it was ugly if the button stayed on the left side.

### DIFF
--- a/interface/usergroup/user_info.php
+++ b/interface/usergroup/user_info.php
@@ -8,9 +8,8 @@
   <head>
     <link rel="stylesheet" href="<?php echo $css_header;?>" type="text/css">
     <script src="checkpwd_validation.js" type="text/javascript"></script>
-    <script src="<?php echo $webroot;?>/library/js/jquery-1.9.1.min.js" type="text/javascript"></script>
     <?php
-      call_required_libraries(['bootstrap']);
+      call_required_libraries(['bootstrap', 'jquery-min-1-9-1']);
       ?>
     <script language='JavaScript'>
       //Validating password and display message if password field is empty - starts
--- a/interface/usergroup/user_info.php
+++ b/interface/usergroup/user_info.php
@@ -85,3 +85,4 @@
 <?php
   //  da39a3ee5e6b4b0d3255bfef95601890afd80709 == blank
   ?>
+  
--- a/interface/usergroup/user_info.php
+++ b/interface/usergroup/user_info.php
@@ -1,96 +1,87 @@
 <?php
-include_once("../globals.php");
-include_once("$srcdir/sql.inc");
-include_once("$srcdir/auth.inc");
-?>
+  include_once("../globals.php");
+  include_once("$srcdir/sql.inc");
+  include_once("$srcdir/auth.inc");
+  include_once("$srcdir/headers.inc.php");
+  ?>
 <html>
-<head>
-<link rel="stylesheet" href="<?php echo $css_header;?>" type="text/css">
-<script src="checkpwd_validation.js" type="text/javascript"></script>
-<script src="<?php echo $webroot;?>/library/js/jquery-1.9.1.min.js" type="text/javascript"></script>
-
-<script language='JavaScript'>
-//Validating password and display message if password field is empty - starts
-var webroot='<?php echo $webroot?>';
-function update_password()
-{
-    top.restoreSession();
-    // Not Empty
-    // Strong if required
-    // Matches
-
-    $.post("user_info_ajax.php",
-        {
-            curPass:    $("input[name='curPass']").val(),
-            newPass:    $("input[name='newPass']").val(),
-            newPass2:   $("input[name='newPass2']").val(),
-        },
-        function(data)
-        {
-            $("input[type='password']").val("");
-            $("#display_msg").html(data);
-        }
-
-    );
-    return false;
-}
-
-</script>
-</head>
-<body class="body_top">
-
-<span class="title"><?php echo xlt('Pass Phrase Change'); ?></span>
-<br><br>
-
-<?php
-
-$ip=$_SERVER['REMOTE_ADDR'];
-$res = sqlStatement("select fname,lname,username from users where id=?",array($_SESSION["authId"])); 
-$row = sqlFetchArray($res);
-      $iter=$row;
-?>
-<div id="display_msg">
-</div>
-<br>
-<FORM NAME="user_form" METHOD="POST" ACTION="user_info.php"
- onsubmit="top.restoreSession()">
-<input type=hidden name=secure_pwd value="<?php echo $GLOBALS['secure_password']; ?>">
-<TABLE>
-<TR>
-<TD><span class=text><?php xl('Full Name','e'); ?>: </span></TD>
-<TD><span class=text><?php echo htmlspecialchars($iter["fname"] . " " . $iter["lname"], ENT_NOQUOTES); ?></span></td>
-</TR>
-
-<TR>
-<TD><span class=text><?php xl('Username','e'); ?>: </span></TD>
-<TD><span class=text><?php echo $iter["username"]; ?></span></td>
-</TR>
-
-<TR>
-<TD><span class=text><?php xl('Current Pass Phrase','e'); ?>: </span></TD>
-<TD><input type=password name=curPass size=20 value="" autocomplete='off'></td>
-</TR>
-
-<TR>
-<TD><span class=text><?php xl('New Pass Phrase','e'); ?>: </span></TD>
-<TD><input type=password name=newPass size=20 value="" autocomplete='off'></td>
-</TR>
-<TR>
-<TD><span class=text><?php xl('Repeat New Pass Phrase','e'); ?>: </span></TD>
-<TD><input type=password name=newPass2 size=20 value="" autocomplete='off'></td>
-</TR>
-
-</TABLE>
-<br>&nbsp;&nbsp;&nbsp;
-<INPUT TYPE="Submit" VALUE=<?php echo xla('Save Changes'); ?> onClick="return update_password()">
-
-
-</FORM>
-
-<br><br>
-</BODY>
+  <head>
+    <link rel="stylesheet" href="<?php echo $css_header;?>" type="text/css">
+    <script src="checkpwd_validation.js" type="text/javascript"></script>
+    <script src="<?php echo $webroot;?>/library/js/jquery-1.9.1.min.js" type="text/javascript"></script>
+    <?php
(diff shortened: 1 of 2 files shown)


MigDinny:
Asking for review: @teryhill @Trodrige

interface/usergroup/user_info.php
```diff
@@ -1,96 +1,87 @@
...
+    </FORM>
+    <br><br>
+  </BODY>
 </HTML>
-
 <?php
-//  da39a3ee5e6b4b0d3255bfef95601890afd80709 == blank
-?>
+  //  da39a3ee5e6b4b0d3255bfef95601890afd80709 == blank
+  ?>
```
aethelwulffe: Add newline at end of code.  This is a GIT line-ending normalization thing.

MigDinny: I always forget, will do it ;)

MigDinny: Done!!

aethelwulffe:
@MigDinny please pop in a screenshot (before/after), and work on taking your [Before] shots before you start work (lots easier that way).  With a lot more contributors here now, that will make things a LOT easier for other folks to evaluate.
-Nice work.

MigDinny:
@aethelwulffe 
Before => (issue)
After =>

image

interface/usergroup/user_info.php
```diff
@@ -1,96 +1,88 @@
...
-
-
-</FORM>
-
-<br><br>
-</BODY>
+  <head>
+    <link rel="stylesheet" href="<?php echo $css_header;?>" type="text/css">
+    <script src="checkpwd_validation.js" type="text/javascript"></script>
+    <script src="<?php echo $webroot;?>/library/js/jquery-1.9.1.min.js" type="text/javascript"></script>
```
teryhill: Add the call for this to the "call_require_libraries"

MigDinny: done

MigDinny:
@teryhill it's good now

aethelwulffe:
OK, now the branch runs.

teryhill:
I think the save button is more convenient at the bottom of the screen since this impacts globals .

aethelwulffe:
So a save at top and bottom, or just at the bottom to ensure everything has been looked at/viewed?
It does look better at the top (clean) but if Terry feels like there would be user issue with it, then he is probably right.
@MigDinny can you make an easy tweak there, or should we bring this to committee? 😸

MigDinny:
@aethelwulffe I'm confused... the button at the top makes it cleaner but if we put it at the bottom would be better to ensure everything was viewed. But if the bottom goes to the bottom, scrolling may be needed (not sure). 

But anyway, I prefer the button at the bottom. I can do that, may I? @aethelwulffe

aethelwulffe:
I like it at the top too @MigDinny , but Grandpa usually doesn't have an opinion for no reason, so I would like to hear his final word on this.  It like it at the top, and frankly, I wish that everything in the system had the same save/submit/new/add/create icon scheme that appears in the same place.  Sometimes though, we put save buttons at the bottom, because we don't have sanity checks that evaluate minimum data entry or the like.  There is a phone conference in a minute, and we will discuss this bit and come up with a final answer.

I actually wish that with globals, we had a submission/review window that says "confirm the following changes to Global settings:" then made you do a final opt-in on the actual changes made.

MigDinny:
@aethelwulffe tell me something after the final answer

aethelwulffe:
Heh...
First, if you want to reference something in your Task Claims for this, it is approved, so go ahead, but in the conversation we figured out what the answer really should be:

The save should float.

Tony is doing a mock-up of what we are looking for.  This is something that we would like to eventually become the standard for everything in the EHR.  We discussed putting things in the tab label etc... but here was the final compromise:

First:  You notice that the tab label says "Edit User Settings" or "Edit Global Settings", then right under that, it says the same darn thing.  That looks silly and eats up space.  The way that is supposed to be is that the title bit should be <hidden> to keep it from printing twice.
Second, while the title should be hidden, the actual SAVE, Cancel, Add, Search and similar buttons should always be in the title div.  Everything else under it should scroll, but not the header/title div.

If we can get this worked out here, then we could start doing it everywhere.  This enables a bunch more issues/tasks that could be done.

You are right about the "everything viewed" concept, but that isn't necessarily a behavior we need to force the user into.  After all, they are not required to open up every tab either. 

 There was some discussion of making the tabs float as well.  Jury is still out on that.

MigDinny:
@aethelwulffe understood. for now, the save button should be there or at the bottom?

aethelwulffe:
Where you have it is totally fine.  Looks better too.  Claim this fix for GCI please.
You are on to something here.  The team is using your input to affect a larger change in the project, so they may or may not give us the second approval and merge.  I see no reason not to go ahead and merge it then do other things later, but that is a team decision...and everyone is all excited about the possibilities.  We must give them a chance to come up with a plan.  Tony is working on the project plan and a set of images.  We would like form controls to be the same for just about every form in the application.

  I think that the controls in the Calendar are what they would like to imitate here and everywhere else.  The Calendar is a separate module, but provides a good example.

MigDinny:
@aethelwulffe got it.

about the gci task i've commented on the 7-issue task but it isn't ready yet because it's a 7-issue task

aethelwulffe:
Let me know if you need a time extension on the 7 bug fix task.

MigDinny:
@aethelwulffe I will probably need, can you extend it?

aethelwulffe:
Extended.

MigDinny:
@aethelwulffe thank you!

tmccormi:
Can this be merged are we still waiting to hear from @teryhill ?

aethelwulffe:
Two-review is the requirement.  We are OK to merge unless we have an actual negative review.
This is another project where we have discovered we want something totally different.  We await the Tony McCormick full_calendar inspired design layout project plan.  In the meantime, I say merge this thang.

teryhill:
How was the Save Button Handled? Was it made to float or was it moved back to the bottom or was it left at the top? @migdinny

aethelwulffe:
Floating is awaiting a new project based on the Tony design.
-Unless @MigDinny you think you can make it float...in which case you would win the "totally cool dude" award for the week.

teryhill:
Then move it back to the bottom. as most users will work from the top down in the screen.

MigDinny:
@aethelwulffe @teryhill I could do it but you gotta tell me what "float" means in this case

aethelwulffe:
https://www.quackit.com/css/codes/css_floating_menu.cfm

aethelwulffe:
Make save button (and maybe also tabs) stay on the screen even when you scroll

teryhill:
@migdinny float means that when the user scrolls down the screen the buttons stay in the viewing window so no matter where the user is in the scroll they can still access the save button.

aethelwulffe:
Just say if you want to do the change or not.  This is otherwise approved.  Floating thingies would be kinda tasty though!
(Idiom translation of "tasty" in this context is "gostoso" not "saboroso"), and "thingie" is "coisinha".

teryhill:
He is looking at making it float @aethelwulffe

MigDinny:
@aethelwulffe @teryhill I'll do it and I have kind of an alternative in mind that you will probably like, just don't wanna tell you for now, it's surprise hehe

aethelwulffe:
I hear a faint "Muahahahahahaha" in the background somewhere....  I am scared.

Just remember...KISS - Keep It Simple Stupid.

MigDinny:
@aethelwulffe (I'm hearing it too haha)

Well, I'm laughing so much because this is so ************** simple 

I'll try to do it now, brb

aethelwulffe:
@tmccormi 
Do not merge this right now!
Another commit coming, and another round of testing is required!

MigDinny:
@aethelwulffe 

image

Now the whole page is floating :)

aethelwulffe:
1.  make "Edit User Settings" hidden.  It only needs to say that in the tab.
2.  Squash save/search and the tabs into as little vertical space as possible.

THAT would be totally awesome.

MigDinny:
@aethelwulffe 
image

Totally awesome now? haha

aethelwulffe:
Yeah, it's too bad we can't integrate the controls in next to the tabs easily (and cleanly) without a lot more effort, but this IS TOTALLY AWESOME and should be done in a whole lot of places.  If padding were reduced, it would be even better, and we can work on that some in the future.

@teryhill @Trodrige @tmccormi  You guys like?
I am going to test this now (will report back), but the code looks good to go.

teryhill:
@migdinny look at this under administration globals.

MigDinny:
@teryhill what exactly?

teryhill:
image

Screen shot

aethelwulffe:
Muahahahahahahahaha!

aethelwulffe:
Might have to leave the tabs out of the globals unless they get compacted waaaaay down.

teryhill:
still a little more refinement Mongo

aethelwulffe:
We are getting closer!!!!

MigDinny:
@aethelwulffe @teryhill I'll be making div height in a % to fit all screen resolutions, tomorrow

aethelwulffe:
Night Miguel!

MigDinny:
@aethelwulffe @teryhill done, check it out

i made it 75%. i use two screens, one in HD and another in FHD and it looks good in both. When i resize the tab it keeps its ratio

teryhill:
That looks good for 1 or 2 tabs open. Open 4 and see what it looks like @MigDinny

MigDinny:
@teryhill I've studied this and the possible solutions. The only solution that came up into my mind was to increase width value of the div so the tabs wouldn't get breaklined and mess the settings

This is what's happening:

image
image

Look at Edit global settings

Those almost 30 tabs are spending too much space. The solution would be to extend width values so it would create a horizontal scroll like Patient finder for example. But I had some problems like: couldn't make it with % because that would be useless; couldn't make it with a fixed PX value because that would make this ugly when maximized

Having 4 tabs opened requires a little bit more scrolling... but it's not an important problem for now, maybe it can be solved with more thinking... Anyway, the user always have the solution to lock the other tabs. This current tab design must expect some space problems like this particular one. It's hard to keep a lot of information and components in this and make them responsive to screen's resolution and number of tabs opened, etc...

Issue #845, opened 2017-12-09
Add Ability to not print statements by insurance company


teryhill:
Working on this to give the ability to mark an insurance company so any patient of that insurance company will not get a statement. EX Medicare

![image](https://user-images.githubusercontent.com/4655366/33792272-15a573de-dc69-11e7-8072-324e9261a7f2.png)

teryhill:
I have this PR ready to send. The Globals entry needs some clarification. Options 3 and 4. None is to not use this feature all is if the insurance company is in any position do not print a statement. @tmccormi @aethelwulffe

teryhill:
PR Merged Closing Issue

PR #901, opened 2017-12-29
Do not print statements by insurance company

teryhill:
Do not print statements by insurance company #845


library/globals.inc.php
```diff
@@ -1144,6 +1144,19 @@
       xl('Total Minimum Dollar Amount of Statement to Allow Printing.(only applicable if Allow Statement to be Excluded from Printing is enabled)')
     ),
 
+    'insurance_statement_exclude' => array(
+       xl('Do Not Print Statements For Insurance Companies'),
+       array(
+            '0' => xl('Primary'),
+            '1' => xl('Secondary'),
+            '2' => xl('Tertiary'),
```
aethelwulffe: Multi-select?  Not looking at the test (doing that tomorrow) but does the logic work out with the default that is only prints out the statement for Primary?

sites/default/statement.inc.php
```diff
@@ -104,6 +104,12 @@ function create_HTML_statement($stmt) {
   #minimum_amount_due_to _print
   if ($stmt['amount'] <= ($GLOBALS['minimum_amount_to_print']) && $GLOBALS['use_statement_print_exclusion']) return "";
   if ($stmt['statement_print'] == "NO" && $GLOBALS['use_statement_print_exclusion']) return "";
+  if ($GLOBALS['use_statement_print_exclusion'] && $GLOBALS['insurance_statement_exclude'] !=4 ){
+   if ($GLOBALS['insurance_statement_exclude'] ==0 && $stmt['insurance_no_statement_print_pri']) return "";
```
aethelwulffe: Could use better {} syntax, and nested statements.  Not sure if the 4 enum values cover the behavior needed.

library/globals.inc.php
teryhill: No Multi select. If the primary is selected then no statement will be printed if that insurance is primary.

sites/default/statement.inc.php
teryhill: May need more for combinations.

aethelwulffe: Terry, does this satisfy the real-world need, or do we need those other combos?
Can the user override these controls at the actual printing UI?

teryhill: The actual UI for printing is pretty dumb . The request was to not print statements for a specific insurance company IE medicare. There is also another request being worked on to make the UI a little more intelligent and offer more printing abilities.

aethelwulffe:
Terry, Tony, Others...
Are you guys OK with merging this now?  It has been sitting here for a while.  We need some reviews and such on this older material.

aethelwulffe:
101 to 2.0 conflict keep popping up.  Editing in the middle of this file seems to cause these issues.

teryhill:
Conflict resolved

aethelwulffe:
I am testing again now.  I am learning that I need a new DB every time there is an upgrade....

teryhill:
@aethelwulffe bump

aethelwulffe:
This is in danger of going stale.  I think we are going to have to go forward with just one review.

PR #911, opened 2018-01-01
Added some setting recommendations.

alpinus4:
I changed bad recommended value of key_buffer_size to value from INSTALL.md (changed from 10M to 1024M).
Fixes https://github.com/LibreHealthIO/LibreEHR/issues/912

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -182,7 +182,15 @@ Make sure that settings in MYSQL /etc/mysql/my.cnf file include:
 ```
 key_buffer_size set to 256M
 table_open_cache set to 256
-innodb_buffer_pool_size set to 70% of available RAM.
+innodb_buffer_pool_size set to 1024M, but can be set to bigger if needed
+innodb_log_file_size set to 256M and uncommented.
+```
+Also add that settings below other "innodb_..." variables:
+```
+innodb_io_capacity = 2000
+innodb_read_io_threads = 64
+innodb_thread_concurrency = 0
+innodb_write_io_threads = 64
 ```
 Make sure you have disabled strict mode in Mysql . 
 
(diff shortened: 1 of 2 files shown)


aethelwulffe:
What about table_open_cache?
These go hand in hand.
-I also swear that we addressed these previously.  I know that I built the XAMPP distro with a similar buffer setting.

alpinus4:
@aethelwulffe I haven't found _table_open_cache_ parameter in both INSTALL.md and setup.php and I don't know anything about it :/ Could you tell me more about it?

aethelwulffe:
That is because we are not giving recommendations and we probably (I 
think) should if we recommend settings for other related parameters.

Searching for "table_open_cache" in the INNODB context should give you a 
ton of resources.

alpinus4:
Thanks @aethelwulffe .
So, should I remove this PR, or learn more about setting up the Mysql server and add recommendations?

aethelwulffe:
No, don't close this PR.  Pretty sure it should go forward.
If you care to look at other elements before we merge, that is fine, but I suppose we need to get this approved.

alpinus4:
@aethelwulffe I added table_open_cache recommendation as 256 and I decreased key_buffer_size to 256M.
I have no idea for change anything other and think that it isn't necessary, but if you have any suggestions....

aethelwulffe:
@alpinus4 No suggestions here.  These seem perfectly reasonable settings to me.
The main issue here is that we are (supposedly) INNODB, which doesn't give a hoot about these settings at all.

alpinus4:
@aethelwulffe How about now?

aethelwulffe:
@alpinus4 I think the new setting recommendations are totally reasonable.  I wish we could do some actual performance testing with various settings.

Trodrige:
Hi, @alpinus4 . We've moved to GitLab. PR owners are moving their work to GitLab. Can you do this buddy?

Thanks,
Rodrige.

teryhill:
Merged in Gitlab

PR #914, opened 2018-01-03
headers.inc.php library calling refactor (demographics.php)

MigDinny:
#915 

My intention is to do this for the rest of the project and this is only the start and an example to show what I want to do. I need your opinion ( @teryhill @yehster @aethelwulffe @tmccormi @Trodrige ).

Basically, the files I've changed so far were: 

> _demographics.php_ (tested and no problems found, only some 404 .png on console but I think that was happening before)

**Changes I've made in** _demographics.php_ **(just to show what I'm doing):**
- made some changes in **headers.inc.php** in order to optimize and increase compatibility

- gritter and common.js were added to **headers.inc.php** and now they are called via **call_required_libraries** instead of hardcoding them

- jQuery and fancybox now are called through **call_required_libraries**, also

_More details for each file refactor in the commits_


library/headers.inc.php
```diff
@@ -42,43 +46,66 @@ function include_css_library($path)
...
-    if (in_array('font-awesome', $library_array)) {?>
+
+    if (in_array('font-awesome', $library_array) || in_array('fontawesome', $library_array)) {   ?>
         <link rel="stylesheet" href="<?php echo $GLOBALS['assets'] ?>/fonts/font-awesome-4-6-3/css/font-awesome.min.css" type="text/css">
     <?php
     }
-    if (in_array("jquery-ui",$library_array)){   ?>
+
+    if (in_array("jquery-ui", $library_array) || in_array("jqueryui", $library_array)) {   ?>
         <link rel="stylesheet" href="<?php echo $GLOBALS['css_path']; ?>jquery-ui-1-12-1/jquery-ui.css" media="screen" />
```
teryhill: Do you think this needs to be made to work like the check for jquery. We at some point would have more that one jquery-ui

library/headers.inc.php
```diff
@@ -42,43 +46,66 @@ function include_css_library($path)
...
-    <?php 
-    } }
+    /* First checking for any library in a directory matching the string "/jquery-min-/".
+     * When one is found, use that value in the URL string, then add "index.js"
+     * In order to decrease the chance of including and library calling problems, some measures were taken:
+     * - The array in the argument is completely converted to lower-case, meaning that there is not any sensitivity between jQuery & jquery or fancyBox & fancybox
+     * - Some libraries can be called with or without "-". For example: "font-awesome" and "fontawesome" => those 2 are the same thing
+     */
+    
+    $library_array = array_map('strtolower', $library_array);
```
aethelwulffe: This forces lowercase names.  That breaks things on Linux if there is a difference.
We could accept that our standard is to do such, but I am not sure if it should be forced.  
@tmccormi 
@teryhill 
@Trodrige 
@yehster

library/headers.inc.php
```diff
@@ -42,43 +46,66 @@ function include_css_library($path)
...
-    if (in_array("datepicker",$library_array)){   ?>
+
+    if (in_array("datepicker", $library_array)) {   ?>
         <link rel="stylesheet" href="<?php echo $GLOBALS['css_path']; ?>jquery-datetimepicker/jquery.datetimepicker.css" media="screen" />
         <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>jquery-datetimepicker/jquery.datetimepicker.full.min.js"></script>
     <?php
     }
-    if (in_array('font-awesome', $library_array)) {?>
+
+    if (in_array('font-awesome', $library_array) || in_array('fontawesome', $library_array)) {   ?>
```
aethelwulffe: Do we have an existing inconsistency here?

library/headers.inc.php
```diff
@@ -42,43 +46,66 @@ function include_css_library($path)
...
         <link rel="stylesheet" href="<?php echo $GLOBALS['assets'] ?>/fonts/font-awesome-4-6-3/css/font-awesome.min.css" type="text/css">
     <?php
     }
-    if (in_array("jquery-ui",$library_array)){   ?>
+
+    if (in_array("jquery-ui", $library_array) || in_array("jqueryui", $library_array)) {   ?>
```
aethelwulffe: Reason we have or might have an inconsistency?

library/headers.inc.php
aethelwulffe: Agree.  The jquery thing was the best stupid idea I have had in a long time.

library/headers.inc.php
MigDinny: @aethelwulffe this actually does NOT force anything :) maybe you didn't understand well the code (maybe I'm wrong) but this only creates a lowercase version of the argument meaning that I am eliminating the case sensitivity. In other words, you can call "jQuery" "JQUERY" "jquery" "jquErY" and it's all the same. The goal here was to prevent possible bugs or doubts when using the function. 

PS: I'm not a linux user (I wish I were) and I don't know if there are known issues with Linux about this.

library/headers.inc.php
```diff
@@ -42,43 +46,66 @@ function include_css_library($path)
...
         
-    if (in_array("bootstrap",$library_array)){   ?>
+    if (in_array("bootstrap", $library_array)) {   ?>
         <link rel="stylesheet" href="<?php echo $GLOBALS['standard_js_path']; ?>bootstrap-3-3-4/dist/css/bootstrap.min.css" type="text/css">
         <script type="text/javascript" src="<?php echo $GLOBALS['standard_js_path']; ?>bootstrap-3-3-4/dist/js/bootstrap.min.js"></script>
     <?php
     }
-    if (in_array("fancybox",$library_array)){   ?>
+
+    if (in_array("fancybox", $library_array)) {   ?>
```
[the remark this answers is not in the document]
MigDinny: Agree!

library/headers.inc.php
MigDinny: Didn't get it. Are you talking about the font-awesome thing? I've explained it in the comments after call_required_libraries

Calling **fontawesome** is the same as calling **font-awesome**

library/headers.inc.php
MigDinny: I wanted to do this to prevent possible future issues when calling libraries (not to fix any inconsistency)

library/headers.inc.php
aethelwulffe: The library array is used to complete a URL string.  That means that you can pass any library you like into it, and if it is jquery, it completes the URL string with the value.
  If the string value does not exactly match the actual file directory and name (case sensitive) it will break.  By default, Windows file systems do not use case sensitive names, though you can configure that.  By default Linux is exactly the opposite unless you use Darwin or another file system tweak.  Removing hyphens will also kill it.

library/headers.inc.php
MigDinny: Maybe, but, why not having the latest version of jQuery? I mean, they won't probably deprecate anything just because they want... Anyway, I think that there should be the option to call only "jquery" and that would get the latest available version of jquery in the local system. I mean, sometimes, we may not need to really specify a version.

library/headers.inc.php
aethelwulffe: So you are fearing someone using a wrong argument in the function, that is right?  OK I can live with that.  In that case, using a capital Font-Awesome or the like will also break it on Linux.  I don't necessarily have a problem with this, but I don't feel it is necessary.

MigDinny: Yes, my fear was that 😋 
When calling a library, sometimes we need to check at headers.inc.php to know exactly what is the library name to put into the argument. But I can remove this if you say so

MigDinny:
@aethelwulffe I can remove it if you say so. Should I do it or should I wait for other opinions?

MigDinny:
@aethelwulffe @teryhill Maybe jQueryUI should be included when jQuery is included. That way when calling for jquery, jqueryUI would come as well. jQueryUI depends on jQuery and that way would prevent from calling jqueryUI and not jquery

(i've said a lot of jqueries here on this sentence but i think you got it :joy: )

aethelwulffe:
@MigDinny - Of course you may wait for more opinions, as we have to have more than one review in any case.  We go with the consensus, you know?
  As to "just include with others" idea, I think we should stick to the "you MUST SPECIFY EVERYTHING" sort of mindset.  Making people specify stuff is harder sometimes.  Not having granular control is asking for trouble though.  We gain easy granular control by making everything explicitly opt-in vs. opt-out.
  This is of particular use when we must use this function call at different points in a script.  We could just include all the libraries, then have a possible order-of-call issue.  The next use case is when you must include one library, then there is a conditional statement that must be evaluated to determine if another include should be made.
  
That avoids compatibility issues as well as reducing bandwidth/performance issues.  a js.min may not be huge, but these things add up.

MigDinny:
@aethelwulffe Yes, I understand and I agree. 

I'll be working on this task for the rest of GCI probably, because this is a very long one. Do you have any suggestion for something I could do as a task? Anyway, I'm worried because I don't think I'll get this PR completed before GCI ends, but I intend to finish it anyway. Will this work be taken into account even if it's not finished?

aethelwulffe:
@MigDinny this is clearly a Function-conjunction task...with extra credit for submitting refactors.

MigDinny:
@aethelwulffe yes

but about the other questions, this PR will be a long one because there are a lot of refactors, is it good to be approved in GCI even though this is not finished?

and do you have any suggestion for a task I could solve (something related to code, PHP and stuff like that) that I am able to do and challenging?

Lately I've been doing work I found for myself and work that I found important to do and not the pre-made tasks on gci

aethelwulffe:
Miguel,
This is a single implementation of the change with demographics.php.  That is good enough (and all we really want) for a completed PR.
 
Further improvements should go to other PR's...big project, small credited deliverables.

The real challenge I present to you is to take this and keep plugging away.  You want something deeper and more challenging?  Then when you fix up the include calls in a file, upgrade the file and feature to use the latest .js lib version.

aethelwulffe:
Meet the changes I asked for, and you will have one approval.  I bet a merge would happen soon after.  This is good work and should NOT sit here and get stale.
I want it.

MigDinny:
@aethelwulffe Then should I create a new PR for each file I do this refactor?

I will do that challenge.

aethelwulffe:
@MigDinny Yes, agree to all the above.  Let's get this one clean, then go to the next.

MigDinny:
@aethelwulffe what is needed for this PR to be approved and merged ?

aethelwulffe:
For my approval, I am asking for the OR conditions to be removed (should just be one, standardized and exacting) and for the str_to_lower function removed.  I think both will cause issues.
  Other than that, you need at least one other review.

MigDinny:
@aethelwulffe i've made a commit a while ago I forgot to push

waiting for one more review

I need this to be merged asap so I can work on more tasks than this one

library/headers.inc.php
aethelwulffe: We have tried that already.
When we have killed every legacy use of JQuery in the system, then we will consider it.  In the meantime, we can always have the plain old JQuery.js file be the latest and greatest.  Incremental changes and granular control are vital when doing "stranglevine" refactors.  That is a method of code refactoring where you gradually replace the old with the new.  It is hard, but you break less things, and you can do it without a huge budget.

aethelwulffe:
Well, I just checked, and the stringtolower function is still there.  This breaks when an element in the array has a capital in it, you remove the capital, and the actual resource directory has a capital in the name.  *we use the function argument to search for a directory*.   This will create a bad URL.

  That breaks stuff *even if the developer puts the resource name in perfectly correct* on a Linux server.  It is an attempt to be robust, but the function for the JQUERY calls is playing a matching game...an exacting matching game.  This is a show-stopper for me.  

Next, the pipes that try to also allow the developers to make a mistake and have things work are also there.  Code does not work when you use the wrong variable.  If the developer uses the wrong name, then it will not function, and they will fix it.  We don't need two string match arguments.  Even if we did, you would need a pretty loose "LIKE" comparison to be ensured of tolerating an error.  The imprecise nature of this is more likely to cause unexpected behavior vs. fixing it.  If the file or directory we are calling has a hyphen or a capital letter, we can't go changing the input and still find the file.  These are not just Boolean function parameters.  They are or can be actual values.

MigDinny:
@aethelwulffe I actually commited those changes. Don't know what's happening.. 

Check this: https://github.com/LibreHealthIO/LibreEHR/pull/914/commits/25f356695a50cc8025d2e904bc21bf1fbd74645d

Check the 4 commits this PR has, it's the last one (removed OR conditionals... )

MigDinny:
About cb1162e, I moved existing assets to the **assets** folder. In headers, only files in **assets** folder are being used. I didn't remove possible unused assets for now because there's a chance they are being called somewhere in the codebase. I will remove them as I see them and I will do it in a separated pull request. To avoid possible bugs, I won't do it in this PR. Furthermore, this was tested and it's working and should not create any problem. 

@aethelwulffe @teryhill

aethelwulffe:
@MigDinny 
OK..
common.js and dialog.js are NATIVE js files, and thus go in the /libraries/js directory.  They are not externally maintained.  The /assets directory readme indicates the usage of that directory.

If you see a good reason to challenge that, please speak up.  Otherwise, I am requesting that the native stuff go back into /libraries

MigDinny:
@aethelwulffe Yep, sorry, that's me trying to understand what a folder is supposed to do before reading README. [sarcastic_mode=ON]

I will get it back to libraries/js directory

MigDinny:
@aethelwulffe commit is reverted. check this out

aethelwulffe:
ARGH!
Dude...Gritter is an external lib...unless this is a custom version of the code, that can go back into /assets.
...or we can ignore it for now if you are fed up (which I would understand).

MigDinny:
@aethelwulffe woops! I'm doing a lot of stuff at the same time, 1001 pull requests and homework and that stuff, I'm not paying full attention, sorry

let's do that in another PR, can i let this commit reverted? or should I revert the reversion? (mindblown)

aethelwulffe:
*poof* my brain gone too.  Try reverting the reversion?  Then revert, but with fewer steps backwards?  👯

MigDinny:
@aethelwulffe i'm definitely blown up

what am i supposed to do? :joy:

(no, i'm not on drugs)

aethelwulffe:
Put gritter into /assets.
Leave the other (two native) libs in /libraries.

We will get through this!

Less hopping.  More merging!

MigDinny:
@aethelwulffe i think it's okay now

aethelwulffe:
OK, I will give this PR a little while to get seen (in it's final format) by someone else, then I think we should be good to merge. (finally!)
W00t!
Good job @MigDinny

MigDinny:
@aethelwulffe Thanks Eaton!

tmccormi:
This PR is still tagged as "Work In Progress" but has 3 approvals ... is it ready to merge?  Does it need more testing?

MigDinny:
@tmccormi 2 Approvals actually

Yours would be the 3rd and I think that it is ready to be merged if you agree :)

Issue #915, opened 2018-01-03
headers.inc.php library calling refactor needed


MigDinny:
I realized that we got a function _call_required_libraries_ that a lot of files don't actually use.

Some files actually call some libraries/scripts twice unnecessarily. 

demographics.php is an example I'm working on at the moment that has that kind of problems.

I could work on a refactor of this kind on the remaining of the project. I'll submit a PR to show what I actually did, it's hard to explain here.

teryhill:
The call_required_libraries is to replace the way assets are currently being called. This needs to be done thru out the code base.

aethelwulffe:
Good idea Miguel.

The call_required_libraries is a fairly new function.  We have been 
migrating the system to use that vs. a ton of js scripts on every page.

The refactor process got stalled several times as we changed the 
function itself.  Lots of people took a stab at it.  We went from bool 
values to the current scheme.

Here is how the process works:

Find an include script in a file.  Copy the dependency from 
/libraries/js (or wherever) to /assets/js/name_of_library/index.js (or 
whatever is appropriate in the /assets directory).

Replace all the js asset calls in the file.  Test.

Find all other files that call the same assets.  Save this list.

Identify the files related to the FEATURE you are working on.

Refactor the calls in all FEATURE files.

Test.

Submit a PR for the new function call for that feature.

NEXT:

Identify another feature that includes any files calling the .js 
dependency from libraries/js.

Repeat the above, submit PR.

When no files are left that call the dependency from the old location, 
submit a PR that DELETES the original.

aethelwulffe:
Please review the calls at the bottom of the file for textformat and dialog.js.
Refactoring to put in conditionals will likely be a very big problem, and outside the scope of this PR, but someone screwed up bigtime by adding them in there. Every file that calls this function will need to be checked for actual usage of textformat.js and dialog.js calls before the function arguments to include them can be implemented. 

This should be done before we actually convert any more files to use this header function.

teryhill:
@aethelwulffe  testformat is used to format the phone numbers , text items etc. Dialog does functions for cascading windows grabbing focus etc. These were done by Rod and are used in most every place that jquery is used. I don't believe they will hurt by being called. There should be a project to check which ones are really used and get rid of what is not.

I am the one who "Screwed up"

aethelwulffe:
[...]

PR #926, opened 2018-01-06
interface\new library refactor

MigDinny:
#915 

now using call_required_libraries and jquery upgraded to latest version


aethelwulffe:
Please indicate a merge order for these related PR's.  We don't want to create conflicts.
Which one first?

MigDinny:
@aethelwulffe the order is the id order (#926 -> this, the creation order)

don't know if i made it clear :D

MigDinny:
@aethelwulffe yes, it's better to respect the order first. If any merge conflict happens i will fix it easily

in the end of these merges I will check headers.inc.php file because i touched it so many times i'm afraid he gets broken or something

MigDinny:
Conflicts with master branch solved in the last merge commit.

PR #928, opened 2018-01-06
interface\usergroup library refactor

MigDinny:
#915 

Firstly, I idented all the code that I touched

Secondly, I refactored all libraries not being called by _call_required_libraries_

This was a big refactor I've made in **ALL** files inside _interface\usergroup\\*_


MigDinny:
@aethelwulffe I'll add newlines at some files, didn't pay attention on that

About those "two assets", can you give me a commit example?

What do you mean about the dynarch calendar? I didn't touch it because I think it should be removed and replaced with datepicker, making it useless to add it on headers.inc.php

What missing library calls are you talking about? (the two assets?)

aethelwulffe:
Yeah...lemme check out your "check this out" link.

The missing library calls stuff was me looking at the last lines of headers.inc.php.
After the comment "these are always getting called" the last two include functions were simply deleted (or moved in some way I have not yet found).

Yes, dynarch calendar is NOT in the scope of this PR.

EDIT: I see you reverted stuff.
This (and all of these) PR(s) are very big stuff, and the whitespace (though necessary) does make it a bit difficult to look through.  Glad to do it though.  Testing is also a big piece here, and I have not even begun on that.  Please be patient with the review speed.
  I am working from home, doing these reviews in between calls (my busiest time of year) and since I am working at the kitchen table due to ALL FOUR desks in my house being inaccessible, my butt and knees are killing me. [whine_mode=ON]

MigDinny:
@aethelwulffe :joy: 

I set something on my editor (Sublime Text 3) and it automatically adds a newline at EOF now. I will review all files and give them a new line.

What do you mean by whitespace?

MigDinny:
Please someone check this last commit. There is an error in the console and it is out of the scope of this PR. More instructions on the commit description

teryhill:
@MigDinny is this the ACL menu that you are having this issue on?

MigDinny:
@teryhill probably. i've acessed it through the link: http://localhost/LibreEHR/interface/usergroup/adminacl.php

teryhill:
That IS NOT testing the code.That error does not happen on the demo site. It most likely was introduces by your changes.

MigDinny:
@teryhill realized now that it was introduced by my changes. going to commit a workaround for this

MigDinny:
Fixed merge conflict

interface/usergroup/adminacl.php
```diff
@@ -28,9 +28,7 @@
   ?>
 <html>
   <head>
-    <?php
-      call_required_libraries(array('jquery-min-3-1-1'));
```
teryhill: @MigDinny I don't see why you cannot use jquery-min-1-2-1. By removing this am I correct in the fact that the only change you made was to reformat the code?

teryhill:
This needs to be completly tested before committing it to the code base.

interface/usergroup/adminacl.php
aethelwulffe: If it works with 311, then I see no issue with proceeding towards dumping the old libs.  Ja?  Less code is less code, and with various recent hard-to-patch security flaws in things (admittedly an issue for clients more than servers) we have good reason to tighten our overview of 3rd party assets.

MigDinny:
Existing conflicts with headers.inc.php were solved in this merge commit.

Trodrige:
Hi, @MigDinny. We've moved to GitLab. PR owners are moving their work to GitLab. Can you do this buddy? 
Thanks,
Rodrige.

Trodrige:
I'll ping you to check if I do it.

If you have the time, you can check these instructions I wrote on how migration to GitLab will be: https://forums.librehealth.io/uploads/default/original/1X/f4fa30c18365a512eeedda1ca850af70d2c26792.pdf

aethelwulffe:
Keep your branch, but we need to get the old unmergable PR's refactored, and small enough to test.

Issue #930, opened 2018-01-06
Patient portal - don't loading image


alpinus4:
## Steps to reproduce
1. Go to _Miscellaneous_ ==> _Portal Dashboard_

![step1](https://user-images.githubusercontent.com/25907897/34644022-29897f4e-f32f-11e7-8ced-85d82583e854.jpg)

2. Click _Mail »_ button

![step2](https://user-images.githubusercontent.com/25907897/34644029-49a7d50a-f32f-11e7-92a1-d9a1c01e1e26.jpg)

3. See missing image

![missing image](https://user-images.githubusercontent.com/25907897/34644036-6bbd8c16-f32f-11e7-9d8d-8bf002a9f8f0.jpg)

## Solution
It's caused by incorrect path, so to solve this problem you should correct path to this image.

Testing platform:
Wamp, Firefox 57.0.4, Windows 10 ver. 1709

PR #931, opened 2018-01-06
Fixed missing image in patient portal

alpinus4:
I added new global in globals.php and corrected path to image.
Now it looks like this:

Fixes: https://github.com/LibreHealthIO/LibreEHR/issues/930

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -5,7 +5,7 @@
  *
  * @copyright Copyright (C) 2016 Terry Hill <teryhill@librehealth.io>
  *
- * No header existed on this file so no other copyright information 
+ * No header existed on this file so no other copyright information
  *
  * LICENSE: This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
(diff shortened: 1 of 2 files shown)


interface/globals.php
```diff
@@ -193,6 +193,9 @@ function undoMagicQuotes($array, $topLevel=true) {
 // images directory
 $GLOBALS['images_path'] = "$web_root/assets/images/";
 
+//patient portal images directory
```
aethelwulffe: Probably a good idea for branding sake.

interface/globals.php
```diff
@@ -289,45 +292,45 @@ function undoMagicQuotes($array, $topLevel=true) {
...
     }
-    
+
 
     // change theme name, if the override file exists.
     if( $rtl_override ) {
         // the $css_header_value is set above
         $new_theme = 'rtl_' . $temp_css_theme_name;
 
-        // Check file existance
```
aethelwulffe: existence, not "existance"  If you correct whitespace, might as well correct everything...

patient_portal/messaging/messages.php
```diff
@@ -679,4 +679,4 @@ class="pull-right glyphicon glyphicon-warning-sign text-danger"></span></span> -
     </ng>
 
 </body>
-</html>
\ No newline at end of file
+</html>
```
aethelwulffe: Thank you for adding a new line.

alpinus4: I'm using Atom, and it did it automatically :D
Will look at it tomorrow..

interface/globals.php
alpinus4: Atom corrects whitespace automatically...

Issue #934, opened 2018-01-07
QA Measures - invalid path to globals.php 


alpinus4:
## Steps to reproduce

1. Enable `display_errors` in php.ini

2. Go to _QA Measures_ ==> _Select Measures_ 
    or _QA Measures_ ==> _Delete Old Reports_ 
    or _QA Measures_ ==> _Remove_Provider_Assignments_ 
    or _QA Measures_ ==> _Provider Encounter Counts_

![](https://user-images.githubusercontent.com/25907897/34645211-a89eb6d8-f347-11e7-992a-f1ac2444a418.jpg)

3. See errors

![](https://user-images.githubusercontent.com/25907897/34645218-d29f4542-f347-11e7-832a-1831d175f14b.jpg)

Using Wamp, Firefox 57.0.4, Windows 10 ver. 1709

PR #936, opened 2018-01-07
QA Measures - corrected path to globals.php

alpinus4:
I corected path to globals.php in api.inc.
Fixes: https://github.com/LibreHealthIO/LibreEHR/issues/934

### DIFF
--- a/library/api.inc
+++ b/library/api.inc
@@ -1,6 +1,6 @@
 <?php
 //our api for 3rd party developers
-include_once("/../interface/globals.php");
+include_once("../../interface/globals.php");
 include_once("{$GLOBALS['srcdir']}/sql.inc");
 include_once("{$GLOBALS['srcdir']}/billing.inc");
 include_once("{$GLOBALS['srcdir']}/formdata.inc.php");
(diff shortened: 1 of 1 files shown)


library/api.inc
```diff
@@ -1,6 +1,6 @@
 <?php
 //our api for 3rd party developers
-include_once("../../globals.php");
+include_once("/../interface/globals.php");
```
MigDinny: You shouldn't use _/.._ because this will create the path from **root**

image

alpinus4: Thanks @MigDinny

library/api.inc
```diff
@@ -1,6 +1,6 @@
 <?php
 //our api for 3rd party developers
-include_once("/../interface/globals.php");
+include_once("../../interface/globals.php");
```
MigDinny: image

Look, something's wrong... try to remove the first ../ and test it

aethelwulffe: require_once '../interface/globals.php';
Up one level, down one level.

alpinus4:
Sorry, I will do it tomorrow :)

library/api.inc
alpinus4: Not working :/

alpinus4:
@aethelwulffe Your solution gives error in _encounters_ and _QA Measures_.
I found another solution. Tested it and it didn't throw any errors... Could you take a look?

PR #940, opened 2018-01-08
Patient documents display fix

### LINKED ISSUE #939: Patient Documents not displaying.
This is what the screen should look like

image

This is what it looks like

image

Get to it from documents here.

image

MigDinny:
#939 

Idented the code and added missing styles.

Fixed the bug this way: _$file_ was replaced by _$this->file_ in all occurrences. _var_dump_'s showed me that _$this->file_ was the correct variable to use and _$file_ was null and that was causing the error

The log clearly says:
> Uncaught Error: Call to a member function get_id() on null in 

That told me that get_id() was being called in something null.

Thanks @pri2si17-1997 for the error logs, they were very helpful!

### DIFF
--- a/templates/documents/general_view.php
+++ b/templates/documents/general_view.php
@@ -1,12 +1,16 @@
+<?php
+  // This is required for (at least) date formats
+  require_once($GLOBALS['srcdir']."/formatting.inc.php");
+  require_once($GLOBALS['srcdir']."/headers.inc.php");
+  $DateFormat = DateFormatRead();
+?>
 <head>
+
+  <?php
+    call_required_libraries(array('jquery-min-3-1-1', 'datepicker'));
+  ?>
+
   <link rel="stylesheet" href="<?php echo $GLOBALS['css_header'];?>" type="text/css">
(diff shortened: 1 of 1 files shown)


teryhill:
The calendar dropdown in this code needs to be updated to the new date picker

MigDinny:
@teryhill Thanks Terry! I'm interested in working on that stuff after GCI.

MigDinny:
@teryhill Check it out

formatting.inc.php is required now
headers.inc.php also

(use $GLOBALS because normal variables aren't available here)

call_required_libraries for datepicker and jquery and removed hardcoded dependencies

inserted datepicker on the input (initially i was getting an old design but i realized i was using an older version of datepicker, call_required_libraries solved that)

removed old calendar setup code

pri2si17-1997:
Good @MigDinny :slightly_smiling_face: 👍

aethelwulffe:
I would rather @pri2si17-1997  merged this, but I don't know if we gave him write permissions yet.
If Tony or someone else doesn't do it, or write a review pretty quickly, I will.

MigDinny:
:D

Issue #941, opened 2018-01-09
MigDinny fix this please


teryhill:
missing images for this pr 

![image](https://user-images.githubusercontent.com/4655366/34701308-8816adde-f4b5-11e7-829c-547f9c4ccc20.png)

The images are 

![image](https://user-images.githubusercontent.com/4655366/34701322-98e67608-f4b5-11e7-9c3d-19319f70c8e6.png)

MigDinny:
@teryhill check this out!

aethelwulffe:
This issue was merged.
Bugfix for Miguel (his own bug, but who wants to get all technical?).  Closing this.

PR #944, opened 2018-01-09
Added images folder of jquery-ui

MigDinny:
#941 

This should solve the 404 problems.

I think we shouldn't split our assets folder into **css**, **js** and **images** like this. I think we should have indeed a folder for **css**, **js** and **images** but not for libraries like jqueryUI. Those libraries could have their own folder and those folders (css, js, images) would only be used for single scripts and stuff that wasn't properly a library. Doing this this way splits the libraries unnecessarily, which is the case of jqueryUI. It is splitted among **css** and **js** folder and jqueryUI's **image assets** are inside CSS folder (necessarily because of .css files dependencies). I don't know if I made this clear, I'm bad at explaining stuff


MigDinny:
I don't know why those commits are here

PR #945, opened 2018-01-10
Added images folder to jqueryUI

MigDinny:
#941 

This should solve the 404 problems.

I think we shouldn't split our assets folder into **css**, **js** and **images** like this. I think we should have indeed a folder for **css**, **js** and **images** but not for libraries like jqueryUI. Those libraries could have their own folder and those folders (css, js, images) would only be used for single scripts and stuff that wasn't properly a library. Doing this this way splits the libraries unnecessarily, which is the case of jqueryUI. It is splitted among **css** and **js** folder and jqueryUI's **image assets** are inside CSS folder (necessarily because of .css files dependencies). I don't know if I made this clear, I'm bad at explaining stuff


pri2si17-1997:
@MigDinny Looks good to me :slightly_smiling_face: . Probably I am not able to understand what you say, but you want each library to have its own js, css and images?

MigDinny:
@pri2si17-1997 

What I meant was:

image

There is css folder, js folder and images folder. It's ok to have them because sometimes we got a script and we can put it inside JS folder. But for example, jqueryUI is a library and we have .css files and .js files for it. Currently, we are splitting jqueryUI's library into 2 different folders (css and js) which is kinda messy and can create 404 problems (like this one). One thing I realized was that the images folder i was adding, i couldn't add it in assets/images because it would break dependencies, so, i had to add it into assets/css/jqueryui

I think we should create a folder for each library inside "assets" folder and css, js and images folders would be used NOT for libraries, but for other stuff which is not a library

Issue #946, opened 2018-01-10
Login Redirection Header


MigDinny:
1. Logout from LibreEHR

2. Try to access a file that needs auth [(example)](localhost/libreehr/interface/usergroup/user_info.php)

You realize now that there is a message displayed. What about redirect to login page instead of showing that message?

teryhill:
Pr was merged closing issue

PR #947, opened 2018-01-10
replaced die() with an header for login page redirection

MigDinny:
#946 

die() was triggered with a message saying that basically a login was required. I replaced it with a header to redirect the browser to login-page

### DIFF
--- a/interface/login/login.php
+++ b/interface/login/login.php
@@ -150,6 +150,14 @@ include_once("$srcdir/sql.inc");
                                 </tr>
                             <?php endif; ?>
 
+                            <?php if (isset($_GET['loginfirst'])): ?>
+                                <tr>
+                                    <td colspan='2' class='text' style='color:red'>
+                                        <?php echo xlt('A login is required to access this platform.'); ?>
+                                    </td>
+                                </tr>
+                            <?php endif; ?>
+
                             <?php if (isset($_SESSION['relogin']) && ($_SESSION['relogin'] == 1)): ?>
                                 <tr>
                                     <td colspan='2' class='text'
--- a/interface/login/login.php
+++ b/interface/login/login.php
@@ -153,7 +153,7 @@ include_once("$srcdir/sql.inc");
                             <?php if (isset($_GET['loginfirst'])): ?>
                                 <tr>
                                     <td colspan='2' class='text' style='color:red'>
-                                        <?php echo xlt('A login is required to access this platform.'); ?>
+                                        <?php echo xlt('You have been logged out due to inactivity, or you have experienced an error. Please re-enter access credentials.'); ?>
                                     </td>
                                 </tr>
                             <?php endif; ?>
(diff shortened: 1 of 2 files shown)


MigDinny:
@aethelwulffe Can I create a dialog message with that information? I mean, we got redirected to login page and a dialog opens up telling us that we need to log-in first.

yehster:
The die() in the original code prevents everything else in the script from running. Redirect with headers is ok, but cannot replace the die().  Otherwise this introduces a possible security risk.

aethelwulffe:
It must die().

MigDinny:
can it be header() + die()?

yehster:
Yes, header + die() would be better, the the URL for the redirect needs to pass the current error message still (as Art mentioned before).  I'm not sure if the login page itself can handle that right now, so there may need to be additional changes to the login.php page as well.

MigDinny:
@yehster It doesn't handle those messages, but I can do that. I was just checking with you two if there is a possibility of creating that error message. I'll work on that.

interface/login/login.php
```diff
@@ -150,6 +150,14 @@ function imsubmitted() {
                                 </tr>
                             <?php endif; ?>
 
+                            <?php if (isset($_GET['loginfirst'])): ?>
+                                <tr>
+                                    <td colspan='2' class='text' style='color:red'>
+                                        <?php echo xlt('A login is required to access this platform.'); ?>
```
aethelwulffe: This is fine (testing by setting session timeout to 5 minutes), but I would prefer some sort of explicit statement like "You have been logged out due to inactivity, or you have experienced an error.  Please re-enter access credentials."
The "Site ID is missing from session data!" message used to appear in a single frame when the application was open but attempted an action in that frame (frames are now iframe tabs).  This message was not explicit enough, and created confusion.  Booting to the gateway page is much better in some ways, *but if someone was working in a text field in another frame and got booted, they will lose their unsaved data without being able to copy it to the clipboard.*  Having a modal window re-login of some kind would be very helpful in this case, especially when it is an actual server error like mysql going poof.

My arguments above may not be totally coherent, but if we can moderate 1. confusion as to why they were booted out of the system (inactivity), 2. Loss of unsaved data, then that would be the preferred approach.

aethelwulffe:
My change request was meant to be a comment.  If others are fine with the wording, then I am game.  This is a good thing here.  I just think we should make sure we do this in the best way for the users.  Even as-is, this is better than just a confusing message, with the exception that they instantly lose unsaved data.

MigDinny:
@aethelwulffe honestly I'm super confused... Is any change needed for this PR except the message change? (I will send a commit to change the message to the one you've said)

aethelwulffe:
You don't have to use my phrase word for word.  We just want to tell the users (in plain speech) WHY they got booted.

aethelwulffe:
Yeah, just changing the phrase would please me enough.  All the same, this PR actually has a pretty significant effect in how the application works, so I would like to discuss the effects before merging.

MigDinny:
@aethelwulffe i'm not creative enough, i've used your phrase haha

didn't have any better idea, check it out now

aethelwulffe:
Three-day bump.
Please review.

aethelwulffe:
This is ready for a second review.  Please, someone...

Issue #917, opened 2018-01-05
Inconsistent use of the term "Issues" -vs- "Medical Problems


tmccormi:
Not sure how to resolve this, but if you click on the 'Issues' Link in the Patient Demographics you get this, labeled "Medical Problems" Which we all refer to as "Issues":

![issues-link-from-demog](https://user-images.githubusercontent.com/285835/34590444-b1fafd50-f16b-11e7-9890-2aa876f49464.png)

If you click on the Popup->Issues link you get this, which is used to review the links between "Issues" and Encounters.

![issues-links-to-diff-pages](https://user-images.githubusercontent.com/285835/34590500-0469c6fc-f16c-11e7-9e9c-56aab0b6d2ca.png)

aethelwulffe:
The issues issue has been an issue since the day this was issued.
There is also overlap in the area of "History".

My suggestion from about 10 years ago (11?) was to change everything to "Health Factors".
with the following category names:

Diagnosis (INCLUDES Dental and Mental Health.  Both are covered by ICD codes.  There is no distinction in the difference!)

Allergies (same as above, should include the diagnosis, but this is a difference with a distinction).

Medications

Surgical Outcomes (Might be Treatment History for surgical or any other currently relevant past outcome factors).

-we should get clinical opinions of the above.

tmccormi:
I think we could start with a simple renaming of the visual elements and move on to a better model after that.

supernour09:
hi everyone so i was wondering if i want to fix this issues what is need just rename

Janaki123456:
can u assign me
 this task

Jdew192837:
@tmccormi @aethelwulffe @teryhill 
Could you all assign @ajeyprasand this issue?

PR #950, opened 2018-01-11
"Issues" -vs- "Medical Problems" - renamed

alpinus4:
### Renamed _Issues_ to _Issues and encounters_

Before:

After:

### and _Medical problems_ to _Issues_

Before:

After:

#### Also removed unnecessary `<title>` tags

Fixes: https://github.com/LibreHealthIO/LibreEHR/issues/917

### DIFF
--- a/interface/patient_file/problem_encounter.php
+++ b/interface/patient_file/problem_encounter.php
@@ -6,7 +6,7 @@
  * Copyright (C) 2005 Rod Roark <rod@sunsetsystems.com>
  * Copyright (C) 2015 Roberto Vasquez <robertogagliotta@gmail.com>
  * Copyright (C) 2015 Brady Miller <brady@sparmy.com>
- * 
+ *
  * LICENSE: This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
  * as published by the Free Software Foundation; either version 2
@@ -72,7 +72,7 @@
     ") VALUES ( " .
     " ?, ?, ?" .
     ")";
-   sqlQuery($query, array($form_pid, $list_id, $encounter)); 
+   sqlQuery($query, array($form_pid, $list_id, $encounter));
   }
 
   echo "<html><body>"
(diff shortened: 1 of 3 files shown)


alpinus4:
@tmccormi

aethelwulffe:
Sorry @alpinus4 , could you pop in before/after screenshots?
I still think that "popups" is a terrible tab title.

It seems like "Popups" actually means widgets that interact with the *current patient*.  To me, I think they should be in the patient/client tab, or next to it, or arranged some other way.  "Popups" doesn't say anything.  Most folks will never even look in that tab.  Doesn't seem like anything they would want.

So:  Some of the "Popups" like the "Issues" one should probably belong to "Patient/Client".  Frankly, so should barcode and a number of other things...though they are less commonly used, but in a sub-menu.

Now the name of "Issues".  Frankly, these are basic medical charting items, but "Chart" means different things to different people.  I think that "Health Factors" _might_ be a good name option.
I would like to hear a clinical opinion on this. @yehster 

All the same @alpinus4 good job so far.

teryhill:
Screen Re-Design for better work flow!!! lets go that route.

aethelwulffe:
@alpinus4   It would be really nice if we can tweak this into a real workflow improvement.
I will come up with some menu json suggestions if you like, but I bet you could make a few tab group changes and come up with something significantly better than it currently is.

Feel like doing that?  We could do it here or in a different PR.

teryhill:
This could be a GSoC Project or one for @alpinus4 after GCI

aethelwulffe:
@teryhill agreed.

All the same, I think that "Issues and Encounters" is not the right name for this feature.

alpinus4:
@aethelwulffe Sorry for the delay, whole day yesterday worked on _managing encounters webcomponent_.
I added _Before_ and _After_ screenshots :)
And... 
> Feel like doing that? We could do it here or in a different PR.

Hmm I think .. yes.. I can do it after GCI, ... but also want to work on my own projects, so it won't be a work as intense as during GCI...

alpinus4:
@aethelwulffe Renamed.
#### After:

PR #951, opened 2018-01-12
QA measure updates

aethelwulffe:
QA measure updates for 2017 reporting.


aethelwulffe:
Must still fix the patient_select changes that broke MIPS reporting, but that is a different PR.
These changes and this code has been tested in a nearly identical branch, and is in production.

PR #952, opened 2018-01-13
interface\login & interface\logview library refactor

MigDinny:
#915 

This was only 3 files. Everything was tested and no errors found.


MigDinny:
@aethelwulffe agree

aethelwulffe:
I need to test again, but this may still be good.  Use ?w=1

We have some inconsistency when calling call_required_libraries(), sometimes using array() and other using [], but it all works.

Issue #962, opened 2018-01-20
Missing logo in printable versions of reports


alpinus4:
# Steps to reproduce:
**1.** Go to patient screen and click _Report_

![](https://user-images.githubusercontent.com/25907897/35188011-4d7d96ba-fe2e-11e7-81ba-073f1d283ffe.jpg)

**2.** Click 3, 4, 5 or  6  _Generate report_ button
![](https://user-images.githubusercontent.com/25907897/35188140-cc2fec86-fe30-11e7-8839-1fc89a7ae288.jpg)

**3.** Go to _Printable Version_ of report
![](https://user-images.githubusercontent.com/25907897/35188146-dd91e13c-fe30-11e7-9f5d-046a53a527e8.jpg)

**4.** See missing image
![](https://user-images.githubusercontent.com/25907897/35188139-c8653818-fe30-11e7-8a07-44d9b95e0064.jpg)

Browser: Firefox 57.0.4
OS: Windows 10 ver. 1709

teryhill:
Most likely someone moved the images with out changing all the code.

teryhill:
$practice_logo = "$OE_SITE_DIR/images/practice_logo.png";

Yep the code needs to be changed

teryhill:
Looked and the images are still in the sites/images directory so missing images is not the issue. It appears that the code is seeing that an image exists but it cannot properly display it.

2208Abhinav:
I solved this problem.
Now the page looks like this
![screenshot from 2018-01-21 18-15-38](https://user-images.githubusercontent.com/30211121/35191759-4e187948-fea9-11e7-9820-25c0bf9c7671.png)

2208Abhinav:
I have pulled a request #960  for this including issue #949

teryhill:
Pr merged

PR #965, opened 2018-01-21
Solution for missing logo in printable versions of reports issue #962

2208Abhinav:
Logo was not previewed in the printable version of report

Before
35188139-c8653818-fe30-11e7-8a07-44d9b95e0064

Now
screenshot from 2018-01-21 18-15-38

As you can see I have reduced the size of image to 56px by 45px because without this the result was

screenshot from 2018-01-21 21-06-29

### DIFF
--- a/interface/patient_file/report/custom_report.php
+++ b/interface/patient_file/report/custom_report.php
@@ -483,18 +483,7 @@ if ($printable) {
   // in HTML view it's just one line at the top of page 1
   echo '<page_header style="text-align:right;" class="custom-tag"> ' . xlt("PATIENT") . ':' . text($titleres['lname']) . ', ' . text($titleres['fname']) . ' - ' . $titleres['DOB_TS'] . '</page_header>    ';
   echo '<page_footer style="text-align:right;" class="custom-tag">' . xlt('Generated on') . ' ' . oeFormatShortDate() . ' - ' . text($facility['name']) . ' ' . text($facility['phone']) . '</page_footer>';
-
-  //use logo if it exists in directory
-  if (file_exists("$OE_SITE_DIR/images/practice_logo.png"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.png";
-  else if (file_exists("$OE_SITE_DIR/images/practice_logo.gif"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.gif";
-  else if (file_exists("$OE_SITE_DIR/images/practice_logo.jpg"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.jpg";
-  else if (file_exists("$OE_SITE_DIR/images/practice_logo.jpeg")) $practice_logo =  "$OE_SITE_DIR/images/practice_logo.jpeg";
-  else $practice_logo = '';
-   $dir = explode('/', $practice_logo) ;
-   $practice_logo = "$dir[4]/$dir[5]/$dir[6]/$dir[7]/$dir[8]" ;
-   if ($practice_logo) {
-        echo "<img src='../../../../$practice_logo' align='left' style='width : 56px ; height : 90.6px'><br />\n";
-   }
+  
 ?>
 <h2><?php echo $facility['name'] ?></h2>
 <?php echo $facility['street'] ?><br>
--- a/interface/patient_file/report/custom_report.php
+++ b/interface/patient_file/report/custom_report.php
@@ -484,12 +484,7 @@ if ($printable) {
   echo '<page_header style="text-align:right;" class="custom-tag"> ' . xlt("PATIENT") . ':' . text($titleres['lname']) . ', ' . text($titleres['fname']) . ' - ' . $titleres['DOB_TS'] . '</page_header>    ';
   echo '<page_footer style="text-align:right;" class="custom-tag">' . xlt('Generated on') . ' ' . oeFormatShortDate() . ' - ' . text($facility['name']) . ' ' . text($facility['phone']) . '</page_footer>';
 
-  // Use logo if it exists as 'practice_logo.png' in the site dir
-  // old code used the global custom dir which is no longer a valid
-   $practice_logo = "$OE_SITE_DIR/images/practice_logo.png";
-   if (file_exists($practice_logo)) {
-        echo "<img src='$practice_logo' align='left'><br />\n";
-     } 
+  echo "<img src='../../../sites/default/images/practice_logo.png' align='left' style='width : 45px ; height : 56px'><br>";
 ?>
 <h2><?php echo $facility['name'] ?></h2>
 <?php echo $facility['street'] ?><br>
--- a/interface/patient_file/report/custom_report.php
+++ b/interface/patient_file/report/custom_report.php
@@ -484,7 +484,17 @@ if ($printable) {
   echo '<page_header style="text-align:right;" class="custom-tag"> ' . xlt("PATIENT") . ':' . text($titleres['lname']) . ', ' . text($titleres['fname']) . ' - ' . $titleres['DOB_TS'] . '</page_header>    ';
   echo '<page_footer style="text-align:right;" class="custom-tag">' . xlt('Generated on') . ' ' . oeFormatShortDate() . ' - ' . text($facility['name']) . ' ' . text($facility['phone']) . '</page_footer>';
 
-  echo "<img src='../../../sites/default/images/practice_logo.png' align='left' style='width : 56px ; height : 90.6px'><br>";
+  //use logo if it exists in directory
+  if (file_exists("$OE_SITE_DIR/images/practice_logo.png"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.png";
+  else if (file_exists("$OE_SITE_DIR/images/practice_logo.gif"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.gif";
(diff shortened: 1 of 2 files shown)


alpinus4:
In my previous comment I meant keeping the proportions @2208Abhinav
It's important

2208Abhinav:
I get it you want that the ratio of dimensions should relate original image

2208Abhinav:
Please review the changes I have changed to 90.6 X 56
Original Size 2913 X 1800
screenshot from 2018-01-21 21-18-11

2208Abhinav:
@alpinus4 I didnt't get your point could you please elaborate.

alpinus4:
@teryhill @aethelwulffe Could you take a look? Is there necessary to cange commits messages?

@2208Abhinav In the future don't use commit messages like update xyz, update xyz, .....
Such messages don't give too much information. Won't be better when you use for ex. _Changed image size in xyz.php_ or _Corrected image path_ or anything similar??

2208Abhinav:
Thank you buddy now I understand

interface/patient_file/report/custom_report.php
```diff
@@ -484,12 +484,7 @@ function next_prev(action){
   echo '<page_header style="text-align:right;" class="custom-tag"> ' . xlt("PATIENT") . ':' . text($titleres['lname']) . ', ' . text($titleres['fname']) . ' - ' . $titleres['DOB_TS'] . '</page_header>    ';
   echo '<page_footer style="text-align:right;" class="custom-tag">' . xlt('Generated on') . ' ' . oeFormatShortDate() . ' - ' . text($facility['name']) . ' ' . text($facility['phone']) . '</page_footer>';
 
-  // Use logo if it exists as 'practice_logo.png' in the site dir
-  // old code used the global custom dir which is no longer a valid
-   $practice_logo = "$OE_SITE_DIR/images/practice_logo.png";
-   if (file_exists($practice_logo)) {
-        echo "<img src='$practice_logo' align='left'><br />\n";
-     } 
+  echo "<img src='../../../sites/default/images/practice_logo.png' align='left' style='width : 56px ; height : 90.6px'><br>";
```
aethelwulffe: At the beginning of this file, the `require_once("../../globals.php");` allows you to use the variable $OE_SITE_DIR
Without this variable, if you have more than one /site directory, this code will be broken for any site other than /sites/default.
  Revise this to use the current site session variable above.

2208Abhinav: okay I will try to implement that, thanks for review.

aethelwulffe:
Next issue with this:
Most clinics will not want the Librehealth logo plastered on the reports.
This code might be best simply taken out, or use a conditional statement that renders the logo if the configuration includes a clinic-specific letterhead.  More to the point, If the feature is included at all, it should have a toggle, and/or specifically have an HTML letterhead file that it uses/includes in these cases.  The letterhead file should be editable within the application, and be /sites directory specific.  There is even a case for making it specific to the facility/clinic combination.

2208Abhinav:
You want to include two options 
First clinic's logo
If that not present than Librehealth logo

teryhill:
The current code looks for the practice_logo if it exists it is used. The user can put what ever logo they want there. I don't think any code changes are needed unless you want to make one to use fidderent file types, jpg,gif,png etc.

2208Abhinav:
So now you want me to just use $OE_SITE_DIR variable in the image path and provide no other options?

2208Abhinav:
@teryhill I also agree with the point that user can manually put the required logo in the directory.

teryhill:
@2208Abhinav  look into allowing the user to put any type of image file in the directory. like they have a practice_logo.gif, or practice _logo.png or practice_logo.jpeg. something like this:

$practicelogo = preg_grep('~practice_logo\.(gif|png|jpg|jpeg)$~i', $practicelogo);

2208Abhinav:
@teryhill please review the changes
I have used the `explode()` function to get the absolute directory of the image while using the `$OE_SITE_DIR` variable as mentioned by @aethelwulffe.
I did this because in the actual code the image path was rendered as **http://localhost/LibreEHR/interface/patient_file/report/sites/default/images/practice_logo.png** that's why the image was not shown and to get the correct path **http://localhost/LibreEHR/sites/default/images/practice_logo.png**.

I have allowed the facility to use any extensions use conditional statements.

aethelwulffe:
Any of the above is fine.  Something in documentation or an obvious tool in the globals config would be better, but this is fine.
The practice logo != login logo, then we are much better off.  I still think that a custom letterhead tool would be a nice feature.

interface/patient_file/report/custom_report.php
```diff
@@ -484,12 +484,17 @@ function next_prev(action){
...
 
-  // Use logo if it exists as 'practice_logo.png' in the site dir
-  // old code used the global custom dir which is no longer a valid
-   $practice_logo = "$OE_SITE_DIR/images/practice_logo.png";
-   if (file_exists($practice_logo)) {
-        echo "<img src='$practice_logo' align='left'><br />\n";
-     } 
+  //use logo if it exists in directory
+  if (file_exists("$OE_SITE_DIR/images/practice_logo.png"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.png";
+  else if (file_exists("$OE_SITE_DIR/images/practice_logo.gif"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.gif";
```
teryhill: Would it be easier to add a global variable and let the user specify the name for the practice logo, then they could specify practice_logo.gif or practice_logo.tiff etc

interface/patient_file/report/custom_report.php
```diff
@@ -484,12 +484,17 @@ function next_prev(action){
...
+  else if (file_exists("$OE_SITE_DIR/images/practice_logo.jpg"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.jpg";
+  else if (file_exists("$OE_SITE_DIR/images/practice_logo.jpeg")) $practice_logo =  "$OE_SITE_DIR/images/practice_logo.jpeg";
+  else $practice_logo = '';
+   $dir = explode('/', $practice_logo) ;
+   $practice_logo = "$dir[4]/$dir[5]/$dir[6]/$dir[7]/$dir[8]" ;
+   if ($practice_logo) {
+        echo "<img src='../../../../$practice_logo' align='left' style='width : 56px ; height : 90.6px'><br />\n";
```
teryhill: Could a variable be added to the globals like $OE_SITE_DIR to get this. The hard coding of the path information will cause issues down the road when things are changes and someone forgets to change this.

interface/patient_file/report/custom_report.php
```diff
@@ -484,12 +484,17 @@ function next_prev(action){
...
-        echo "<img src='$practice_logo' align='left'><br />\n";
-     } 
+  //use logo if it exists in directory
+  if (file_exists("$OE_SITE_DIR/images/practice_logo.png"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.png";
+  else if (file_exists("$OE_SITE_DIR/images/practice_logo.gif"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.gif";
+  else if (file_exists("$OE_SITE_DIR/images/practice_logo.jpg"))  $practice_logo = "$OE_SITE_DIR/images/practice_logo.jpg";
+  else if (file_exists("$OE_SITE_DIR/images/practice_logo.jpeg")) $practice_logo =  "$OE_SITE_DIR/images/practice_logo.jpeg";
+  else $practice_logo = '';
+   $dir = explode('/', $practice_logo) ;
+   $practice_logo = "$dir[4]/$dir[5]/$dir[6]/$dir[7]/$dir[8]" ;
```
teryhill: I am not sure what you are doing here>

interface/patient_file/report/custom_report.php
aethelwulffe: Should be a global config to use an optional file specifically for this purpose...not sure if we even need a logo on a report like this at all.
  My final vote is that either we make the effort to put in a letterhead feature, or get rid of the image here.  It does nothing for the quality of the report.

2208Abhinav: I also suggest that there is no specific need of image in the report.

aethelwulffe: Abhinav,
Could you just do us all a favor and get rid of the image completely for now?  That is the simple and sensible solution.  It should have never been there in the first place.
Thanks,
-Art

2208Abhinav: @aethelwulffe I suggest two options 
1)To completely remove the image.
2)To put a general medical logo like a plus sign or something like that.
Which one you favor?

aethelwulffe: Totally remove the logo is my preference, but I am open to either option.

2208Abhinav: Thank you for telling I will remove it.

aethelwulffe: Just waiting on the new commit, and then I will test and merge.  I imagine this has been a very irritating issue for you.  Sorry such a simple thing got so complex.

2208Abhinav:
I have completely removed the image, @aethelwulffe  please review the changes.
And sorry for the long absence, I got ill in between the process but now I am back to work :smiley:

2208Abhinav:
I also worked for the general medical  logo.
Here's the result
screenshot from 2018-01-27 19-20-49

aethelwulffe:
Thanks!
  We can worry about adding any letterhead or logos in another project.  This is good, clean, tested, and approved...and a very simple commit.
Merging....

Issue #967, opened 2018-01-25
Changes to the procedure order form. Currently it is just a lab order form, but other procedures should be allowed.


teryhill:
Currently the procedure order form will send information to any Labs that are setup to Receive orders from the clinic. There is a dropdown at the bottom (which will need to be moved to the top of the form) of the form in which you could tag the procedure as some thin else. If any thing in the bottom section is chosen besides procedure or laboratory test there needs to be a print button. The field that is called procedure would need to look for the information as a different data type in the procedure codes table. 

![image](https://user-images.githubusercontent.com/4655366/35411328-5ab71fe8-01e6-11e8-979b-ffa4a43b36cd.png)

tmccormi:
Add some controls to the Orders List that designate the type as "manual/electronic/both" to control whether the there is a Save/Transmit  Save/Print button visible.

tmccormi:
I would rename this Form to be "Orders" in general, not Procedure Orders.

aethelwulffe:
Agree on all points in this issue.

teryhill:
Changes here to accommodate the electronic or paper options. Add two check boxes one for paper and one for electronic. This will allow for one or the other and both.

![image](https://user-images.githubusercontent.com/4655366/35414488-9969b48a-01f0-11e8-96b6-3135e20d7680.png)

aethelwulffe:
Counseling, Psych eval....onandonandon.

teryhill:
Changes so far

![image](https://user-images.githubusercontent.com/4655366/35414970-129b5754-01f2-11e8-8a71-be8465fe23b5.png)

htuck:
Re: documenting the processing of a provider's order, I believe that everybody who touches it has to put their name to it.  I.e., the ordering provider has to sign it, and the transcribing staff has to sign that it was transcribed.  See the ovals in the pic below of the prov Order display in the encounter summary; is the encounter form's 'procedure order by' name legally equivalent to an e-sig?  If so, can have the transcribing staff e-sig the Provider Order.  Else will need the provider to e-sig the order form itself, and have the transcriber e-sign somewhere else that it was transcribed.

Would it be hard to add another e-sig button to the orfer form, saying 'transcribed'?  I have no idea what other things are touched by an e-sig.
Incidentally, how has this requirement been dealt with in the past, needing legally binding documentation of provider orders having been transcribed?

In other news, the results field (rectangle) will need to be cut out for non- lab orders.

[...]

PR #968, opened 2018-01-26
Changes for order type

teryhill:
Changes for order type #967


teryhill:
This will give Harley the setup options for the video

teryhill:
This will show the print button. Moved the drop down to the top these changes should get Harley thru the video.

teryhill:
@tmccormi this is the pass to get harley going with the video. The frame is there will just need to connect the dots. Perhaps one of the GSoC folks would like to finish this up as a task. If not I will try to get back to it first of next week.

PR #969, opened 2018-01-26
Access control based on Patient Facility

aethelwulffe:
Feature addition to allow a clinic to restrict access to all patient
data based on the confluence of the `user` table schedule_facilities and
the `patient_data` 'facility' fields.

This feature allows multiple clinics to share a database, while
restricting whole popluations of patient data from view.  This has the
effect of both controlling access as well as simplifying the data for
users that have no need to see what is going on in other physical or
organizational clinical service locations.

This feature is of extreme importance for Behavioral Health,
Secure/institutional Care facilities, and Assisted Living Facility
organizations.  It is also very useful for temporary care provider
situations and many others.

This is the third attempt at implementing this feature.  Prior PR's
included data schema changes that were controversial, but those changes
have since been included in recent pull requests from other
contributors.  Rather than have the history of those discussions, this
branch was created fresh for simplification of review.


library/globals.inc.php
```diff
@@ -610,6 +611,12 @@
        '0',                             // default
       xl('Restrict non-authorized users to the Schedule Facilities set in User admin.')
     ),
+        'facility_acl' => array(
+      xl('Restrict User access by Facility'),
+       'bool',                          // data type
+       '0',                             // default
+      xl('Restrict User access to patients by assigned patient facility.')
+    ),
```
teryhill: When the role based menus are utilized this should be a function of a setting in the user field. Just a note.

interface/tags_filters/start.php
```diff
@@ -86,4 +86,93 @@ function update_tags_filters()
 }
 add_action( 'update_plugin', 'update_tags_filters' );
 
-//}
\ No newline at end of file
+///////////////////////////////////////////////////////////////////////////////
+////test add of facility acl stuff.
```
tmccormi: Should this comment say "Test ..."

aethelwulffe: No....I will change.

aethelwulffe:
Added START and END comments for the block of functions related to User Facility Access Control.

@teryhill this *may* be an example of how to use the "do_action" function calls.  I am very sorry that I didn't have the skill necessary to get the whole tags setup moved out to a module, and do the same with this code.  I am piggybacking on top of the tags.

aethelwulffe:
Those affects are *currently* controlled by normal ACL access restrictions.  Yes, there is some expansion needed.

teryhill:
Could be included in the GSoC report generator project.

aethelwulffe:
Also, the schedule_facilities selector in admin/users sucks bigtime (more stuff like this needs to be handled in other screens anyway), user facility-specific provider info screen needs grouping per-provider or per-facility, needs multiple taxonomy code stuff, and a lot of other user-config at the admin level functions.

There is also one remaining boo-boo.  Patient Demographics for the `facility` field are not set as a required field in search/add, and that means it is possible to add a patient that does not have a facility...meaning the patient gets orphaned and no-one can see it without the wonders of direct database admin tools.

teryhill:
Make an issue for those after this is committed. See if Mig wants to do the work or one of the other GCI folks.

tmccormi:
Those are all the responses I expected.  Having tried this before (the whole reason the Tags stuff was created, in fact) ...

aethelwulffe:
@teryhill Will do.  I intend to take a stab at moving this stuff to the /modules directory (as well as tags...whatever they do), and will try to make some implementation for some reports or possibly (but not probably) certain billing features.

I think my main issue there was use of globals in all the includes...or something...hard to remember all the problems.  Most of them involved identifying the bits of scattered code that actually made things work vs the "other stuff"

tmccormi:
I'll proceed with the merge...

aethelwulffe:
At least it has a toggle...

teryhill:
@aethelwulffe let me know how to let the system know that there is a module that needs to be used.

PR #977, opened 2018-01-31
Adding global toggles for TAGS and Facility ACL

aethelwulffe:
Merged PR did not include global conditional statements.
-Requires review of functionality of the tags as well as the facility
access control.

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -2378,7 +2378,7 @@ $GLOBALS_METADATA = array(
 
     'SMTP_PASS' => array(
       xl('SMTP Password for Authentication'),
-      'text',                           // data type
+      'pass',                           // data type
       '',                               // default
       xl('Must be empty if SMTP authentication is not used.')
     ),
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -74,7 +74,8 @@ if (stristr(PHP_OS, 'WIN')) {
   $perl_bin_dir        = 'C:/xampp/perl/bin';
   $temporary_files_dir = 'C:/windows/temp';
   $backup_log_dir      = 'C:/windows/temp';
-} else {
+}
+else {
   // Everything else
   $mysql_bin_dir       = '/usr/bin';
   $perl_bin_dir        = '/usr/bin';
@@ -604,6 +605,12 @@ $GLOBALS_METADATA = array(
       xl('Support export/import of configuration data via the Backup page.')
     ),
 
+    'tags_filters_enabled' => array(
+      xl('Enable Demographics Tag filtering feature'),
+       'bool',                          // data type
+       '0',                             // default
+      xl('Enable patient Tag filtering feature.')
+    ),
     'restrict_user_facility' => array(
       xl('Restrict Users to Facilities'),
        'bool',                          // data type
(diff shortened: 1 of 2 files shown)


aethelwulffe:
I have not tested this at all yet.  This is a work in progress in the very short near term.

library/globals.inc.php
```diff
@@ -2372,7 +2378,7 @@
 
     'SMTP_PASS' => array(
       xl('SMTP Password for Authentication'),
-      'pass',                           // data type
+      'text',                           // data type
```
teryhill: You are clobbering my last commit

library/globals.inc.php
```diff
@@ -74,7 +74,8 @@
   $perl_bin_dir        = 'C:/xampp/perl/bin';
   $temporary_files_dir = 'C:/windows/temp';
   $backup_log_dir      = 'C:/windows/temp';
-} else {
+}
+else {
```
teryhill: You are clobbering my last commit

teryhill: also did you mean to add features back to the user menu . It has no items under it.

teryhill:
So tell me how I test it?

library/globals.inc.php
aethelwulffe: Boy, did I hose this up...

I think you test this by giving me a minute to make another go...this commit was crap.

library/globals.inc.php
aethelwulffe: I will rebase and have another go...I missed this.

library/globals.inc.php
aethelwulffe: No.  I suppose you mean in another PR.  I did not intend to re-add features.

aethelwulffe: features has items under it...
I don't understand.

teryhill: That section is for the user preferences under miscellaneous

teryhill: Don't worry about it . I will fix it after the demos get up

aethelwulffe: demo site is a mess.

aethelwulffe:
I am testing this branch from ground zero.  Creating everything from scratch.  Your install isn't working as it should, the demo is hosed up, and my install did everything exactly how I expected.

teryhill:
Things I have noticed. You set the patient file to have a facility of '1" we send out the code with a facility of "3" so there will never a facility numbered 1. on my system I have them as 4 and 5.

I have tried every combination I can think of to break my system but it will not.

aethelwulffe:
Never noticed that...wonder why we do it... doesn't seem to set the default anyway.

teryhill:
My entire patient_data table has 1 except for the patient I changes and he has a 5

aethelwulffe:
By adding two patients normally after a clean setup and UI only configuration, I have a two patients, one of whom I set to Facility 3 (the renamed default) and the other patient has a 0. 
  Nothing is working since I killed the old DB, I snuffed the old branch where this was working, and I only have a copy of the code intended for BPUC to use.  

I guess we have to revert this completely again.  I will see if I can re-create the feature, maybe outside of the "plugin" system.

aethelwulffe:
Closing this PR.  Putting in an issue to revert the implementation of this feature.
If we do it, I suppose we should do it just using the tags system.
I will do a write-up on how to use it, put in issues to take care of the form formatting issues (really screwy) and create tools to do mass-setup features that link facilities to tags.

PR #978, opened 2018-02-01
Add Lost Plugin Doc back

tmccormi:
It was removed at some point in a cleanup ...


tmccormi:
No need to really review this, unless you want some more info about it ...

aethelwulffe:
Questions:
There are broken hrefs in the plugins files.  This is in the "autoload" function/file.  What were those supposed to do?
Second:  Does the composer.json file name actually have anything to do with composer?

Perhaps the instructions really are clear, and it is just all the naming conventions that make everyone really really unsure of what is actually going on.  It is easy to confuse things by using bad names.  Example:  I write a variable named $_off.  This variable is actually used to enable something.  I add a variable named $_on.  This variable is actually used to run a fork bomb.

Issue #966, opened 2018-01-25
Admin -> Lists start at "language" instead of at the top of the list


tmccormi:
The selection of the list, for some reason has always started with Language, instead of the Alphabetical top of the list

This is a easy task.
Current:
![current-default](https://user-images.githubusercontent.com/285835/35410721-26003430-01cb-11e8-9658-b1f384d12915.png)

Proposed:
![alphabetical-top](https://user-images.githubusercontent.com/285835/35410731-30aba93c-01cb-11e8-9df3-f8d15f2ac3f9.png)

tmccormi:
Extra points for a search as you type list selector :-)

aethelwulffe:
TEN!  Ten extra points for a type-ahead selector!

teryhill:
@Perseus it should display the first item in the list. I am not sure what @aethelwulffe  is asking for on the type ahead cause I have always just used the first letter of a known list.

Perseus:
Great. I'll do a PR first thing tomorrow morning with that fix!

tmccormi:
type ahead filters the list as you type each letter.  it can be 'sequential' as in the order of the list item or "contains" as in filter the list based on anything the matches what has been typed so far.   We need the second because the list names are not consistent in anyway.   For instance try and find the list that has the "rooms" in it without scrolling the whole thing...

Perseus:
@tmccormi Would a JS library to make that happen be allowed or do I need to make a native solution?

teryhill:
@Perseus Look in the assets directory and see if there is an asset there that you can use.

Perseus:
@teryhill I looked around but was not able to find anything that would be useful. The library I was proposing is this: https://selectize.github.io/selectize.js/

Perseus:
@teryhill I looked around but was unable to find anything that would be useful. The library that I was proposing is this : https://selectize.github.io/selectize.js/

teryhill:
Okay use it.

Perseus:
Is there any intention to eventually update the jQuery versions? The edit list page is currently using 1.2 which is quite outdated and hard to use current libraries in

aethelwulffe:
@Perseus 
Check out the /library/headers.inc.php file.  We invoke that by calling globals.php, then using the directory variables there to call this file, then invoke the functions to call whatever assets are required as an array.
[...]

PR #980, opened 2018-02-01
Fixed the list to display entries alphabetically

Perseus:
Fixes issue #966 

Removed the conditional check for `$list_id` hence displaying the list items, which are retrieved alphabetically from the database, in order.


Perseus:
I'm also working on the type-ahead selector. Wanted to know how you guys wanted it done. A separate text field which the user can type in to find a list or something else?

Perseus:
@tmccormi

teryhill:
Fails testing.

Perseus:
Is the top option still "language"? What is the issue?

teryhill:
With the change you get this

image

The option changes in the drop down but the listed items do not change. Then if you select a list you get this.

image

Which shows the list but the wrong header.

teryhill:
This is what I came up with.

image

Then when you select a list you get the list and the save button.

image

Perseus:
@teryhill I changed it to be blank initially and nothing displays and then can be selected as needed. 
2438bc9e03fc3f0e1ce0f864c8ccc03f
dbed3daef352c64399db5a38f4f83f67
f6438d0248f62a2a29c6e32cbae7dbfb
Does this look fine?

teryhill:
Yes it does look fine.

interface/super/edit_list.php
```diff
@@ -858,7 +861,7 @@ function mysubmit() {
 </p>
 
 <center>
-
+<?php if ($list_id != ' '): ?>
```
teryhill: Should use {

interface/super/edit_list.php
```diff
@@ -1046,7 +1049,7 @@ function mysubmit() {
 </center>
 
 </form>
-
+<?php endif; ?>
```
teryhill: Should use }

interface/super/edit_list.php
Perseus: Alright I'll fix that.

aethelwulffe:
OK, I just went and tested an out-of-date commit...Lemme try again.

Perseus:
@aethelwulffe Are there any future plans to upgrade the jQuery versions being used? The edit list page uses 1.2 and most of the libraries having type-ahead selectors don't really support that. I could work on something just for that but it'd be easier if jQuery was updated instead. Less legacy code to maintain

teryhill:
@Perseus it is an on going project. Issue 416 has the info on it.

Perseus:
Please review the commit @teryhill @aethelwulffe . I couldn't find any errors with the updated jquery version in the edit_list page and the list seems to work properly. Please let me know if any changes are needed.

interface/super/edit_list.php
```diff
@@ -618,9 +618,11 @@ function writeITLine($it_array) {
 <?php html_header_show();?>
 
 <!-- supporting javascript code -->
-<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/jquery.js"></script>
-
+<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/jquery-3.3.1.min.js"></script>
```
aethelwulffe: `require_once $GLOBALS['srcdir'].'/headers.inc.php'; ` is on line 22 of this file.  

To call a single library from the assets/js/ directory, use `function include_js_library($path)` where $path is the directory name of the asset under assets/js/ and the actual file name is index.js.

`function include_css_library($path)` is the same as the above, just requires the full file path.

`call_required_libraries($library_array)` allows multiple libraries to be called.  In your case, you would invoke it with `call_required_libraries('jquery-min-3-3-1','select2');`

The above requires renaming the files to index.js, and requires that the directory naming standard be put into the jquery-min-* naming standard.  Either a default case switch or a final elseif else statement needs to be added that invokes whatever path is passed in the argument array, OR add "select2" to the list of 'if' conditions.

These functions require some rework but when we change the functions themselves to make them more efficient (lots of people touched the file) and provide default vs. specified behavior of include_js_library($path) to be added to call_required_libraries , but we should not have to ever change the calls.

Perseus:
@aethelwulffe I've moved the jquery file to the assets directory,  added a select2 clause to the headers file and added the required libraries function call to the edit list page

aethelwulffe:
Yep, all the pieces work.

Issue #982, opened 2018-02-01
View Facility Specific User Information


aethelwulffe:
Feature is broken.
Administration/Users/View Facility Specific User Information
Tab results [UNKNOWN]

teryhill:
Fix on the way

aethelwulffe:
Tested, approved with screenshot.
It is a one-character bug-fix.  I suggest we go all maverick and merge...  @tmccormi  ?

aethelwulffe:
This is merged.
Thank you Terry.

PR #984, opened 2018-02-01
Fix for unknown

teryhill:
Fix for unknown #982  The fix is in line 71 all else is refactoring by the editor


aethelwulffe:
Lemme switch to this branch and see what happens.

This is what happens.  Looks good.
image

aethelwulffe:
This is a simple and obvious bugfix. 
Feature didn't work, now it does.  We sat on it for a day, gonna merge now.

PR #987, opened 2018-02-02
Added a function that will give information for all insurances.

teryhill:
Sherwin Gaddis used this function and It will be used in the Order form
changes. #967 
I also included the suffix in the getproviders function.

### DIFF
--- a/library/patient.inc
+++ b/library/patient.inc
@@ -1440,9 +1440,9 @@ function getPatientAge($dobYMD, $nowYMD=null)
         $age = $yearDiff;
         if (($monthDiff == 0) && ($dayDiff < 0)) { $age -= 1; }
         else if ($monthDiff < 0) { $age -= 1; }
-    }
-    else  {
-        $age = "$ageInMonths " . xl('month');
+    
+    } else  {
+        $age = "$ageInMonths " . xl('months');
     }
 
     return $age;
--- a/library/patient.inc
+++ b/library/patient.inc
@@ -166,6 +166,7 @@ function getProviders() {
     $rez = sqlStatement($sql);
     for($iter=0; $row=sqlFetchArray($rez); $iter++) {
         if (($row["fname"] != "") && ($row["lname"] != "")) {
+          if ($row["suffix"] != "") $row["lname"] .= ", ".$row["suffix"];
             array_push($returnval, $row["fname"] . " " . $row["lname"]);
         }
     }
@@ -1082,6 +1083,7 @@ function updatePatientData($pid, $new, $create=false)
   // the code below does not yet attempt it.
 
   $new['DOB'] = fixDate($new['DOB']);
+  $user=$_SESSION['authUser'];
 
   if ($create) {
     $sql = "INSERT INTO patient_data SET pid = '$pid', date = NOW()";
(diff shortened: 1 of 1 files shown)


library/patient.inc
```diff
@@ -1439,7 +1442,7 @@ function getPatientAge($dobYMD, $nowYMD=null)
         else if ($monthDiff < 0) { $age -= 1; }
     }
     else  {
-        $age = "$ageInMonths month"; 
+        $age = "$ageInMonths " . xl('month');
```
aethelwulffe: This is a numeric value, not the string "month".
Note that the whole getPatientAge function, is very primitive, and should look something like this:

function getPatientAge($dob)
{
$diff= date_diff(date_create($date), date_create($today));
$diff_months =$diff->format('%m')
if ($diff_months <=24)
{
 return "$diff_months" . " " . "xl('months')";
}
else
{
return "$diff->format('%y')" . " " . " xl('years')";
}
}

library/patient.inc
```diff
@@ -1819,6 +1822,31 @@ function getEffectiveInsurances($patient_id, $encdate) {
...
+function getAllinsurances($pid){
+   $insarr = array();
+   $sql = "SELECT a.type, a.provider, a.plan_name, a.policy_number, a.group_number,
+           a.subscriber_lname, a.subscriber_fname, a.subscriber_relationship, a.subscriber_employer,
+           b.name, c.line1, c.line2, c.city, c.state, c.zip
+           FROM `insurance_data` AS a
+           RIGHT JOIN insurance_companies AS b
+           ON a.provider = b.id
+           RIGHT JOIN addresses AS c
+           ON a.provider = c.foreign_id
```
aethelwulffe: You could also grab (this may be useful) the default x12 partner of the insurance company.

library/patient.inc
aethelwulffe: Should be "months" vs. "month" in any case...unless they are less than two months old...which is not a permanent situation.

teryhill: $age would be "15 month" with the word month being translated. I can change it to months

teryhill: This is getting the age in months not years.

aethelwulffe: I was just thinking you could handle both with one function.  In which case, the calling file would hardly need an external function if using date_diff.
Here is the best example of all general use value syntax I have found:
```<?php
$start  = date_create('1988-08-10');
$end 	= date_create(); // Current time and date
$diff  	= date_diff( $start, $end );

echo 'The difference is ';
echo  $diff->y . ' years, ';
echo  $diff->m . ' months, ';
echo  $diff->d . ' days, ';
echo  $diff->h . ' hours, ';
echo  $diff->i . ' minutes, ';
echo  $diff->s . ' seconds';
// Output: The difference is 28 years, 5 months, 19 days, 20 hours, 34 minutes, 36 seconds

echo 'The difference in days : ' . $diff->days;
// Output: The difference in days : 10398```

aethelwulffe:
Feb 4 this gets merged unless we are given notice otherwise.

aethelwulffe:
@2208Abhinav Thanks for digging in with a review!
Merging this...

PR #988, opened 2018-02-02
Visibility of PRINT button according to the selection of PAPER

2208Abhinav:
If in edit lists-> order types **Paper** is chosen or both **Paper** and **Electronic** are chosen then **PRINT** button will be shown.


aethelwulffe:
Does this PR have commits from all your branches in it on purpose?  Seems cool to me (only two files).

2208Abhinav:
I didn't monitored for globals.php and it got under the radar, should I remove that thing??

interface/forms/procedure_order/new.php
```diff
@@ -669,6 +678,40 @@ class="deleter"
                   format: "<?= $DateFormat; ?>"
               });
               $.datetimepicker.setLocale('<?= $DateLocale;?>');
+
+              var paper = <?php echo json_encode($paper);?>;
+              $('#procedure_type_names').on('blur', function(){
```
teryhill: Tell me what you are doing here? with this case statement.

aethelwulffe: Uh, gives you a print button if the index==1 in case "imaging"?

interface/globals.php
```diff
@@ -271,6 +271,7 @@ function undoMagicQuotes($array, $topLevel=true) {
       $gl_user[$iter]=$row;
     }
   }
+}
   // Set global parameters from the database globals table.
```
aethelwulffe: This bracket was already included in another merged PR, wasn't it?

aethelwulffe: I am not sure why another bracket would be needed there...
Looks like you have a branch Procedure_order, so I am not sure why you would have excess stuff unless this was a twig branch.

aethelwulffe:
Since the globals.php thing would break it, I do suggest you remove it from this PR, unless there is something major I am missing here...and I don't think I am.

interface/forms/procedure_order/new.php
2208Abhinav: @teryhill I extracted the data from the database all at once for paper being selected and linked it with selected procedure using case statement with the help of index number.I used the json_encode function to transfer PHP varaible values to JavaScript.

2208Abhinav: @aethelwulffe are we not supposed to have Print button for paper even if paper is selected??

2208Abhinav:
@aethelwulffe I removed the extra } from globals.php

interface/forms/procedure_order/new.php
teryhill: @2208Abhinav  You will need to read the variables. This is a list that can be changed by the user. It can be added to or have fields removed and names can be changed. Hard coding them will be a problem.

2208Abhinav: @teryhill I understand that you are saying not to depend on **index**, directly extract the value from the variable because the list can be changed and the code that I have written is not dynamic when its using the case statement to provide index.

2208Abhinav: I will have it solved by tomorrow positively :smiley:

teryhill:
@2208Abhinav I like where you are headed. I noticed this when the order type is changed the print option does not change until you click on it then it goes away.

interface/forms/procedure_order/new.php
teryhill: @2208Abhinav  You are doing great. I like your thought process.

teryhill: @2208Abhinav  are you able to test this on your system? Do you need some lab data ?

2208Abhinav:
Thank you @teryhill :smile:

interface/forms/procedure_order/new.php
2208Abhinav: @teryhill I will be glad if you provide the lab data for exact interpretation, right now I am working for the **index** problem.

aethelwulffe:
...we ALL need lab data...

2208Abhinav:
I removed the dependency on **index** making the code more self dependent.

teryhill:
[procedure_providers.zip](https://github.com/LibreHealthIO/LibreEHR/files/1691753/procedure_providers.zip)

@2208Abhinav  if you use phpMyadmin to import these two files it should give you some lab tests . I added imaging tests also.

2208Abhinav:
Did you test this new code?

teryhill:
not yet. but I will

teryhill:
It is switching the print button, Now you need to make it work when the order type is changed.

2208Abhinav:
I think if I will change **blur** event to **change** event then it will work fine.

teryhill:
I made that change and it is working fine.

2208Abhinav:
I will also implement that.

teryhill:
@aethelwulffe can this be merged?

2208Abhinav:
I will do it @tmccormi .

2208Abhinav:
I have pulled request #997 hope its not messy.

PR #991, opened 2018-02-02
Add pre_billing_issues report

aethelwulffe:
Adding report to menu.  date fields to be updated to html5, and overall
structure needs updated to LH standards.


aethelwulffe:
Code existed.  Tested on Demo server.  Just added menu item.

teryhill:
May need some opinions on the menu item name

aethelwulffe:
We got one.  this is my opinion!  :)

I kicked a few around.  Letting them know this is a QA test before they run billing is the idea.  "Check", "Issues" and all that is hard to pin down into a label that mostly deals with the patient records vs. actual system configuration issues.
  I am happy with anything...but the important thing is to make it look relevant and important enough that people will want to check it out.  Most reports go totally ignored.

Issue #990, opened 2018-02-02
Update ADODB


aethelwulffe:
Required for php 7.1 compatibility (I believe).  Other fixes are relevant to the same.

Perseus:
I've been using the site on php 7.1 for a bit and the only issue I ran into was the AdoDB one while installing. Are there any other known issues?

teryhill:
I think some of the other issues are warning that will have to be addressed as we find them. 

Did you upgrade to the latest AdoDB or just fix the error?

Perseus:
I just fixed the error manually by copying the current adodb xml schema code from their git repo. But updating the whole thing is obviously better. Should I create a PR with the updated AdoDB library?

teryhill:
Yes create the pr . update it on your system also so you can test it .

PR #992, opened 2018-02-03
PHP 7.1 Updates

Perseus:
Refers issue #990 

Started the process of updating dependencies for PHP 7.1. AdoDB is the first library that I found fatal errors with on PHP 7.1 in the installation process of LibreEhr


library/adodb/adodb-active-record.inc.php
```diff
@@ -1124,7 +1124,7 @@ function adodb_GetActiveRecordsClass(&$db, $class, $table,$whereOrderBy,$bindarr
 	// arrRef will be the structure that knows about our objects.
 	// It is an associative array.
 	// We will, however, return arr, preserving regular 0.. order so that
-	// obj[0] can be used by bootstrap developpers.
```
aethelwulffe: Ha-ha!  This was changed by someone that did a huge SED on the whole code base replacing "app" with "bootstrap".  Though I caught all those a long time ago.

library/adodb/.gitattributes
```diff
@@ -0,0 +1,17 @@
+# Default behavior
+* text=auto
```
aethelwulffe: .gitattributes file not needed in our repo... :)

aethelwulffe:
The following files have includes from adodb directories:

- [ ] library/sql.inc
- [ ] \contrib\util\menu\db_connection.php
- [ ] \gacl\gacl.class.php
- [ ] \gacl\setup.php
- [ ] \interface\reports\index.php

Would it be too much to ask to move adodb at this time over to /assets/adodb?

aethelwulffe:
BTW, in case you would like a good reason for moving this, I really like being able to set an exclusion for the /assets directory when doing searches across files... 😉

Perseus:
I'm on it!

interface/reports/index.php
```diff
@@ -3,7 +3,7 @@
 require_once (dirname(__FILE__) . "/../globals.php");
 require_once (dirname(__FILE__) . "/../../library/sql.inc");
 require_once (dirname(__FILE__) . "/../../library/Smarty.class.php");
-require_once (dirname(__FILE__) . "/../../library/adodb/adodb-pager.inc.php");
```
aethelwulffe: If you are calling globals.php, you can use the /library /assets and /modules global variables in the include names.  These are:

-  $GLOBALS['srcdir'], 
- $GLOBALS['assets'],
- $GLOBALS['modules_dir']

There is also $GLOBALS['standard_js_path'] for /assets/js standard third party javascript libraries.

Then you can call $GLOBALS['srcdir'] . 'headers.inc.php'  (or use $srcdir) to be able to call the functions in there like function call_required_libraries($library_array).
These functions may be expanded on to allow different parameters to be used as you require.
For instance, the adodb includes, and even sql.inc are not yet added to that header file.
  This is all very functional at the moment, but since you are doing it, bringing it absolutely up to the current consistency standards would be very nice.  It also isn't a bad time (especially as a series of commits) to do the code changes (as you have) then do the directory changes (as you have), then finally do any white-space adjustments within the files.

Finally, please please go ahead and add yourself to any attributions you deem appropriate.  It is good to be able to quickly see who "owns" the latest iterations.

aethelwulffe:
-This hits 8 tracking labels.  That is like the most ever.

interface/reports/index.php
Perseus: So the directory paths need to be changed to include the $GLOBALS variable, right? Is there anything else I need to do? Also what d'you mean by adding myself to any attributions? Didn't quite understand that.

aethelwulffe: It would be nice to do the directory changes, add common libraries to the headers functions etc...  Mostly, we are just waiting on someone else to review this...  I am approving it as-is now in hopes it attracts some attention.

aethelwulffe:
contrib/util/menu/db_connection.php has a conflict due to a merge that dropped the /contrib directory and moved the menu builder (which might be obsolete) to modules/util_menu/db_connection.php

Perseus:
I'll resolve the conflict and do the directory changes and push a commit asap

interface/reports/index.php
Perseus: @aethelwulffe Which part of the application does the reports/index.php file govern? I can't seem to figure it out to test whether changing the directories breaks anything.

Perseus:
@aethelwulffe  Which part of the application does the reports/index.php file govern? I can't seem to figure it out to test whether changing the directories breaks anything.

teryhill:
@Perseus is this ready? it still has the conflict. @aethelwulffe can you test this?

aethelwulffe:
@Perseus I am not sure if index.php is called by anything at all.  I will investigate.  Sorry I have been out of town for a while.
@teryhill I will start a round of testing later tonight.  If it checks out, Aniruhd will still have to take care of the conflict on his fork.

aethelwulffe:
Bump. Need some conflict resolution here.

Perseus:
Will get on it ASAP

On 21-Feb-2018 3:24 AM, Art Eaton <notifications@github.com> wrote:

Bump. Need some conflict resolution here.

Perseus:
@aethelwulffe If the reports/index.php file is not being referred by anything then is there a need to change the directory references to have the global src variables? Are there any other files which need the same to be updated?

aethelwulffe:
reports/index.php is called by the file interface/main/main_navigation.php
which is called by \interface\main\finder\patient_finder_keyboard.php
which has two commented out references, one being circular, which are in
\interface\patient_file\summary\browse.php:
and the circular one which is in \interface\main\main_navigation.php

So, my best evaluation is that the following files are trash:
interface/reports/index.php
interface/main/main_navigation.php
\interface\main\finder\patient_finder_keyboard.php

The remaining comment reference in \interface\patient_file\summary\browse.php line 32 should be removed as well.
I will create a PR that removes those files if you like.

Perseus:
I can do those updates on my end and commit those. The conflict should be fixed by me pulling all the commits that have been done into my branch, I believe.

aethelwulffe:
https://jeffkreeftmeijer.com/git-rebase/
REbase vs merge.  That is why some folks have like 19 commits in their PR for a one-line change.

Perseus:
Ah. I think I pulled the updates before I saw your message. I'm pushing a commit that has all those files deleted and should fix the conflict.

Perseus:
@aethelwulffe Fixed the conflict and those files. Did I miss anything that needs updating?

aethelwulffe:
Nothing jumps out at me.
@teryhill I think this is ready for a second review.  (Everyone else...you guys can pitch in and become the testing and review guru for LibreHealthEHR at any time.  I will gladly post glowing reviews on your linkedin site.)

aethelwulffe:
@teryhill can I get a review here?

teryhill:
Really over all not much changed looks like a lot of comments. Our programs are very few changes.

Issue #979, opened 2018-02-01
Contrib Directory is incorrectly named


tmccormi:
There are quite a few things (maybe all of them) that are listed as 'contrib' but are actually core parts of the project.

They need to be moved to a core area and dependent code changed to support that move.

We've already decided (a while back) that actual  non-core contributed code would be housed in a separate repo.

Perseus:
By moving to a "core" area, d'you mean create a new folder "core" and put those there? Also, how do I differentiate between the "contrib" parts and the "core" parts in the folder?

aethelwulffe:
First, "contribs" is and should be a separate repo.  This should contain optional code and modules...all optional code and modules.

Second, things that are in contribs that are actually USED or have any hard-coded references (and there are things that have such) then they need to be re-implemented (if any good) in /modules or in the actual core libraries or interface directories if they are actually vital non-optional components....or listed as a "core" module.

@Perseus You are looking at a code-base that is merely in the middle of a cleanup timeline.  This organizational refactor has been proceeding in a stranglevine approach over time.  To all appearances, the code base still looks like a trailer house that no-one has ever taken the trash out of, but the truth is we actually cut the code base by 50% before it started to grow a bit again.  

I think the essence of this issue is to start moving files that are actually called by active code into appropriate directories.  These could be /modules, /assets if they are third-party maintained, /data_sets or a subdirectory under /sql for things like the medical code lists, or even (possibly) fully integrated into /libraries and /interface if there is no-where else for them to go.

Currently, /libraries and /interface each contain both libraries and UI components, third-party assets, data loads, sql and everything else with no actual separation of code for specific features.
While breaking everything up into directories by feature obviously sounds great for troubleshooting, customizing, code-merges and the like, one of the huge benefits will be when we can look at the directory structure and be able to actually _identify and name_ off a list of features and components that comprise LibrehealthEHR.  Right now, we have _no idea_ what all features exist.  We can't begin to do a gap analysis, identify legacy stuff...none of that.

Let's kill /contribs ASAP.  It is a good next move.

aethelwulffe:
TODO:

Fully modularize the Translations code, the medical code external dataloads system.
Evaluate the utility of the menu creation script.
Evaluate all the util_* named directories located in #993 for refactor, enabling, deleting, or banishing to the Contribs repo.

PR #993, opened 2018-02-05
Contribs rename

aethelwulffe:
Ref #979 
Moved all code packages to
modules/

Corrected urls.

Corrected documentation / comment references where practical.

Renamed directories with utility packages that are not currently called by anything directly (including the menu builder tool and Perl utilities) to util_* as a identification standard.

Suggest moving the util_* features to the Contribs repository.

Note:  The log recovery Perl tool does not work, and is not IAW current scheme, and there is a better way to do that now.  I removed it.

the "dupecheck" feature is not called by anything, nor is it in the menu, and I am not sure if it works.  Same for the de-identification script.  These may require refactoring, removal, or relocation to the contribs repo.


interface/fax/fax_dispatch.php
```diff
@@ -200,8 +200,7 @@ function mergeTiffs() {
       }
 
       if (!$info_msg) {
-        // The following is cloned from contrib/forms/scanned_notes/new.php:
-        //
+// scanned notes must be installed, and does not natively exist.
```
aethelwulffe: Without re-creating the scanned forms code, this does not exist and is a bad reference.  The fax tool here is probably legacy poop anyway.

tmccormi: The version that bowen has is more generic and useful.  Perhaps we should pull that in to another P/R

aethelwulffe:
FAX code?

tmccormi:
Yes interface/fax   I modified it to be used for a directory tree of incoming scanned documents (PDF) and dropped all the TIFF stuff.

Needs cleanup and maybe renaming to scan* instead of fax ..  I can push a branch with it.

Will merge this in the meantime

PR #994, opened 2018-02-07
Adjustment code report

teryhill:
This report gives a total for the selected adjustment code over a given time period. #443

### DIFF
--- a/interface/reports/adjustments.php
+++ b/interface/reports/adjustments.php
@@ -47,9 +47,9 @@ else {
 <head>
 <title><?php echo xlt('Adjustments'); ?></title>
 
-<script type="text/javascript" src="../../library/overlib_mini.js"></script>
-<script type="text/javascript" src="../../library/textformat.js"></script>
-<script type="text/javascript" src="../../library/js/jquery-1.9.1.min.js"></script>
+<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/overlib_mini.js"></script>
+<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/textformat.js"></script>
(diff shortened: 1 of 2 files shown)


aethelwulffe:
I'll get reviewing again tomorrow.  Flu got me down hard.

sites/default/menu_data.json
```diff
@@ -988,41 +994,41 @@
         "global_req" : "addr_label_type"
     } ],
     "requirement" : 0
-}, {
+}, { 
     "label" : "QA Measures",
     "menu_id" : "mips",
-    "children" : [ {
+    "children" : [ {
```
aethelwulffe: How did the extra white pace come in?  Was that from your editor, or was it generated by the menu editor?  If it was from the menu editor, we should probably leave that as a standard.

interface/reports/adjustments.php
```diff
@@ -0,0 +1,264 @@
...
+
+<body class="body_top">
+
+<div id="overDiv" style="position:absolute; visibility:hidden; z-index:1000;"></div>
+
+<span class='title'><?php echo xlt('Report'); ?> - <?php echo xlt('Adjustments'); ?></span>
+
+<div id="report_parameters_daterange">
+    <?php date("d F Y", strtotime(oeFormatDateForPrintReport($_POST['form_from_date'])))
+    . " &nbsp; to &nbsp; ". date("d F Y", strtotime(oeFormatDateForPrintReport($_POST['form_to_date']))); ?>
```
aethelwulffe: Does this date format function allow for localization?  If not, then Y m d is preferred.  If using Portuguese as the language doesn't result in "Januario" then we are just making trouble.

teryhill: It is the same date function in all the other reports so I hope so 8^)

teryhill:
Pictures of what it looks like.

image

image

aethelwulffe:
OK,  Works.

PR #996, opened 2018-02-09
Dupecheck incorrect column name for DOB

aethelwulffe:
Query error due to DOB column being called dob.

dob>DOB for everything in file for full consistency


aethelwulffe:
Work in progress.
Tool also needs special character scrubbing where the name has an apostrophe in it.  "O'Hare" will break the code.  That is actually what prompted me looking at this.

aethelwulffe:
Next issue:  "limit results" seems to not work as expected.  The default 100 setting returned four results in a test data set known to have thousands.  Increasing that value to 1000 gave 59 results.  10,000 gave 540 results.  1,000,000 gave 754 results.
-Something ain't right there.

aethelwulffe:
This has been approved by one person for over a month.  I guess it can be merged by me.

PR #997, opened 2018-02-12
Visibility of PRINT button according to selection of PAPER

2208Abhinav:
Print button will be shown if paper is selected previously.


teryhill:
@tmccormi will this merge now?

aethelwulffe:
I think I have tested this...meaning that if I am correct on what I am testing, it looks like it works.  This should be squashable.

aethelwulffe:
I am not that great on reviewing the JS parts of this.

PR #1000, opened 2018-02-15
favicon has been added to login screen

naveen17797:
There is no libre favicon on both main and login page, both have been added in this patch


naveen17797:
sorry for the 2 commits, it was a mistake.

pri2si17-1997:
Looks good... :slightly_smiling_face:

PR #1006, opened 2018-02-16
DatePicker in add patients section can add future dates.

naveen17797:
before fix
screenshot from 2018-02-16 09-12-05
after fix
screenshot from 2018-02-16 11-57-32

**# Even if the user adds future date, it will be converted to current date.**

### DIFF
--- a/library/patient.inc
+++ b/library/patient.inc
@@ -931,6 +931,7 @@ function newPatientData(    $db_id="",
             )
 {
     $DOB = fixDate($DOB);
+
     $regdate = fixDate($regdate);
 
     $fitness = 0;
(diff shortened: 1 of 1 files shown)


teryhill:
yes server side

aethelwulffe:
$(document).ready(function(){
    $('#myDate').attr('min', todayDate());
});

teryhill:
@aethelwulffe does that stop you from manually entering a date in the future. The current set up of the date picker will stop the future date but not if it is entered manually.

naveen17797:
yes it is not validated in server side.

aethelwulffe:
Well, our test patients "Martha Washington" and George Washington will be be harder to replicate with a 150 year age limit, but this looks good.  Getting it into my test line-up now.

naveen17797:
See also pull #1012 its more important than this one.

teryhill:
@aethelwulffe this is a duplicate pr @naveen17797 will be deleting it

aethelwulffe:
I think I realized that just a little while ago...

Issue #971, opened 2018-01-31
Login and run a specific program


teryhill:
I would like to see the option to create a login "flowboard" and when that user is entered the Flow Board runs in full screen and only the flow board. This will need changes to the login screen to check for the flowboard login and perhaps changes to the flowboard to run in full screen.

apooravc:
@teryhill I didn't get what you meant by a "flowboard". Would you please elaborate a bit?

prabhakarpd7284:
Is it fine if it looks like this after login

![flow_board_full](https://user-images.githubusercontent.com/23030978/36340123-132787f8-13fc-11e8-952e-c8358b4b2899.png)

apooravc:
This looks good. But what about the navigation bar? Where should that be put @teryhill ? Also, I think the "Status" drop down should be aligned properly, with this issue.

prabhakarpd7284:
This preview looks better.

![flow_board_full_](https://user-images.githubusercontent.com/23030978/36341563-189a7136-1416-11e8-960c-47b5105f0ece.png)

aethelwulffe:
I will check out that PR.  It looks like a useful option.

aethelwulffe:
I like the PR.  I think that we need a toggle in globals to enable the feature, or maybe even a fine control in the user or user pref settings for "specific_application" or "single_application" or something.
  
The goal of such would be to allow a clinic to decide if they want this option on the login page, then allow the option to be a hard-set access control restriction on a user with or without the checkbox appearing on the login page.

PR #1010, opened 2018-02-17
Add option of FlowBoard login only in full screen

prabhakarpd7284:
Solve #971 

- Add FlowBoard Login check in login page.
login

- When box is checked, only FlowBoard opens in full screen.

flb

### DIFF
--- a/interface/login/login.php
+++ b/interface/login/login.php
@@ -190,7 +190,7 @@ include_once("$srcdir/sql.inc");
                                     <tr>
                                         <td><span class="text"><?php echo xlt('Language'); ?></span></td>
                                         <td>
-                                            <select class="entryfield" name=languageChoice size="1" style = "height : 28px">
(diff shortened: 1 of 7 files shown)


aethelwulffe:
@prabhakarpd7284 
1.  We should toggle this option with a global (enable_flowboard_only or something).
2.  This might be something that could/should be toggled within user prefs or within the user config.

prabhakarpd7284:
@aethelwulffe 
This seems like a fine approach.
I will work on your suggestions.

aethelwulffe:
Thank you @prabhakarpd7284 .  You have a functional piece here.  This little bit of polish turns it into a significant feature.  The push to enable a "kiosk" or special terminal access port is really helped by this.  I encourage you to look at this as an opportunity to put hooks in so that any number of single-function-only settings could be selected.  If we have the methods formalized, we can start getting away from GACL, and replacing it with something coherent and native.

teryhill:
replaced by pr 1043 Duplicate

PR #1012, opened 2018-02-18
server side validation on the menu>user preferences

naveen17797:
server sided validation has been done on menu > user preferences
@teryhill review this.


library/patient.inc
```diff
@@ -1057,6 +1058,62 @@ function pdValueOrNull($key, $value) {
     return "'$value'";
   }
 }
+// Supported input date formats are:
+//   yyyy/mm/dd
+//This function is to prevent user from entering the future date
+function fixDOB($date) {
```
pri2si17-1997: This is good. But I would suggest to make one more change. Do not allow to get past date of lets say beyond 100 yrs(we can sort out a suitable date range). Like in present case, any date like : "1001-01-01" is accepted. So this should be prevented.

naveen17797:
Also i forgot to write for other date formats, need to do it and make the
PR ready for merge.

naveen17797:
@pri2si17-1997 fixed, set the maximum life expectancy to 150 years

teryhill:
@aethelwulffe bump

interface/super/edit_globals.php
```diff
@@ -126,9 +126,30 @@ function checkBackgroundServices(){
...
+          $search = $grparr[$fldid][1];
+          $boolean = array_key_exists(trim(strip_escape_custom($_POST["form_$i"])),$search);
+          }
+          elseif ($grparr[$fldid][1] == bool) {
+          $_POST["form_$i"] == 0;
+          $boolean = true;
+          }
+          elseif ($fldid == "css_header") {
+            //styles array created since the globals does not have styles array.
+            $styles_array = array("style_prism.css", "style_light.css", "style_purple.css", "style_tan.css", "style_tan_no_icons.css");
```
aethelwulffe: Can't have anyone inserting their own CSS here now... :)  TODO, figure out where the enumeration should come from.  Probably against existing database values, vs hard-coding.

aethelwulffe:
Test verifies that Tut Ank Amun does not have a valid birth date.
Merging.

Issue #986, opened 2018-02-02
Tags, Patient/Tags, Filters UI's are badly formatted.


aethelwulffe:
These features have not had any attention lately in the current layout scheme.

First thing I noticed was that the search field was getting eclipsed.  Then I noticed that the field content was overlapping and doing other strange things.

steps to replicate:
Administration/Globals
Features
Enable Demographics Tag Filtering Feature

Create a tag.
Create a filter rule that does something like restrict user access to a patient with the tag you created.
Add the tag in patient demographics.

-In this process you should see all the locations that have bad CSS or tables.
![image](https://user-images.githubusercontent.com/553751/35710189-172093a6-0784-11e8-9ddb-a79894c1f408.png)

Trodrige:
@aethelwulffe I cant' do this: 'Enable Demographics Tag Filtering Feature'

aethelwulffe:
Apparently, I was looking at a dev instance that had that toggle added.  On the demo the tags system is hard-enabled.  Just look under Administration/Tags.  The feature is there.

plant99:
@aethelwulffe  As @Trodrige  has pointed out, even I couldn't find how to 'Enable Demographics Tag Filtering Feature' but regardless, I went as you said and I found the same bugs on resizing window. Is anyone working on this? If not, can you please assign me this issue?

plant99:
@aethelwulffe  I followed what the bugs are. Can you please assign me here to fix this issue?

teryhill:
@plant99  Let me check with @Trodrige If he does not want it you can have it

teryhill:
@plant99 @Trodrige wants to do it

teryhill:
@plant99 are you on the chat here https://chat.librehealth.io/channel/gsoc

teryhill:
@aethelwulffe close this?

tmccormi:
This seems to related to other changes in the way we use bootstrap etc ...

aethelwulffe:
There is plenty of issue and dev related to this going on right now, and I am pretty sure it was addressed.

PR #1013, opened 2018-02-18
Fixing tags, patient-tags, and filters ui format

Trodrige:
Fixes issue #986


aethelwulffe:
Direct file swap test is fine.  I did the change, then had to take a nap before my brain would remember what I was doing!

PR #1016, opened 2018-02-18
Fix #1011: Create an attached Document validation tool

### LINKED ISSUE #1011: Create a attached Document validation tool
Administration tool that checks each document in 'document' table to see that it actually exists on disk Or in couchdb, if used) and the hash still matches.

Produces a browsable report.

apooravc:
Fixes #1011 

Added tool as another option `docs_validate` inside Misc. on main-menu. This tool checks for each record in `documents` table for whether the file in record actually exists on disk or CouchDB and a report is generated. I've used `file_exists()` for checking in disk but need help for checking same in CouchDB. I've adapted UI from `addressbook_list` layout.

Preview:
When some records are found in database:
img1

When no record is found in database:
img2

2) After cloning, only the master branch would be available locally. Enter `git checkout -b add-validation-tool origin/add-validation-tool` or `git branch add-validation-tool origin/add-validation-tool` to get the branch _add-validation-tool_ locally.
3) The changes could then be seen on main page.

@tmccormi  @teryhill @pri2si17-1997 Please have a look.


apooravc:
@aethelwulffe I added 500 as a limit because of an error which was telling me to add such a limit but I was thinking of showing 15-20 results, page-by-page and add a "Next" & "Prev" buttons for that.

interface/usergroup/docs_validation.php
```diff
@@ -0,0 +1,87 @@
...
+      <td><?php echo xlt('Storage'); ?></td>
+    </tr>
+
+  <?php
+    $listEmpty = true;
+    while ($row = sqlFetchArray($res)) {
+      $listEmpty = false;
+      $fileURL = $row['url'];
+      $location = substr($fileURL, strpos($fileURL, "LibreEHR"));
+      $size = ((float)$row['size'])/(10**6) . " MB";
```
teryhill: Is the ** here correct?

aethelwulffe: Does that equate to MB?  Ten to the 6th power.  MiB is 1024*1024.  Never seen that style of notation/calculation before.  I guess we could like... _test the code_ to validate!

teryhill: I got a php error on it that is why I asked.

aethelwulffe: Good reason to ask.
The answer is "no" that doesn't work :)

aethelwulffe: Is it the math, or a division by zero issue?  Could be either if the row does not return anything.  Good idea to check if a value exists before dividing.  Also, formatting to megabytes isn't the best option for file size checks that should be <1MB

apooravc: Silly mistake by me! I will replace (10**6) with (1024*1024), which seems much better.

aethelwulffe:
Well, _technically_ 10 to the 6th is MB, and 1024squared is MiB...but either way.  You could just divide by 1000000.  Or multiply by .0000001

apooravc:
@aethelwulffe Did some changes:

- Replaced (**) operator.

- Added a check for size < 1 MB.

- Added `Documents validation` under a sub-menu `Utilities` in Misc. menu.

@tmccormi Yet to add a check for CouchDB. My question: When we upload a file using `Misc > New Documents`, there is an option to "Choose files" and it opens a window to select a document from disc and creates an entry in DB (phpMyAdmin) with attributes such as `url` which can be used to validate a file stored on disc. But how to do this with CouchDB? There is no option to ask user to upload a file from CouchDB. How to create an entry in DB with `couch_docid` and `couch_revid` being not null? Validation can then be done with a `HEAD` request to CouchDB with respective parameters.

aethelwulffe:
I just fixed a conflict that popped up and pushed it to this.  I think it had to do with removing the Refer menu item.
  Is this still a work in progress?  If so, let us know what you would like help with.

apooravc:
Yes, please, could you check my previous comment, about CouchDB validation?

teryhill:
@tmccormi this looks good to me , is this what you were wanting?

aethelwulffe:
@apooravc Sorry for lack of advice in the couchDB implementation.  I have never attempted to use it myself.  The fact is, if it is in the database as a blob, then the file exists, and that is it.  There should be no orphans if it is all stored in the database.
  Like I said though, I know next to nothing about the couchDB implementation.  I always feared that if a table became corrupt or crashed, we would lose all the documents.

interface/usergroup/docs_validation.php
```diff
@@ -0,0 +1,87 @@
+<?php
```
teryhill: This needs a documentation header then I think it can be merged. @apooravc look at a file like the style.php the naveen did in the color picker pr

tmccormi:
@apooravc, I am in a similar boat regarding CouchDB, it's a feature I've never used, but it is rumored to work if you install couchdb and turn it on in globals.   I'd be happy to spit this into two issues.

tmccormi:
I would also entertain removing CouchDB support entirely ... :-)

aethelwulffe:
Simplification by removal of unused stuff is a great thing.  Cyberhoarding sucks.

aethelwulffe:
I have not run a real test against it at all myself.  This is going to take some set-up time, and I would rather do that setup by finishing the document additions to the test data generator.

apooravc:
@tmccormi I'm thinking of creating another issue for CouchDB validation if it's needed. As of now, this only checks if a document in document table actually exists in the specified directory on disk. It doesn't check for corrupted documents.

aethelwulffe:
OK, finally hitting test lists.  This one was HARD to set up a test for.  Sorry it took so long.

PR #1021, opened 2018-02-20
validation for main globals added.

naveen17797:
validation for main globals has been added,
it is used to validate the administration > globals


naveen17797:
i have tested almost all fileds before making this commit, hope these changes wont give rise to new issue 😁

PR #1027, opened 2018-02-21
text fields, menu layouts are not saved after validation PR

naveen17797:
text fields and menu layouts which are not saved after validating the globals, patch has been made.


teryhill:
@pri2si17-1997 do you have any Idea why his commits look like this?

interface/super/edit_globals.php
```diff
@@ -269,6 +271,16 @@ function checkBackgroundServices(){
               $boolean = false;
             }
           }
+          elseif ($fldid == "theme_tabs_layout") {
+            //menu array created sine the globals dont have the array
```
aethelwulffe: *Since*

interface/super/edit_globals.php
```diff
@@ -269,6 +271,16 @@ function checkBackgroundServices(){
...
+            $menu_array = array('tabs_style_compact.css', 'tabs_style_full.css');
+            if (in_array($_POST["form_$i"], $menu_array)) {
```
aethelwulffe: Suggestion:
*expression* ? true : false;

pri2si17-1997:
@teryhill maybe he is not doing in one branch and then merging them to a single branch and raising PR. :thinking:

PR #1030, opened 2018-02-22
colour picker has been added (experimental)

naveen17797:
colour picker has been made in this branch as experimental feature, since themes needs to be refactored a lot,this cant be merged until we update themes.
1.theme setter has been added
screenshot from 2018-02-22 16-39-18


naveen17797:
screenshot from 2018-02-22 17-26-13
much better look

aethelwulffe:
Hi Naveen,
This is pretty darn cool.  Should we add the "experimental" tag to this for now?

We need to figure out how to rebase your fork master branch, so that your dev branches don't have every commit in the history since it's creation.  Something isn't quite right here.

naveen17797:
I have added experimental tag since i dont know about the ui in most of the pages,
Yes, the old commit messages are being displayed when i make a PR, i will look in to it.

library/headers.inc.php
```diff
@@ -19,6 +19,21 @@
  */
 ?>
 <link rel="stylesheet" href="<?php echo $css_header;?>" type="text/css">
+<style type="text/css">
```
aethelwulffe: I am not sure where/when a stylesheet reference crept into this file, but the intention of this file is just a library of asset function calls.  Everything in this file should be a function that calls includes.  Nothing else.
$css_header variable should contain the file location for holding this stuff (I think).

naveen17797: We need to make .css files in to .php files to use php variables inside them, so that we can use the colors we choose

naveen17797: By adding this feature i didnt want to break anything inside the ui, so i separated the bg classes and echoed out with our php variables

aethelwulffe: I get why you did it.  I am just hoping you can find a better place to put it than this particular file...unless it turns out that it is really really really the best place, in which case it should just have a function to call the code file, not the code.

naveen17797: Im planning to remove the <link rel= "stylesheet" and replace it with style.php file, i need to know are we ready to remove themes and adopt the theme setter.

aethelwulffe: Yes...provided we have things that look good, and we have padding control too, then we are set to move forward.  We just can't have the project go into an unusable state during a transition.

naveen17797: I will continue it to make a simple single theme with customizable options and make it ready for merge as soon as possible.

aethelwulffe: Sounds great.  Sending you a collaborator invite.

naveen17797:
@pri2si17-1997 the color picker varies for different browsers, so the alternative may be creating color picker on js.

pri2si17-1997:
http://www.eyecon.ro/colorpicker/

See if it can be used.

aethelwulffe:
I highly recommend using the browser color picker.  It isn't just a matter of application responsiveness (though everything helps) but also a package maintenance thing.  People can get over the difference in a color picker image in the documentation...which they almost certainly would not read anyway.

Is there an actual application code-side issue with using native color pickers via HTML5?
*Yes, I am an HTML5 advocate.

naveen17797:
@pri2si17-1997 as @aethelwulffe  said, i think we could use default browser color picker for picking colors instead of using js for the color picking.

pri2si17-1997:
Ok. I was just suggesting. :slightly_smiling_face:  We can use browser color picker.

naveen17797:
i had the same idea as you said before, but i feel there wont be much difference except the fact that we could control that color picker, in case of browser color picker we could not

naveen17797:
see #1044

PR #1031, opened 2018-02-22
Fix #414: 'Back to Patient' button on screens linked from demographics

### LINKED ISSUE #414: 'Back to Patient' button on screens linked from demographics
Hi Folks-

What do you think about adding that 'Back to Patient' button to the screens linked to at the top of the summary screen?

History | Report | Documents | Transactions | Issues | Ledger

'History' has 'Back to Patient' and I see Issues has 'Back'.  I think that's a very worthwhile button to have, and it'd be good UI continuity to put it on the others.  And make that 'Back' -> 'Back to Patient'...

lehr-hx-btns

apooravc:
Fixes #414 
Added "Back To Patient" button on Report | Transactions | Issues | Ledger, on demographics screen for patient.

Preview:
Report
report

Transactions
transactions

Issues
issues

Ledger
ledger

Questions:

- Transaction screen has a "cross"/close button and it opens as a window over demographics screen. I think there is no need to add button here or if added, it should close the transaction window? 
Preview:
transactions2

- I didn't add the button to Documents because there is also another way to go to Documents screen, the 'Misc. > New Documents' route. So should there be something which checks which route user took and then add the button?

2) After cloning, only the master branch would be available locally. Enter `git checkout -b add-back-button origin/add-back-button` or `git branch add-back-button origin/add-back-button` to get the branch _add-back-button_ locally.
3) The changes could then be seen on respective screens.

### DIFF
--- a/interface/patient_file/report/patient_report.php
+++ b/interface/patient_file/report/patient_report.php
@@ -2,29 +2,29 @@
   /*
    * Patient report
    *
-   * Copyright (C) 2016-2017 Terry Hill <teryhill@librehealth.io> 
+   * Copyright (C) 2016-2017 Terry Hill <teryhill@librehealth.io>
    * No other information in the previous header
    *
-   * LICENSE: This program is free software; you can redistribute it and/or 
-   * modify it under the terms of the GNU General Public License 
-   * as published by the Free Software Foundation; either version 3 
-   * of the License, or (at your option) any later version. 
-   * This program is distributed in the hope that it will be useful, 
-   * but WITHOUT ANY WARRANTY; without even the implied warranty of 
-   * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the 
-   * GNU General Public License for more details. 
-   * You should have received a copy of the GNU General Public License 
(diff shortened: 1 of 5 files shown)


aethelwulffe:
Funny how History isn't part of demographics, or listed under demographics, but is rather a separate sort of item.
Under issues, you can access surgery and dental issues, but everything else (allergies etc...) already has it's own widget.  

For documents, pass a $from_page value to the URL.  As a matter of fact, we can do that in a lot of places.

apooravc:
What about "Transaction" widget? Does it also need the back button?

aethelwulffe:
Transaction window can just be closed.  It is fine, as the background iframe never goes away.

apooravc:
Removed button from Transaction screen and added button to Documents (thanks to @teryhill ).
Used `(int)$GLOBALS["_GET"]["patient_id"]` to figure out when the user navigates to Documents screen via route `Misc. > New Documents`, which would give `0` value.

aethelwulffe:
I'll test after things calm down here!

apooravc:
@teryhill Could you also please review this to check if it can be merged?

teryhill:
@aethelwulffe  bump

PR #1032, opened 2018-02-22
Not sure how these got missed

teryhill:
I thought they were pushed in the commit for this, but it looks like
they were missed.


teryhill:
I am pretty sure. I could use that development tip demo. There is a time lag between committing and re testing against my system.

aethelwulffe:
Well, we are not getting a lot of review help these days....so I am going to merge this.  It is your baby, and you are in charge of it anyway.

aethelwulffe:
I modified the commit statement so it...meant something :)

PR #953, opened 2018-01-16
Update README.md and INSTALL.md

Kachachan:
Followup on PR #899

PR #1033, opened 2018-02-23
Minor README update

apooravc:
Did some minor changes in the README.md file. Some of the changes were done so as to make a newcomer more accessible to the development of the application. Other changes were minor re-formatting. I was also thinking of adding an image of login page but that may be an extra thing to do for a README.

@teryhill Please have a look.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,40 +1,29 @@
-# Welcome to LibreHealth EHR!!!
+# LibreHealth EHR
 
-LibreHealth EHR is a free and open-source electronic health records and
-medical practice management application. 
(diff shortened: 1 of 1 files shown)


aethelwulffe:
Check out PR #953 for duplication of effort.  That PR has been waiting a long time.  Just do a check against it, and see what we have.  Likely we cannot integrate both.

apooravc:
@aethelwulffe I saw the changes done in that PR and I'm not sure but I think the README part matches with the current live one and so if that PR is accepted, I think there should be no problem in stacking my changes upon it. Or I will close this PR if there happens to be any conflict.

PR #1035, opened 2018-02-26
validation for file manager (interface/super/manage_site_files.php)

naveen17797:
validation for file manager has been added with mime checks


naveen17797:
closed, see #1036

PR #1036, opened 2018-02-27
validation for files upload and multiple file uploads enabled

naveen17797:
1. validation for image files
2. validation fix for pdf files
3. multiple file upload option enabled


aethelwulffe:
Naveen, this feature might be a dead one.
Would you please check the code in /modules/filemanager (almost identical to sites/default/filemanager) and see what you think about improving these as a general file manager in the system?  Currently, they are two blended and sandboxed tools built out of sample code.  The one under /sites exists to enable the ability to upload .sql files, and retrieve .xml report exports from the feature in /modules/MIPS.  That feature is enabled by the "demo" flag in Globals/MIPS.  The file manager copy in /modules/filemanager is not pointed at by the .json menu by default.  The hope has been that a tool like this, perhaps using the same .js library could replace all the file upload/download features for billing files, patient documents, as well as exported report copies.

naveen17797:
i had my opensource project called newway, https://github.com/naveen17797/newway-filemanage, it is written in php and jquery.
will it be useful?
below is the screenshot of my open source project
screenshot from 2018-02-28 20-55-15
screenshot from 2018-02-28 20-54-58

interface/super/manage_site_files.php
```diff
@@ -211,5 +228,4 @@ function msfFileChanged() {
 
 </form>
 </body>
-</html>
-
+</html>
```
aethelwulffe: Newline at end of file please.

aethelwulffe:
@naveen17797 You wanna look at using your project to replace the filemanger in /modules and in the sites directory?  Currently used with /modules/MIPS?

PR #1042, opened 2018-02-28
calendar should not accept negative values

naveen17797:
slight change in the calendar validation


aethelwulffe:
Someone MIGHT want to back-date an appointment to make sure it was on the schedule.  There could be times when someone loses system access, shifts to another method, but wants to record everything in the EHR.  I would suggest that the validation generate a warning, vs. a hard fail.

teryhill:
Back dating appointments are needed

naveen17797:
FYI the validation is for selection box consists of the calendar starting and ending hour, i think it does not interfere with backdating appointment.

Issue #961, opened 2018-01-20
Warning when generating report - from patient screen


alpinus4:
# Steps to reproduce: 
**1.** Enable displaying errors in _php.ini_

**2.** Go to patient screen and click _Report_

![](https://user-images.githubusercontent.com/25907897/35188011-4d7d96ba-fe2e-11e7-81ba-073f1d283ffe.jpg)

**3.** Click the second _Generate report_ button

![](https://user-images.githubusercontent.com/25907897/35188038-b5a582de-fe2e-11e7-9f09-0a6532a7784b.jpg)

**4.** See error

![](https://user-images.githubusercontent.com/25907897/35188041-c5168592-fe2e-11e7-8fbf-23f437154493.jpg)

Browser: Firefox 57.0.4
OS: Windows 10 ver. 1709

teryhill:
Not getting the error on my system or the demo system, Do you have popups turned on?

alpinus4:
Popups? I haven't enabled anything in _Administration_

teryhill:
Enable popups in the browser

alpinus4:
I have enabled popups on Chrome and Firefox. And on both throws error.

teryhill:
Those are warnings. They are caused because the generatedXml directory is missing create it and the warnings will go away.

LibreEHR\ccr\generatedXml

alpinus4:
Yea

PR #1046, opened 2018-03-02
fix for issue reported in GCI

teryhill:
fix for issue reported in GCI This was giving some console warnings. #961


teryhill:
@aethelwulffe there you go new line and an extension.

PR #1049, opened 2018-03-03
UI Improvement : messages>Add new > patient

naveen17797:
screenshot from 2018-03-03 20-53-00
- previously to find patients you need to click search button
- But now when you  type, patient list will appear.
- used  key up event with some modifications to improve ui

### DIFF
--- a/modules/calendar/find_patient_popup.php
+++ b/modules/calendar/find_patient_popup.php
@@ -164,7 +164,7 @@ form {
    </select>
  <?php echo htmlspecialchars( xl('for:'), ENT_NOQUOTES); ?>
    <input type='text' id='searchparm' name='searchparm' size='12' value='<?php echo htmlspecialchars( $_REQUEST['searchparm'], ENT_QUOTES); ?>'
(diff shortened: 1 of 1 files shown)


aethelwulffe:
Very cool.  How does it work when there is a tag access restriction?
That is what I will be testing.

teryhill:
I like the change , Is there any reason to have the search button? Let us know how it tests with the tags stuff @aethelwulffe

aethelwulffe:
I am waiting on Marcus for something, then I will try testing it with some tags in place.

aethelwulffe:
I think that if the search button is useless that is fine.  If the typeahead is looking at the right field when you are doing it for whatever your are searching for, that is better.

aethelwulffe:
Which apparently, it does.  Remaining question is to attend to the (unneeded?) submit button.  If this feature were on a toggle, it would still be needed, but otherwise, not so much.
  I think it should also have a "full name" last name" "first name" search criteria option.

naveen17797:
the pr just changes the way how the form was submitted, before it was done using submit button, now it is done using keyup event. it doesnt affect form contents such as full name, lastname.The function will be carried as default @aethelwulffe @teryhill

PR #1051, opened 2018-03-05
Autocomplete function has been added to patients>add new

naveen17797:
screenshot from 2018-03-05 20-39-49
- in hospitals this type of autocomplete function is necessary, so that there is no need to type the same name again and again
- it will help to add patients quickly and reduce the time spent in typing

### DIFF
--- a/interface/new/new.php
+++ b/interface/new/new.php
@@ -117,7 +117,7 @@ while ($orow = sqlFetchArray($ores)) {
    <span class='bold'><?php xl('First Name','e');?>: </span>
   </td>
   <td>
-   <input type='entry' size='15' name='fname' value='<?php echo $form_fname; ?>'>
+   <input type='entry' size='15' name='fname' value='<?php echo $form_fname; ?>' id="fname">
   </td>
  </tr>
 
@@ -126,7 +126,7 @@ while ($orow = sqlFetchArray($ores)) {
    <span class='bold'><?php xl('Middle Name','e');?>: </span>
   </td>
   <td>
-   <input type='entry' size='15' name='mname' value='<?php echo $form_mname; ?>'>
+   <input type='entry' size='15' name='mname' value='<?php echo $form_mname; ?>' id="mname">
   </td>
  </tr>
 
@@ -135,7 +135,7 @@ while ($orow = sqlFetchArray($ores)) {
    <span class='bold'><?php xl('Last Name','e');?>: </span>
   </td>
   <td>
-   <input type='entry' size='15' name='lname' value='<?php echo $form_lname; ?>'>
+   <input type='entry' size='15' name='lname' value='<?php echo $form_lname; ?>' id="lname">
   </td>
  </tr>
 
--- a/interface/new/new_comprehensive.php
+++ b/interface/new/new_comprehensive.php
@@ -70,7 +70,7 @@
       //  Include Bootstrap, Fancybox, date-time-picker
   call_required_libraries(array("jquery-min-3-1-1","bootstrap","datepicker","fancybox-addpatient"));
 ?>
-
+<link rel="stylesheet" type="text/css" href="../../assets/css/jquery-ui-1-12-1/jquery-ui.css">
 <style>
 body, td, input, select, textarea {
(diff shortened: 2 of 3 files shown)


naveen17797:
currently in this PR it is enabled only to the fname, mname, lname.
If this approach is accepted then we could extend it to fields wherever it is needed.

naveen17797:
@aethelwulffe  @teryhill @pri2si17-1997  please review this

aethelwulffe:
Heh, this is cool Naveen, but also scary in a way.  With all the messed up spellings out there...uh...
How about we put in a toggle to enable this for user settings (accessible right on the demographics screen).  Personally, I think most folks would rather finish typing at their normal 120 words per minute than stop and pick an option.

aethelwulffe:
I notice TWO copies of Naveen showed up....

naveen17797:
I have added my name 2 times in DB, that's why it showed up, need to fix it

aethelwulffe:
What happens if you have 50 patients named "Michael" or "Muhammed"?  Does it list all of them?  I suggest we test this against a very very large database.

naveen17797:
Yes, i need to fix that, if a similar name is present then we can show it only one time .

naveen17797:
screenshot from 2018-03-06 17-17-29
fixed

naveen17797:
@aethelwulffe @teryhill  i would like to know where this feature is most useful in this application, i have just explored only primary areas of it , i hope you know where it will help to reduce time

aethelwulffe:
Hi @naveen17797 ,
Could you put in a "use search enhancements" feature toggle into this?

naveen17797:
@aethelwulffe like making a user to choose whether they need search enhancements or not?

aethelwulffe:
I would put it in globals and in user preferences, not in the user admin/creation.  I think we need to move more things out of the user creation modal window, and into user preferences.  Additionally, there is the whole "address book" part of the user table.  External users are stored in the users table as contacts.  Many of the user table fields are only accessible through the address book.  Many of these fields are very useful, and most should be added to the user preferences settings.  Admin view of these should still be available somewhere, though.  For now, it is the ancient clunky address book.

aethelwulffe:
Caveat:
Currently the only way to enable granular access control over what users see what patients is the TAGS system.  The ACL by Facility enhancement has failed to get done several times now.  Nonetheless, there must be a way to filter the search results to only names that the user actually has access to, probably even in the type-ahead feature.  @teryhill @tmccormi  how do you guys feel about this level of paranoia?

teryhill:
@aethelwulffe where are you in your thought process on this. 

Where are you on the facility ACL modification.

Where is Waldo?

aethelwulffe:
Waldo is saying "Don't implement search features unless ACL and TAG rules are minded.

naveen17797:
So patients with restriction tags should not be shown in autocomplete feature?

aethelwulffe:
Where are we on this folks?

tmccormi:
Let's re-do this PR as just an commit of useful autocompete functions that can be used.  Implementing this feature in the UI shown (New/Search) is not really useful, but the functions are.

aethelwulffe:
@Trodrige Can you grab the little /ajax part of this (which is the real code) and use it in the report engine?

teryhill:
Any more progress on this?

aethelwulffe:
@naveen17797 The autocomplete part of this is indeed valuable as a core function.  I think I agree with Tony that as a feature, it doesn't gain you much considering the access control hooks that are required in this context, but we really want the library!

naveen17797:
yes @teryhill  @aethelwulffe 
lets redo the PR with access controls like a function which can be used in any form fields, so as to take data from particular column of table and fill the field
more or less like 
function autcomplete($tablename, $columnname, $restrictionclause) { }
like this

aethelwulffe:
When we take this on...

function autcomplete($tablename, $columnname, $restrictionclause, $option1, $etc...) { }

Needs to work with `lists`, layout_options etc...

aethelwulffe:
Stale.  If this is an active PR, please create new.

PR #1053, opened 2018-03-06
Box 14 onset date with qualifier

teryhill:
These were in the old version before the fork, Just added them back. Had
an OB Clinic that needed box 14 with LMP as of the first of the year insurance companies are requiring it. 
I removed these a couple of years back when the onset date was added to the encounter form.


library/gen_x12_837.inc.php
```diff
@@ -496,6 +496,8 @@ function gen_x12_837($pid, $encounter, &$log, $encounter_claim=false) {
     "*Y" .
     "~\n"; 
 
+   // above is for old claims that use the encounter onset date field. Onset is now back in misc_billing_options
+
```
aethelwulffe: Default should still pull from current encounter onset date unless it is already set.  Warning would be good in the misc_billing if the two are different when you edit.  Encounter dates are sometimes corrected belatedly, then the form has to get corrected. (Not relevant to this file).
Relevant to this file, I _think_ the below will set the onset date to whatever is in the encounter if it is not specifically set in misc_billing_options, correct?
  This is very important behavior for Statewide Inpatient Psychiatric Program (SIPP) (which exists in every state).

aethelwulffe: I have some questions, but yes, It seems to fire.  I didn't try to generate an x12 file yet, so I am holding off on approval until I do that tomorrow.

aethelwulffe:
I think we are the only ones that care about this, so I am merging.
Note, sometime pulling down a branch with a long name runs...slower than usual.  Weird.  Probably not the cause, but...

PR #1062, opened 2018-03-16
Fix for #1061: User edit UI inaccessible

### LINKED ISSUE #1061: User edit UI inaccessible
When you have more than 50 or so users in either the active or inactive list, and try to edit them, the modal window locates itself at the bottom of the list, and becomes impossible to interact with.  This is a big issue for two clinics I know right now that have respectively 450 and 1500 user table records due to long use and a sprawling organization.

This issue was fixed for the facility editor (I believe).

apooravc:
Fixes #1061 
Before:
image1

After:
after

The popups such as "user edit" and "user add" were inaccessible because of the dependence of css top property values calculated in fancybox css/js files on the scroll of document. A temporary solution is to make corresponding custom files with _position: fixed_ and _top: 0_ and call these changes from wherever the iframe of the popups are. That solution was already available in the code base after [PR 869](https://github.com/LibreHealthIO/LibreEHR/pull/869) was merged. So in this PR, fancybox-custom css/js files are added, identical to the fancybox-addpatient files in PR 869 and code is added in headers.inc, usergroup_admin to call fancybox-custom instead of fancybox, similar to additions in PR 869.


interface/usergroup/usergroup_admin.php
```diff
@@ -103,19 +103,6 @@
               $tqvar = formData('mname','P');
               sqlStatement("update users set mname='$tqvar' where id= ? ", array($_POST["id"]));
       }
-      if ($_POST["role_name"]) {
```
teryhill: This code needs to be left in for the full screen pr. You need to update your code and resubmit.

teryhill:
@apooravc  a quick look. It looks good except you are overwriting some of the code for the full screen and roles stuff. You need to update your code and send it again.

interface/usergroup/usergroup_admin.php
apooravc: At it.

teryhill:
@apooravc  I will test it later today. Thanks.

PR #1063, opened 2018-03-16
Fix for #882

### LINKED ISSUE #882: Pressing Lock and then Refresh on a UI Element Creates Two Instances
**Steps to Reproduce**
1. Click on the 'Lock' on one of the UI elements. This hides the element.
libreehr9

2. Whilst the element is hidden, click on the 'refresh' button. Observe two instances of the same element being created.
libreehr10

Seems like if you open the hidden one, the opened one gets replaced, though. So in the end only one instance of the element will be displayed anyway. Minor UI bug.

**Reproduction Rate**: 5/5

System Specs LibreHealth EHR v2.0.0, Chrome Version 63.0.3239.84 (Official Build) (64-bit), Windows 10 Version 10.0.14393 Build 14393

apooravc:
Fixes #882 
Removed the refresh icon from tabs when in locked/ hidden state.

Tabs UI when locked:
bef

Tabs UI when unlocked:
af

### DIFF
--- a/interface/main/tabs/templates/tabs_template.php
+++ b/interface/main/tabs/templates/tabs_template.php
@@ -5,7 +5,6 @@
         <!--ko if:!locked() -->
             <span class="tabSpan tabFloat bgcolor2">
                 <span  data-bind="text: title, click: tabClicked, css: {tabHidden: !visible()}"></span>
-                <span class="fa fa-refresh tab-button" data-bind="click: tabRefresh" title="Refresh"></span>
                     <!--ko if:!locked() -->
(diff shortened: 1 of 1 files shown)


teryhill:
Better than grits and fried catfish for breakfast.

aethelwulffe:
Not as good as Fried Chicken and Watermelon though...but I do want me some grits...the catfish doesn't sound bad either...  Gee, I think I am getting hungry.  Time to go catch me some shrimp!

teryhill:
Send me a couple of Boxes

aethelwulffe:
I do the five-gallon bucket.  Tonight is a new moon, so I will check to see what the tide is like...if I have time, which I won't.  Nice to dream.

PR #1065, opened 2018-03-19
Fix for #343: Admin tool to show who's logged in now is needed

### LINKED ISSUE #343: Admin tool to show who's logged in now is needed
I wrote this SQL a while back - we could use a simple UI for admins that does this..

select user, IF( (mod(count(user),2)=1),'YES','NO') as logged_in from log where DATE_FORMAT(`date`,'%Y-%M-%d') = DATE_FORMAT(now(),'%Y-%M-%d') and (event = "login" or event = "logout") group by user

apooravc:
Fixes #343 

Added Users Activity as an Administration menu item to display user's last activity during present date. 

Used following SQL query:

SELECT `user`, `event`, `date`
FROM `log`
WHERE `date` 
IN 
(SELECT  MAX(`date`) 
 FROM `log` 
 WHERE DATE_FORMAT(`date`,'%Y-%M-%d') = DATE_FORMAT(NOW( ),'%Y-%M-%d') 
 GROUP BY `user`)

On a sample of table `log` today

on

And got

gives

Finally displayed as

displays

Properly logging in and properly logging out cases are covered. About the auto-logout case, when a user login but doesn't logout and the event auto-logout occurs, does that make a new record in table `log` like previous two cases? If it does, I could add the corresponding `event` attribute value in this to make sure the user is active when LOGGED IN. 

Also what are other events (besides `login` & `logout`) like `view` for in `event` attribute of `log` table?

2) After cloning, only the master branch would be available locally. Enter `git checkout -b users-last-activity origin/users-last-activity` or `git branch users-last-activity origin/users-last-activity` to get the branch _users-last-activity_ locally.
3) The changes could then be seen on Users Activity page.

@tmccormi  @aethelwulffe Please have a look.


aethelwulffe:
Observations:
1.  This does only work if the globals for logging access is set.
2.  SELECT user, event, date
```
FROM log
WHERE date
IN
(SELECT MAX(date)
FROM log
WHERE DATE_FORMAT(date,'%Y-%M-%d') = DATE_FORMAT(NOW( ),'%Y-%M-%d')
AND `event` = 'view' //(or other event value type
GROUP BY user);
```
Is another option.
3.  In practice, enabling a tiny tool display for who is active in the message view might be good.  The format you have here would be useful, and so would a sort of "who's online" display like you commonly see in forum software is also a potential useful thing.  This certainly makes it easier for sysadmins to check and see if anyone is active before doing maintenance.
4.  Digdown on a single user's daily activity is a great boss spy kit item.

Concerns:  Very large logs (and they are often the greatest volume of a whole database as currently configured defaults set them) may present an issue for any type of active refresh activity indicator.

apooravc:
@aethelwulffe about the 4th point and the concern point, this tool will display a user's last activity and not all the activities of a user, preventing large logs and maximum no. of rows would be the total no. of users. Is my understanding correct here?

About the 3rd point, this tool's current format lacks the active checking part because of auto logout case, so if the system auto logout, does that get registered in `log` table? Working on the tiny tool display part.

aethelwulffe:
@apooravc  actually, I kind of think tracing auto logouts is sort of a nifty feature in of itself.  Folks that don't log out of the application when they leave their desk for lunch, leave their laptop on all the time (lid closers) etc...

I believe that autologs are recorded in the log as logout.  I could look and see exactly what actions take place then...

apooravc:
@aethelwulffe @teryhill Worked on something, please have a look.

online

The above list shows users who have logged in successfully (`event = login` & `success = 1`) and are active. The test for active is based on a refresh interval (taken 5 seconds in PR below). After each such interval of time, tool checks for the last activity of all users during present date in log table. Inactive users would have (`event = logout` & `success = 0` due to autologout) and logged out users would have (`event = logout` & `success = 1`), since tool registers only logged in successfully cases, all active users are listed after each such interval. So, active here means the state of user being registered as login till either registered as logout or autologout in log table. 

There are two methods to refresh list(request latest data from log table & recreate list from it):

1.  Refreshing complete tab to run php script again at an interval.
`<script type="text/javascript">
  $(document).ready(function() {
    setInterval(function() {
      window.location.reload(true);
    }, 5000);
  });
</script>`

2. Running php script again at an interval by AJAX to avoide complete tab refresh, as I've done in my commit below. On testing, it runs but there is a `deprecation warning` of synchronous XHR in inspect window even though this makes asynchronous XHR.

tmccormi:
@aethelwulffe are you ready to approve this one or does it need more testing?

tmccormi:
@apooravc how well does this work (UI wise) with say 60-100 users?  I have customers like that...

interface/main/messages/messages.php
```diff
@@ -51,19 +51,87 @@
...
+    setInterval(function() {
+      $.ajax({
+          type: "GET",
+          url: "user_activity.php",
+          dataType: "html",
+          success: function(response) {
+            $(".listbox").empty().append(response);
+          }
+      });
+    }, 5000);
```
teryhill: The time interval needs to be configurable as a global. Maybe times in 10 sec intervals from 10 to 60. Will this will impact performance? (I don't think so I think it should have the same impact as the flow board refresh)

interface/main/messages/messages.php
```diff
@@ -51,19 +51,87 @@
...
+  //select users whose last activity during
+  //present date is logging in successfully
+  $query = "SELECT `user`
+            FROM `log`
+            WHERE `date` IN (SELECT  max(`date`)
+                             FROM `log`
+                             WHERE DATE_FORMAT(`date`,'%Y-%M-%d') = DATE_FORMAT(now(),'%Y-%M-%d')
+                             GROUP BY `user`)
+                  AND `event` = 'login' AND `success` = 1";
+  $query .= " LIMIT 500";
```
teryhill: Is this limit going to be an issue o the larger systems? @aethelwulffe will 500 users be enough?

interface/main/messages/user_activity.php
```diff
@@ -0,0 +1,30 @@
...
+  //select users whose last activity during
+  //present date is logging in successfully
+  $query = "SELECT `user`
+            FROM `log`
+            WHERE `date` IN (SELECT  max(`date`)
+                             FROM `log`
+                             WHERE DATE_FORMAT(`date`,'%Y-%M-%d') = DATE_FORMAT(now(),'%Y-%M-%d')
+                             GROUP BY `user`)
+                  AND `event` = 'login' AND `success` = 1";
+  $query .= " LIMIT 500";
```
teryhill: Same question on the limit

teryhill:
@apooravc does the auto logout still work, There is a provision in the system where the user is logged out if there is no activity, If the user has this screen up will they still be auto logged out by the system due to inactivity or does the refresh here look like activity ?

sites/default/menu_data.json
```diff
@@ -322,6 +322,14 @@
         "children" : [ ],
         "requirement" : 0,
         "acl_req" : [ "admin", "users" ]
+    }, {
+        "label" : "Users Activity",
+        "menu_id" : "adm0",
+        "target" : "adm",
+        "url" : "/interface/usergroup/users_activity.php",
```
aethelwulffe: Please document this feature...somewhere.

aethelwulffe: Need to test this again from a server and at least two client machines.  Honeypot Terry?

teryhill: Ok when .

apooravc:
Removed the "Users Activity" option from menu, it was for my initial solution.

interface/main/messages/messages.php
apooravc: I see in forums that they have a fixed refresh time interval saying "Based on last 5 minutes activity of users here is the list of online ones".

interface/main/messages/messages.php
apooravc: 500 online users at same time seems like a reasonable limit but maybe not enough for larger staff

aethelwulffe:
I set Administration>Globlas>Security>Idle Session Timeout to 60 seconds.  Works.  That was on older code.
OK, I am all straight on the menu thing.
demo is here:
http://librehealthehr.suncoastconnection.com/LibreEHR/interface/login/login.php?site=default
Create a user and log in...while we chat here or something.  admin/pass to get in.
I set the autologout to 120 seconds.

aethelwulffe:
Hunh...Uh, I saw terry login to the test site, then after a bit, only his name was visible.  At first, it said ART and TERRY.  Then just his name.
What happened?  Is this intentional?

teryhill:
Did you get auto logged out?

apooravc:
@aethelwulffe I also wanna try this live site. How to create user and login?

aethelwulffe:
I didn't get logged out.  It is still active due to this tool I imagine.
@apooravc  go to the URL, log in as admin/pass 
go to administration/users and create a new user.  Log out and log in again with your new user.

apooravc:
Shows me as Tony!

aethelwulffe:
Yes, you were logged in as "Tony Stark"...We thought it was Tony McCormick...Confused us at first.
Apparently the client-side function of updating the list has the effect of keeping you logged in.

interface/main/messages/messages.php
aethelwulffe: 500 is a crazy lot, but this can be tweaked in an in-page configuration setting.

sites/default/menu_data.json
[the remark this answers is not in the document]
apooravc: Yes, I removed it because it was a part of my initial commit (earlier solution in which I showed last activity of each user on that page). Not needed in this solution.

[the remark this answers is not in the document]
aethelwulffe: Gotcha.  Did a test, and the only thing we are now worried about is the autologging.  Otherwise, it is quite nice.

apooravc:
@aethelwulffe Yes, the client side update keeps me logged in and auto logout works when I close the tab. Need to think of something. Any ideas?

aethelwulffe:
@apooravc I am hunting...I think the normal messages check doesn't interfere.  Maybe we can find a hint there.  I am SURE that the queries from this tool need to be of the NoLog variety though.  These will fill the log up fast.

apooravc:
Continued on Gitlab as [MR 1105](https://gitlab.com/librehealth/ehr/lh-ehr/merge_requests/1105). Closing this.

apooravc:
Replaced on Github by #1119.

Issue #1066, opened 2018-03-19
Patient Portal Email Validation Update Required


amughalbscs16:
The function used for validation of email in patient_portal is using a self defined function. I think it should be using the php built in function for checking whether the email is correct or not.
![register function](https://user-images.githubusercontent.com/28474804/37605735-a5fabbd8-2bb5-11e8-87cc-a0abfcbbbd40.PNG)
**Presently implemented Function:**
function validEmail($email)
{
    if (preg_match("/^[_a-z0-9-]+(\.[_a-z0-9-]+)*@[a-z0-9-]+(\.[a-z0-9-]+)*(\.[a-z]{2,3})$/i", $email)) {
        return true;
    }

    return false;
}

**The Suggested Update For the Function:**

function validEmail($email)
{
if (filter_var($email, FILTER_VALIDATE_EMAIL)) {
      return true; 
    }
 return false;
}

teryhill:
I like using "Built In" Functions. I think that the current function chokes on a .io site.

amughalbscs16:
So should I make a commit with this new function? If you also agree with it.

teryhill:
Yes send a commit for it.

amughalbscs16:
Here is the pull request I have sent please accept it soon. This function works for .io domain as well I have tested it on that too.
The Pull Request Link
https://github.com/LibreHealthIO/LibreEHR/pull/1067

amughalbscs16:
Solved

Issue #1057, opened 2018-03-11
Patient ledger report needs to say no data to display , Currently it just disappears. 


Trodrige:
The patient ledger report results aren't displayed long enough.
- Fill the form and submit.
Observe the results show up, and disappear almost immediately.

teryhill:
@Trodrige  it will do that if there is nothing to display.

amughalbscs16:
@teryhill please share an image so, I can work on resolving the matter. Please specifically address where is the problem coming.

aethelwulffe:
Yeah, posting menu breadcrumbs or full "reproduce this" instructions might be a good standard while GSOC is going on.

amughalbscs16:
@aethelwulffe where on which page is this error coming, I am searching for it. guide me where is the error coming so I can solve it.

teryhill:
Select a patient and then select ledger from the demographics screen.

![image](https://user-images.githubusercontent.com/4655366/37616624-ebb26484-2b86-11e8-82e6-d0be646d27de.png)

amughalbscs16:
what should be displayed do you want a modal to open or an alert which tells the patient that no records were found?

amughalbscs16:
I think there should be an alert or a modal opening which tells the user that 0 records were found.

teryhill:
Sounds good to me. An alert would be best I think since this can also be ran from the report menu.

amughalbscs16:
Plus a sleep for 5 seconds which will help user see the empty table as well
with the alert?

teryhill:
Okay

amughalbscs16:
I thought putting a sleep wouldn't be the best option, so I added an alert when there are no records to show. Here is the link to the pull request. https://github.com/LibreHealthIO/LibreEHR/pull/1067.
Solved.

Trodrige:
@amughalbscs16 I noticed you solved this after doing a PR. I used another approach. 
1. The table doesn't appear if there's no result and
2. The message is in red.

I'm not going to steal this from you! This is how my change looks like.

![patient-ledger](https://user-images.githubusercontent.com/16147419/37698494-7f3c1dfa-2ce3-11e8-815f-50436ce04068.png)

amughalbscs16:
I have given a screenshot of the solution in the PR it shows an alert only.
when there are no records.

aethelwulffe:
We have GOT to get assignments worked out here.  Too many cool people are working on great solutions...for the same things.
[...]

PR #1067, opened 2018-03-19
Fix for #1066 Email Validation Functions

amughalbscs16:
Email Validation Function Updated, It choked at .io domain before now it doesn't.

### DIFF
--- a/patient_portal/account/account.lib.php
+++ b/patient_portal/account/account.lib.php
@@ -122,10 +122,9 @@ function generatePassword($length = 8, $strength = 1)
 
 function validEmail($email)
 {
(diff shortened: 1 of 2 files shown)


amughalbscs16:
I've added an extra commit regarding the alert error similar to issue #1057, but this one occured in Patient Portal -> Account - > Ledger When there were no records to be fetched. Please accept that commit too.
no matches on reports patient ledger patient portal

teryhill:
@amughalbscs16  you will need to , after this one, submit separate PR's for your changes. @tmccormi can you help with git learning for this user?

amughalbscs16:
Terry I understood how to do it, thanks. I searched and found the solution
later

PR #1069, opened 2018-03-20
Fix for #858: Log the reasons statements did not print. 

### LINKED ISSUE #858: Log the reasons statements did not print.
In the /sites/default directory is the statement.inc.php. It will currently ignore a statement if the dollar amount is less that the setting in Globals. The code in the statement file to do this is at about line 106 and line 462. There needs to be some way of letting the end user know that statements that they selected did not print for these reasons. Instead of just returning these lines need to have more logic and then return.

  #minimum_amount_due_to _print
  if ($stmt['amount'] <= ($GLOBALS['minimum_amount_to_print']) && $GLOBALS['use_statement_print_exclusion']) return "";
  if ($stmt['statement_print'] == "NO" && $GLOBALS['use_statement_print_exclusion']) return "";

apooravc:
Fixes #858 

Added a GLOBAL variable `$GLOBALS['stmts_below_minimum_amount']` to keep count of ignored statements having amount value less than `$GLOBALS['minimum_amount_to_print']`. This reason along with number of ignored statements are logged to user at end of `$STMT_TEMP_FILE` after all the `make_statement` calls are done.

2) After cloning, only the master branch would be available locally. Enter `git checkout -b log-reasons origin/log-reasons` or `git branch log-reasons origin/log-reasons` to get the branch _log-reasons_ locally.
3) The changes could then be seen on Fees/Posting page.

@teryhill Please have a look.

### DIFF
--- a/sites/default/statement.inc.php
+++ b/sites/default/statement.inc.php
@@ -106,13 +106,22 @@ function create_HTML_statement($stmt) {
 
   #minimum_amount_due_to _print
   if ($stmt['amount'] <= ($GLOBALS['minimum_amount_to_print']) && $GLOBALS['use_statement_print_exclusion']) {
-    //increment whenever statement is ignored due to if condition
+    //increment whenever amount value of statement is less than minimum amount value and
+    //to keep count of such ignored statements in final print
     $GLOBALS['stmts_below_minimum_amount']++;
     return "";
   }
 
-  if ($stmt['statement_print'] == "NO" && $GLOBALS['use_statement_print_exclusion']) return "";
+  if ($stmt['statement_print'] == "NO" && $GLOBALS['use_statement_print_exclusion']) {
+    //increment whenever the patient is set to no statement and
+    //to keep count of such ignored statements in final print
+    $GLOBALS['stmts_set_no_patient']++;
+    return "";
+  }
   if ($GLOBALS['use_statement_print_exclusion'] && $GLOBALS['insurance_statement_exclude'] !=4 ){
+   //increment whenever the insurance company does not allow statements and
+   //to keep count of such ignored statements in final print
+   $GLOBALS['stmts_not_allowed_insurance_company']++;
    if ($GLOBALS['insurance_statement_exclude'] ==0 && $stmt['insurance_no_statement_print_pri']) return "";
    if ($GLOBALS['insurance_statement_exclude'] ==1 && $stmt['insurance_no_statement_print_sec']) return "";
    if ($GLOBALS['insurance_statement_exclude'] ==2 && $stmt['insurance_no_statement_print_tri']) return "";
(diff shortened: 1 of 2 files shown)


teryhill:
@apooravc This looks great. Can you expand it to the other reasons they will not print?

apooravc:
@teryhill Sure. Identifying other cases...

teryhill:
This one is  The patient is set to no statement. 
if ($stmt['statement_print'] == "NO" && $GLOBALS['use_statement_print_exclusion']) return "";
This is The Insurance company does not allow statements.
 if ($GLOBALS['use_statement_print_exclusion'] && $GLOBALS['insurance_statement_exclude'] !=4 ){

apooravc:
Added the other cases @teryhill Please have a look.

teryhill:
@apooravc  get this on testing.

image

apooravc:
@teryhill I didn't add a check for count 0, wrong slash in newline and I need to give in more newlines.

teryhill:
image

Much better.

teryhill:
@apooravc This is really great. Is there a way that you can add the patient name of the statements that did not print. So if the user needs to check them they can?

apooravc:
@teryhill May be by using `getPatientName(pid)`?

aethelwulffe:
An array for the output would be great.  What might also be good is to have a db entry (ar_session table? ar_activity? billing table?) which exists somewhere with the "last statement" or using the statement count feature to change a status to a value for 'failed to print' or something.  Not stating that this is a thing that must be done, just an idea we should check into to see if it gives a valid addition to workflow features.

apooravc:
@teryhill Added the functionality for corresponding patient names to appear in the log. Please have a look.

teryhill:
Looks good 

image

teryhill:
The array for the insurance company does not allow statements is picking up both the ones that do not print and the ones that actually print.

sites/default/statement.inc.php
```diff
@@ -105,9 +109,29 @@ function create_HTML_statement($stmt) {
...
+    //also add corresponding patient name
+    array_push($GLOBALS['pat_set_no_stmt'], getPatientName($stmt['pid']));
+    return "";
+  }
   if ($GLOBALS['use_statement_print_exclusion'] && $GLOBALS['insurance_statement_exclude'] !=4 ){
+   //increment whenever the insurance company does not allow statements,
+   //to keep count of such ignored statements in final print
+   $GLOBALS['stmts_not_allowed_insurance_company']++;
+   //also add corresponding patient name
+   array_push($GLOBALS['pat_not_allowed_insurance_company'], getPatientName($stmt['pid']));
```
teryhill: move these lines

sites/default/statement.inc.php
```diff
@@ -105,9 +109,29 @@ function create_HTML_statement($stmt) {
...
    if ($GLOBALS['insurance_statement_exclude'] ==0 && $stmt['insurance_no_statement_print_pri']) return "";
    if ($GLOBALS['insurance_statement_exclude'] ==1 && $stmt['insurance_no_statement_print_sec']) return "";
```
teryhill: To this section

sites/default/statement.inc.php
```diff
@@ -105,9 +109,29 @@ function create_HTML_statement($stmt) {
...
+    array_push($GLOBALS['pat_set_no_stmt'], getPatientName($stmt['pid']));
+    return "";
+  }
   if ($GLOBALS['use_statement_print_exclusion'] && $GLOBALS['insurance_statement_exclude'] !=4 ){
+   //increment whenever the insurance company does not allow statements,
+   //to keep count of such ignored statements in final print
+   $GLOBALS['stmts_not_allowed_insurance_company']++;
+   //also add corresponding patient name
+   array_push($GLOBALS['pat_not_allowed_insurance_company'], getPatientName($stmt['pid']));
    if ($GLOBALS['insurance_statement_exclude'] ==0 && $stmt['insurance_no_statement_print_pri']) return "";
```
teryhill: To this section

sites/default/statement.inc.php
```diff
@@ -105,9 +109,29 @@ function create_HTML_statement($stmt) {
...
    if ($GLOBALS['insurance_statement_exclude'] ==1 && $stmt['insurance_no_statement_print_sec']) return "";
    if ($GLOBALS['insurance_statement_exclude'] ==2 && $stmt['insurance_no_statement_print_tri']) return "";
```
teryhill: To this section

apooravc:
@teryhill Did the required changes. Please have a look.

teryhill:
I am looking and so far they look good.

PR #1070, opened 2018-03-20
Fix for #1057 Patient Ledger Reports Alert

### LINKED ISSUE #1057: Patient ledger report needs to say no data to display , Currently it just disappears.

amughalbscs16:
fixes #1057  
Now when there are no records, in the patient ledger. There is an alert showing that there are no records. 
They should search again. 
Here is the screenshot.

no matches on reports patient ledger


Trodrige:
The message appears in the alert and before the table. Keep one!
Do we need the table head if there's no result to display?

amughalbscs16:
It gives user a better view and idea and hints about the fact that no
records exist. It's about user experience

amughalbscs16:
I appreciate that you are trying to help me but I did that because It gives a better and fast appearance of empty table to the user and a user feels like easier to interpret the meaning. We have studied this thing in our subject Human Computer Interaction.

aethelwulffe:
Let's not go against the HCI instructor!  😈 

If you want to tweak this, you could always say 
echo GLOBALS[whatever the clinic name value is]." EHR says: you didn't find nuthin'"  Instead of echoing the host name...which could be very weird and frightening to some non-tech ppl.  
  It's something I used to teach in my Control Interface classes.  ⚙️ 🙉 😸   Don't print out any technobabble where the "normal" people can see it if you don't want sysadmin support calls in the middle of the night.

amughalbscs16:
So you want that it should remove the word local Host from the alert right?

teryhill:
Can you remove the "Localhost" from the dialog. Just give the message. @amughalbscs16 . I have and am testing the changes.

naveen17797:
@amughalbscs16 use jquery modal instead of alert, it will be user friendly
its just a suggestion :relaxed:

amughalbscs16:
Okay, I'll do that.

amughalbscs16:
I gave that suggestions to Terry Hill you can read in the issue he told me
to tryalert instead so I went with alert.

teryhill:
What does teryhill know. 8^) Try it with the modal and lets see what it looks like.

naveen17797:
screenshot from 2018-03-21 21-37-34

it would like this, when done with jquery ui. design a model like this

aethelwulffe:
alert() is less expensive of course, but I would pay the price to keep 
from freaking out or confusing our mental health counselors...they are 
kinda nuts anyway.

-Sorry for the noise over little nit-pickies.  Few people actually enjoy 
the actions of publishing editors...and every one of us needs to play 
that role for each other.

amughalbscs16:
I'm on it

aethelwulffe:
Commit that!

amughalbscs16:
Same has been setup for Admin Patient Ledger now you can merge the commit @aethelwulffe  Thanks for the guidance.

interface/reports/pat_ledger.php
```diff
@@ -347,20 +347,72 @@ function sel_patient() {
...
   }
+  .modal-content {
+      background-color: #fefefe;
+      margin: 15% auto; /* 15% from the top and centered */
+      padding: 20px;
+      border: 1px solid #888;
+      width: 80%; /* Could be more or less, depending on screen size */
+  }
+
+  /* The Close Button */
```
aethelwulffe: I like the documentation here 🥇

interface/reports/pat_ledger.php
```diff
@@ -347,20 +347,72 @@ function sel_patient() {
...
 
 <title><?php echo xlt('Patient Ledger by Date') ?></title>
 
 <script language="JavaScript">
+  
  $(document).ready(function() {
   var win = top.printLogSetup ? top : opener.top;
   win.printLogSetup(document.getElementById('printbutton'));
  });
+function hideNoRecordsModal(){
```
aethelwulffe: Another function for the report generator system to grab.

patient_portal/report/pat_ledger.php
```diff
@@ -436,10 +458,38 @@ function sel_patient() {
       <?php // can add any additional javascript settings to datetimepicker here; need to prepend first setting with a comma ?>
  });
   });
+  function showNoRecordsModal(){
+    var modal = document.getElementById('noLedgerRecords');
```
aethelwulffe: Be good in an public library.  I am just taking note @Trodrige

PR #1072, opened 2018-03-20
Check for clinic hours in creating calendar events

### LINKED ISSUE #1071: Calendar events need to respect clinic hours
The system will and does give an alert that the provider is not available but allows the action anyway. 

If you create an appointment for 1 pm and then change the time to 11 but forget to change the PM to am you now have an appointment set for 11 pm

image

The appointment should not be allowed if it is before or after the clinic hours.

image

The AM/PM flag should be set to reflect the clinic open hours.

So the clinic does not end up with this 

image

prabhakarpd7284:
Fix #1071.

An alert box pops on wrong timing entry.

image


modules/calendar/add_edit_event.php
```diff
@@ -230,6 +230,15 @@ function DOBandEncounter()
         if ($_POST['form_ampm'] == '2' && $tmph < 12) $tmph += 12;
         $duration = abs($_POST['form_duration']); // fixes #395
     }
+       
+     if($tmph*60+$tmpm > $GLOBALS['schedule_end']*60 or $tmph < $GLOBALS['schedule_start']){
+        echo "<html>\n<body>\n<script language='JavaScript'>\n";
+        echo 'alert("Please select time between clinic hours.");';
```
aethelwulffe: xlt()

prabhakarpd7284: I will put a global to turn this feature on and off as suggested by Teryhill.

prabhakarpd7284:
@teryhill Have a look.
image

teryhill:
closed in gitlab so closing here

PR #1073, opened 2018-03-21
Fix for #1071: Calendar events need to respect clinic hours

### LINKED ISSUE #1071: Calendar events need to respect clinic hours

apooravc:
Fixes #1071 

Added a check for user selected time (taking AM/PM into account) to be within clinic hours. The check would alert the user and prevent the appointment from being made otherwise.

2) After cloning, only the master branch would be available locally. Enter `git checkout -b check-cal-event origin/check-cal-event` or `git branch check-cal-event origin/check-cal-event` to get the branch _check-cal-event_ locally.
3) The changes could then be seen on Calendar screen.

@teryhill Please have a look.

### DIFF
--- a/modules/calendar/add_edit_event.php
+++ b/modules/calendar/add_edit_event.php
@@ -231,7 +231,7 @@ if ($_POST['form_action'] == "duplicate" || $_POST['form_action'] == "save")
         $duration = abs($_POST['form_duration']); // fixes #395
     }
     //check for selected appointment time
-    if ($GLOBALS['check_appt_time'] == '1') {
+    if ($GLOBALS['check_appt_time'] == 1) {
       //before starttime gets the values from form_hour & form_minute after clicking save,
       //check if they (tmph & tmpm) are within clinic hours
       $user_ampm = $_POST['form_ampm'];
--- a/modules/calendar/add_edit_event.php
+++ b/modules/calendar/add_edit_event.php
@@ -230,24 +230,26 @@ if ($_POST['form_action'] == "duplicate" || $_POST['form_action'] == "save")
         if ($_POST['form_ampm'] == '2' && $tmph < 12) $tmph += 12;
         $duration = abs($_POST['form_duration']); // fixes #395
     }
-    //before starttime gets the values from form_hour & form_minute after clicking save
-    //check if they (tmph & tmpm) are within clinic hours
-    $user_ampm = $_POST['form_ampm'];
-    echo var_dump($_POST['form_ampm']);
-    if ($user_ampm == 1) {
-      #AM, 00:00 midnight to 11:59 noon
-      $user_selected_time = $tmph + ($tmpm/60);
-    } else {
-      #PM, 12:00 noon to 23:59 midnight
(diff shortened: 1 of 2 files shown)


aethelwulffe:
I will look, but we also have #1072 PR on this.  We need to do a collaboration of methods.  We also need to start assigning people to issues.  Assignments should be made to people who are validly active, and we need to start maintaining out lists of contributors by ensuring they can add labels and get assignments.
-We need to review our github ACL...

apooravc:
I understand your concerns @aethelwulffe , how does a gallery of active contributors with their Github profile image and link on the README of this repo sounds? Sort of like [OpenEMR](https://github.com/openemr/openemr) has on their Github repo.

About this PR and issue, I was already assigned by @teryhill and I pushed a solution as soon as I could.

aethelwulffe:
We have the gallery here too...Complete with the top contributor who committed/attested almost everything under his own name...author or not...from the fork source.  What I am talking about is being able to cleanly allow walk-in contributors to add labels, and easily enable assignment of issues to people (or self-assign).  Since word is that a move to gitlab is being forced, I suppose that is an issue for later.

apooravc:
@teryhill Added a `GLOBAL` to turn this feature on/off. Please have a look.

modules/calendar/add_edit_event.php
```diff
@@ -230,6 +230,28 @@ function DOBandEncounter()
         if ($_POST['form_ampm'] == '2' && $tmph < 12) $tmph += 12;
         $duration = abs($_POST['form_duration']); // fixes #395
     }
+    //check for selected appointment time
+    if ($GLOBALS['check_appt_time'] == '1') {
```
teryhill: I had to make this if ($GLOBALS['check_appt_time'] == 1) to get it to work on my system

teryhill:
@apooravc I am having an issue with the global working. See my code comment.

modules/calendar/add_edit_event.php
apooravc: @teryhill I thought this global was carrying the string 1 rather than numeric 1 when checked. Apparently, that's not the case.

apooravc:
@teryhill Did the required changes. Please have a look.

Trodrige:
@apooravc I'm testing this. Nothing happens when I click 'All', 'Current' or 'Future'. I can only cancel. See below
apoorv

My 'Check clinic hours' global is checked
apoorv1

aethelwulffe:
OK, it is working in either case with the gobals.

aethelwulffe:
I am good with leaving it as a default off (unset in the array).  I checked it before testing though, so I don't know what happens if you don't set the global array value before starting.  It has a conditional statement, and probably would not crash even if it didn't.

PR #1075, opened 2018-03-21
Fix for #956: Tabs not always resizable

### LINKED ISSUE #956: Erratic behavior on resizing tabs
Mentioned in #954 .

Behavior:
On log-in, rightmost tab has resize tooltip.
Add third tab, and the rightmost no-longer has tooltip.
Open fourth tab, and only the left two are resizable.
Close the two left tabs, and none of the remaining are resizable.
Open or close any tabs, yet never get your resize back.

apooravc:
Fixes #956 

Upon initial loading of the main page, a DOM of `mainframe` is constructed with two tabs and two divs having `handle` classes are displayed as spaces next to tabs. Tabs are resized by using such `handle` divs. This functionality is achieved through `drags()` calls to resize plugin which attach events `mousedown`, `mousemove` and `mouseup` to `handle` divs on which the calls are made. These events are responsible for resizing.

Upon initial loading, DOM is constructed and the `drags()` calls are made on all `handle` divs in DOM. Problem is, initial DOM contains only the first 2 `handle` divs next to first 2 tabs. That's why `drags()` calls are made only on these and resize plugin works for the initial 2 tabs. After opening a new tab, a new `handle` div appears in DOM but the `drags()` call isn't made on it to resize plugin for attaching those events to it. Hence, resize plugin doesn't work for newly opened tabs.

A solution would be to make `drags()` calls to a latest complete array of `handle` divs or `$('.handle')` whenever a new tab is opened or _clicked_ and not only upon initial loading.

2) After cloning, only the master branch would be available locally. Enter `git checkout -b make-tabs-resizable origin/make-tabs-resizable` or `git branch make-tabs-resizable origin/make-tabs-resizable` to get the branch _make-tabs-resizable_ locally.
3) The changes could then be seen on main screen.

@teryhill @aethelwulffe Please have a look.


aethelwulffe:
Darn it.  I wanna take a look at this right now, but my time is up for an hour or so... :(

teryhill:
@apooravc try it with opening 4 tabs and then locking the two in the middle.

apooravc:
@teryhill Some problems in resizing, though it's resizing but a bit. Might be due to 2 `handle` divs in such cases.

aethelwulffe:
Still very much wanting to see progress on this.  If there are way too many issues to solve, and we need a new approach, we better get it into the lineup for GSOC, eh?

apooravc:
@aethelwulffe This PR solves the issue for `tab-space-tab` cases for which the plugin is made. The case mentioned by @teryhill falls outside the current scope of plugin and requires changes to the plugin core itself, I think but I maybe wrong.

teryhill:
@aethelwulffe I tested this and it worked as needed and agree with @apooravc  comments

PR #1078, opened 2018-03-23
Fix for #1076 Function update register patient

### LINKED ISSUE #1076: Increase readability Creation of a new Patient Code Updation
There is a condition checking for if the result['pid'] is > 1 it should instead pe result['pid'] >= 1 as in the case of there is no patient the 1st value for patient id  = 0; The patient id for the 2nd will be 1 and hence the code seems to be something difficult to understand what. The New code is easy to understand for a new person who is starting to read the code. What do you think @teryhill.
Old code:
function getNewPid()
{
    $result = sqlQuery("select max(pid)+1 as pid from patient_data");
    $newpid = 1;
    if ($result['pid'] > $newpid) {
        $newpid = $result['pid'];
    }
    if ($newpid == null) {
        $newpid = 0;
    }
    return $newpid;
}

New Proposed Code:
function getNewPid()
{
    $result = sqlQuery("select max(pid)+1 as pid from patient_data");
    $newpid = null;
    if ($result['pid'] >= 1) {
        $newpid = $result['pid'];
    }
    if ($newpid == null) {
        $newpid = 0;
    }
    return $newpid;
}
Should I send a commit, if yes please tell so I can do it.

amughalbscs16:
Fixes #1076 Introduced a modification in patient id generation function, so that the readability becomes better.


amughalbscs16:
Wrong Pull Request Needs some changes.

PR #1083, opened 2018-03-24
Country table filled by the code

2208Abhinav:
This is the code to fill country table in lists with country names and their codes.


sql/country_table.sql
```diff
@@ -0,0 +1,257 @@
+<?php
+
+include_once("../globals.php");
```
teryhill: Check this and see if it works from this directory.

2208Abhinav: Checking....

2208Abhinav: I had to change it to ../interface/globals.php

2208Abhinav: and extension to .php because I am running the script manually

teryhill: Yep

teryhill:
Give it a .php extension. Since it is actually a program and not just a sql script

2208Abhinav:
Changed the extension

2208Abhinav:
I  think this script should be put in such a way that it runs when the software is being installed

teryhill:
Let the user decide. if they need them all then they can add them. Don't force the user to do something. Like going to gitlab

2208Abhinav:
Okay I understand. I think you don't feel happy about the gitlab decision. :thinking:

PR #1084, opened 2018-03-26
Fix for #242: Notify when globals saved

### LINKED ISSUE #242: Some Frames need be configuration to Close on Select
Frames like the Patient Finder and Globals should close themselves after you take the end-action (like selection of the patient or "Save".

This probably could be configurable but maybe not necessary.

apooravc:
Fixes #242 
When the user clicks on save, and the edit_globals iframe is refreshed, there is no notification given to users. This PR adds a notification shown below which fades out after `3 seconds`:
image1

Works on the value of `$_GET['unique']` which is `null` on first opening edit_globals or refreshing it by clicking on refresh icon. It becomes equal to `yes` after clicking on save button.

### DIFF
--- a/interface/super/edit_globals.php
+++ b/interface/super/edit_globals.php
@@ -1,5 +1,5 @@
 <?php
-/** 
+/**
  * Edit Globals
  *
  * This program allows the editing of the system Globals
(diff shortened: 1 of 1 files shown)


teryhill:
Looks good and tests well.

interface/super/edit_globals.php
```diff
@@ -374,6 +374,16 @@ function checkBackgroundServices(){
 
 <body class="body_top">
 
+<!--Settings Saved Notification for Globals-->
+<div id="set_alert" style="width: 130px; background-color: #2ECC40; color: #fff; font-weight: bold; font-size: 17px; text-align: center; margin-bottom: 5px; display: none;">Settings Saved</div>
+<?php if ($_GET['unique'] == "yes") { ?>
+<script type="text/javascript">
+  $(document).ready(function() {
+    $("div#set_alert").css("display", "block").fadeOut(3000);
```
aethelwulffe: Very long fade.  No real harm.

PR #1085, opened 2018-03-28
Appointment time checking code added

2208Abhinav:
This will prevent appointment if time selected is absurd

### DIFF
--- a/modules/calendar/add_edit_event.php
+++ b/modules/calendar/add_edit_event.php
@@ -1678,6 +1678,16 @@ function SubmitForm() {
   return true;
 }
 
+$('[name="form_save"]').on('click', function() {
+    var hours = $('[name="form_hour"]').val();
+    var minutes = $('[name="form_minute"]').val();
+
+    if(hours > 6)   //the condition is just for testing purpose I will put the correct one after discussion
+    {
+        window.alert("Time is not valid");
+        $('[name="form_save"]').preventDefault();   
+    }
+});
 </script>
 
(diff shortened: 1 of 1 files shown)


2208Abhinav:
@teryhill @robbyoconnor please review this PR

2208Abhinav:
@teryhill I think we can make a table in database which will store the hours of service for clinics and use that data for condition inside **if**.

naveen17797:
screenshot from 2018-03-28 22-41-41
we could follow this kind of approach , we can let user set working hours for each day (like shown below)
screenshot from 2018-03-28 22-44-01

teryhill:
That sort of happens now. When you set up the system you should add "Provider" events in the calendar. These are Lunch, In time and out time. The system will pop up an alert stating no appoints for this provider if it is outside the in and our provider events. The issue I originally opened happens because the user is not heeding the alert warning.  Perhaps we should "Table" this until after GSoC .

But You guys are very creative with your ideas and suggestions, I like the fact that y'all are helping one another.

2208Abhinav:
@naveen17797 can you add your table part with my code.

2208Abhinav:
And tell me how can I help.

teryhill:
Sorry my use of the word "Table" means not work on right now. Let's see if the project for calendar get's accepted and then we can push the Ideas to the one doing that project.

2208Abhinav:
@teryhill that sound good

PR #1089, opened 2018-03-29
color picker improved

naveen17797:
Color picker can now change colors instantly in parent frame

## BEFORE 
screenshot from 2018-03-29 17-53-02

##AFTER
screenshot from 2018-03-29 17-53-33

@aethelwulffe @teryhill review this :relaxed:


aethelwulffe:
I will play with it later today.

PR #1091, opened 2018-04-01
Fix for appointments after 12 noon

teryhill:
Fix for appointments after 12 noon . The addition of 12 was not needed since the calculation was done previously.


apooravc:
@teryhill Oh, I didn't see the code at line `230` already adding `12` before. Thanks.

PR #1098, opened 2018-04-10
Refactoring and relocating MIPS module

aethelwulffe:


robbyoconnor:
Note: it's not mirrored bi-directionally. Only GitLab to GitHub. Other direction caused grief.

teryhill:
@aethelwulffe from the menu the "View Reports" , "Create MIPS Report" and "XML Editor" have broken links.

PR #1104, opened 2018-05-04
Profile picture feature for users

naveen17797:
screenshot from 2018-05-04 22-09-39
@teryhill done, need some info on how to add this to menu

### DIFF
--- a/interface/usergroup/user_info.php
+++ b/interface/usergroup/user_info.php
@@ -3,6 +3,9 @@
   include_once("$srcdir/sql.inc");
   include_once("$srcdir/auth.inc");
   include_once("$srcdir/headers.inc.php");
+  $uid = $_SESSION['authId'];
+  $query= sqlQ("SELECT picture_url FROM users WHERE id=$uid");
+  $r = sqlFetchArray($query);
   ?>
 <html>
   <head>
@@ -40,8 +43,13 @@
     </script>
   </head>
   <body class="body_top">
+<table>
+<tr>
+  <Td colspan="1">
     <span class="title"><?php echo xlt('Pass Phrase Change'); ?></span>
-    <br><br>
+  </Td>  
+</tr>
+</table>    
     <?php
       $ip=$_SERVER['REMOTE_ADDR'];
       $res = sqlStatement("select fname,lname,username from users where id=?",array($_SESSION["authId"])); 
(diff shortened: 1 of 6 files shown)


interface/usergroup/user_info.php
```diff
@@ -83,5 +122,103 @@ function(data)
 </HTML>
 <?php
   //  da39a3ee5e6b4b0d3255bfef95601890afd80709 == blank
+
+if (isset($_FILES)) {
+  //images will be saved with their uid
+  $uid = $_SESSION['authId'];
+    //MAKE THE UPLOAD DIRECTORY IF IT DOESN'T EXIST
+  if (realpath("../../profile_pictures/")) {
```
aethelwulffe: 755 permission if linux...

interface/usergroup/user_info.php
```diff
@@ -83,5 +122,103 @@ function(data)
...
+  //extension check done.
+  //if any validation needed be added, please add it below.
+  $bool = 0;
+  $target_file =  basename($_FILES["profile_picture"]["name"]);
+  $imageFileType = strtolower(pathinfo($target_file,PATHINFO_EXTENSION));
+  $verify_image = getimagesize($_FILES["profile_picture"]["tmp_name"]);
+  if($verify_image) {
+    $mime = $verify_image["mime"];
+    $mime_types = array('image/png',
+                            'image/jpeg',
```
aethelwulffe: Just need to know if it is an image...your code works, but getimagesize returns false for non images, so....
```
if(isset($_POST["submit"])) {
    $check = getimagesize($_FILES["fileToUpload"]["tmp_name"]);
    if($check !== false) {
        echo "File is an image - " . $check["mime"] . ".";
        $uploadOk = 1;
    } else {
        echo "File is not an image.";
        $uploadOk = 0;
    }
```

aethelwulffe:
Could always use this for printing ID cards...

teryhill:
@aethelwulffe have you been looking at MY notes?

interface/usergroup/user_info.php
naveen17797: already i have made the validation, if it is not a image then $bool = 0;

naveen17797: if $bool=0 then no file upload takes place, and no db records get updated.

interface/usergroup/user_info.php
naveen17797: windows doesnt need permissions it seems, are you working in windows?
i am using ubuntu it tests well here.

aethelwulffe:
I guess I should do a re-test soon.   Anyone else tested this yet (for a review)?

interface/usergroup/user_info.php
aethelwulffe: Good!

interface/usergroup/user_info.php
aethelwulffe: Gotcha

sql/2_0_0-to-2_0_1_upgrade.sql
```diff
@@ -117,3 +117,7 @@ ALTER TABLE `users` DROP `fullscreen_role`;
 DROP TABLE IF EXISTS `menu_trees`;
 
 DROP TABLE IF EXISTS `menu_entries`;
+
+#IfMissingColumn users picture_url
+ALTER TABLE `users` ADD `picture_url` VARCHAR(2000) NOT NULL AFTER `suffix`;
```
aethelwulffe: 2000?  Heh...uh...

teryhill:
Have tested and it works well.

Issue #1088, opened 2018-03-29
Unable to open patient in a new window


prabhakarpd7284:
Steps to recreate the issue : -
1. Select **Patient->Find Patient**.
2. Tick `Open in a new window` and select any patient.
![image](https://user-images.githubusercontent.com/23030978/38083616-ba625a40-3367-11e8-8107-ba69e1bb0d48.png)

It redirects to a blank page. 
![image](https://user-images.githubusercontent.com/23030978/38083659-dc066c68-3367-11e8-914f-8b6879086f71.png)

naveen17797:
@teryhill same issue i had yesterday. if this open in new window is enabled then the form must send post request to main_screen.php which resulted in a empty behaviour

aethelwulffe:
I swear that I responded to this earlier in the morning.  I asked about your global settings for "new window".  I hope no-one is messing with the repo re: the gitlab thingie.

naveen17797:
@aethelwulffe it is the reply for @prabhakar ?

teryhill:
@prabhakarpd7284  and @naveen17797 that "open in a new window" , does not open a new tab in the software but it opens a new tab in the browser. So it starts a new secession. It ,in my opinion, has no use in the current software.

teryhill:
@prabhakarpd7284 look in the error log and see if you are getting an error

prabhakarpd7284:
Nope, nothing is getting logged in`log` table.

teryhill:
@aethelwulffe is there a reason to keep this option. Since we are using the tabs I don't think opening another instance (session) is needed. I have always kept this turned off so the customers would not get any cross secession issues. @tmccormi

aethelwulffe:
Let's dump the option here.  We should change the behavior of overlapping tabs with a toggle control (maybe right on the header bar) to do "tab organization" things.

aethelwulffe:
Mangie,
Did you get this into a PR?

aethelwulffe:
Yes, you did, and it was merged.  Closing this.

PR #1105, opened 2018-05-04
removed the open in new window functionality #1088

Mangielijah:
Am sure this should remove the open in new window functionality
Open for feedback


teryhill:
@aethelwulffe yes dynamic finder is correct , the open in another window was on flowboard and dynamic finder. This PR is correct.

PR #1107, opened 2018-05-05
changing text color

naveen17797:
screenshot from 2018-05-05 06-57-11
@teryhill see this.


teryhill:
Good commit makes the information stand out.

PR #1116, opened 2018-05-12
Logging appointment delete in calendar

apooravc:
Issues which this PR deals with:
1) When user selects "Deleted" status and saves, status is logged but the appointment isn't deleted off calendar.

2) When user clicks on delete in add-edit-event panel, appointment is deleted off calendar but it is not logged in patient-tracker and patient-tracker-element tables.

Status "Deleted" is added in database.sql file. 

@teryhill This first commit attempts on addressing 2nd point. Currently, the issue with it is that "event_date" is passed as NULL upon clicking delete to manage-tracker-status which makes the control flow to go into block at line 204 instead of block at line 218 in patient_tracker and it results in 2 inserts (one in each table) made when clicking delete with one NULL apptdate value coming up in patient-tracker table instead of one update (into patient-tracker) and one insert (into patient-tracker-element). If I pass the date manually, it works by updating lastseq in patient-tracker table in correct id, eid, pid and creating a new record with corresponding id, status "Deleted" and seq updated.

### DIFF
--- a/library/patient_tracker.inc.php
+++ b/library/patient_tracker.inc.php
@@ -1,19 +1,19 @@
 <?php
-/** 
-* library/patient_tracker.inc.php Functions used in the Patient Flow Board. 
-* 
+/**
+* library/patient_tracker.inc.php Functions used in the Patient Flow Board.
+*
 * Functions for use in the Patient Flow Board and Patient Flow Board Reports.
-* 
-* 
-* Copyright (C) 2015-2017 Terry Hill <teryhill@librehealth.io> 
-* 
+*
+*
+* Copyright (C) 2015-2017 Terry Hill <teryhill@librehealth.io>
+*
 * LICENSE: This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0
-* See the Mozilla Public License for more details. 
+* See the Mozilla Public License for more details.
 * If a copy of the MPL was not distributed with this file, You can obtain one at https://mozilla.org/MPL/2.0/.
-* 
-* @package LibreEHR 
+*
+* @package LibreEHR
 * @author Terry Hill <teryhill@librehealth.io>
-* @link http://www.libreehr.org 
+* @link http://www.libreehr.org
 *
 * Please help the overall project by sending changes you make to the author and to the LibreEHR community.
 *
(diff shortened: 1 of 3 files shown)


apooravc:
Features implemented in these 2 commits:

1. When user clicks on Delete in event panel, lastseq is updated (increased by 1) in patient-tracker table (no new record added) corresponding to the selected appt. and a new record is added corresponding to the selected appt. in patient-tracker-element table with seq increased by 1 and status as "Deleted".

2. When user selects status as "Deleted" and clicks on Save in event panel, the corresponding appointment is deleted from libreehr-postcalendar-events and removed from calendar. The two table updates, mentioned in 1st, were already happening in this case.

Recurring appointments case:
When user makes a recurring appt., it is not logged in patient-tracker or patient-tracker-element (like a single/non-recurring appt. does with date/time/eid/pid etc.). It is logged in libreehr-postcalendar-events with eid.

In case of deletion, it is logged (as a new record with status "Deleted", seq as "1" in tracker-element and as a new record with lastseq as "1" in tracker table) only when:
1. User clicks on Delete, and selects applying changes to "all".
2. User selects "Deleted" status, clicks on Save, and selects applying changes to "all".

@teryhill please have a look.

teryhill:
It causes an error just an FYI

tmccormi:
would you be a bit more specific, Art tested it live on line with me before we merged and it worked fine.  You do have to run the sql_upgrade....

teryhill:
change a provider item not a patient
    On Wednesday, May 23, 2018, 7:36:52 PM EDT, Tony McCormick <notifications@github.com> wrote:  
 
 
would you be a bit more specific, Art tested it live on line with me before we merged and it worked fine. You do have to run the sql_upgrade....

tmccormi:
Error? Screen shot?  or are you just saying we should revert?

teryhill:
No to the revert. I will fix it in a commit I am doing to change the provider tab some. It only happens when you delete a provider event

image

teryhill:
oops I committed the fix directly to master. Sorry it was just a provider zero fix. @tmccormi

tmccormi:
you da man!

PR #1117, opened 2018-05-12
Integrating UI-Editor

2208Abhinav:
UI-Editor for the software

### DIFF
--- a/interface/themes/style.php
+++ b/interface/themes/style.php
@@ -21,6 +21,11 @@
 * returns html color code
 * @author Naveen
 */
+
+require_once("../globals.php");
+require_once("$srcdir/report.inc");
+require_once("$srcdir/patient.inc");
+
 function validate_html_color_code($arg) {
 	$color_code = "";
 	if (strlen($arg) == 6) {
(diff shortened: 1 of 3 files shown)


apooravc:
I tested it with my system @2208Abhinav in Fee Sheet, the positive type shows green as it should but after making it color red from the editor and saving it, upon refreshing of Fee Sheet tab, all buttons appear grey.
Before:
before
After:
after

Other things such as menu link and real time changes within the editor are working properly.

aethelwulffe:
This is due to a non-bootstrap button property (background).  /modules/MIPS/measures_select.php had the same issue (thus only worked with the purple theme, otherwise had grey and white css buttons), which is an issue I corrected and am going to push with new commits soon.  Removing the background property, probably declared in the fee_sheet/new.php somewhere should take care of this.  Bootstrap and all.

aethelwulffe:
Let me just say that all the UI tweaker bits we have been adding lately have been *very good to see*.  This has been a neglected area for a long time, and all this effort is making a huge difference.

naveen17797:
screenshot from 2018-05-16 19-20-58
Admin can now add,remove the users
The delieverables in this week , ie 21 may 2018 will be ui to set necessary settings and some other things as mentioned in proposal
@nileshprasad137

2208Abhinav:
@aethelwulffe sorry I was away from solving the bugs here for sometime, was busy with GSOC project. I will try to solve the bugs here now.

aethelwulffe:
We can hang onto this PR just fine.  It is non-critical, and you have to prioritize.

aethelwulffe:
Bump.
Folks looking at the Theme and UI stuff should really review this.

PR #1130, opened 2018-05-17
/custom directory cleanup.

aethelwulffe:
NOTE:  this is a housekeeping branch related to organizing the /custom directory (with the goal of losing it forever).  THUS, there are several very minor "features" that are touched here, and junk that is tossed out.

Feature "chart_tracker" is mostly contained in /reports.  Moved the /custom file utility that handles the feature there.  All of that feature should be moved out of /reports

Outdated and broken qrda and other related reporting features have been moved into the "boneyard" directory of the MIPS module...for now.  They will be replaced by current types of similar tools later.

Functions and configuration files aimed at extending the "prescription print to fax with doctor's name on it"  feature have been moved into the /sites/default directory right along side the other hylafax customization config and template stuff.  This feature extension only supports a single image signature, thus a single provider for ALL sites.  Not sure if it is a relevant feature at all, but it is moved to a more appropriate spot.

The old quicksearch filter extension was dumped.  We don't have the original patient_select.php doing the same jobs it used to.


tmccormi:
Merged on one review ha!

PR #1132, opened 2018-05-18
MigDinny patient_report accordion

aethelwulffe:
Uses accordion to make the patient reports cleaner.

### DIFF
--- a/interface/patient_file/report/patient_report.php
+++ b/interface/patient_file/report/patient_report.php
@@ -50,9 +50,8 @@
   <head>
     <?php
       html_header_show();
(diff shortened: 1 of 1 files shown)


tmccormi:
I'll test this and see.

aethelwulffe:
Any test results?
This might be getting left behind with the PDF changes (probably not), but this is a very useful piece for folks that have to view/export data...which is everyone.

apooravc:
Old interface:

ol1

ol2

New interface:

new1

new2

muarachmann:
roger that @apooravc

aethelwulffe:
Thank you.  I have thought so for a while.
- @MigDinny

Issue #1121, opened 2018-05-14
misc_billing_options and claim class fatal errors


aethelwulffe:
```
/interface/billing/billing_process.php

PHP Parse error:  syntax error, unexpected 'new' (T_NEW) in /opt/www/vhosts/assisted-living/openemr/interface/billing/billing_process.php on line 79, referer: https://libre.bowenmd.com/interface/billing/billing_report.php

if (isset($_POST['bn_process_ub04'])) {
$pdf =& new Cezpdf('LETTER');
$pdf->ezSetMargins(trim($_POST['ubtop_margin'])+0,0,trim($_POST['ubleft_margin'])+0,0);
  $pdf->selectFont($GLOBALS['fileroot'] . "/library/fonts/Courier.afm");
}
```
The above error with `$pdf =& new Cezpdf('LETTER');`
can be bypassed by removing the ampersand, to see the bigger error:

```
/interface/billing/billing_report.php

PHP Fatal error:  Uncaught Error: Call to undefined method Claim::miscOnsetDate() in /library/gen_x12_837.inc.php:516\nStack trace:\n#0 /interface/billing/billing_process.php(241): gen_x12_837('207171', '1824658', 'Generating clai...', false)\n#1 /opt/www/vhosts/assisted-living/openemr/interface/billing/billing_process.php(155): process_form(Array)\n#2 {main}\n  thrown in /library/gen_x12_837.inc.php on line 516, referer: https://Art_is_Great.com/interface/billing/billing_report.php

    if ($claim->miscOnsetDate() && ($claim->box14Qualifier()) && ($claim->miscOnsetDateValid())) {
    ++$edicount;
        $out .= "DTP" .       // Date Last Seen
            "*" . $claim->box14Qualifier() .
            "*" . "D8" .
            "*" . $claim->miscOnsetDate() .
            "~\n";
    } 
```
 
    Obviously, miscOnsetDate() is not defined in the code base (claim.class.php?).
  This was all changed in #1053, and looks like a partially complete merge.
  Deeper testing (meaning minimal actually) of the billing system in a production environment was required to hit this particular condition apparently.

aethelwulffe:
Not resolved:
From user:
We have two functions working with the /Fees/Billing and one that doesn't.
  CMS 1500 PDF generates a pdf,
   CMS 1500 TEXT generates a text document, and Generate X12 causes a white page with following error

[...]

PR #1133, opened 2018-05-18
Fix for the onset dated code

teryhill:
Fix for the onset dated code #1121


aethelwulffe:
Merging this, as it CANNOT get broke when it is broken.

teryhill:
Yes they were there to test with and never removed.

PR #1148, opened 2018-05-24
GSoC 2018 :  Calendar Improvements-Hyperlink the patient name

### LINKED ISSUE #791: In the Calendar when a user clicks on the patient allow the demographics tab to open.
Currently in the calendar when you click on the time slot with a patient assigned to the slot the add_edit_event opens. What needs to happen in addition to this is when the user clicks on the actual name the patients demographics screen needs to open.

apooravc:
Fixes #791 

A patient event slot in calendar has it's title (name) inside a `div` element with class `fc-title`. Using [eventRender](https://fullcalendar.io/docs/eventRender) which is triggered when an event slot is being rendered on calendar, all such patient slots/events (having `pid != 0`) can be selected, and their titles can be converted into corresponding links to patient's demographics. 

Thus, when the title link is clicked, demographics tab would open of corresponding patient, with their information displayed in patient data area (on upper-left and upper-right sides). If any other area is clicked in event slot, then calendar event panel would open, which is default behavior.

@teryhill @pri2si17-1997 Attempted a fix for this, please have a look. There is one issue with the opening of tab. If only calendar tab is open and you click on a title, it is not able to fetch the document upon first time click, there it opens a tab with 404 error. If you click again, then it is works as intended. Opening of tab here is based on how demographics tab opens when you click on a patient row in patient finder (block `118-134` in dynamic_finder.php). Couldn't get properly how `top.RTop.location` is used here.

### DIFF
--- a/modules/calendar/index.php
+++ b/modules/calendar/index.php
@@ -264,7 +264,8 @@ require('includes/session.php');
           if (event['pc_pid'] > 0) {
             //only when event is a patient event
             //event['pc_pid'] is number string for patient events, empty "" for provider events
-            var link = "../../interface/patient_file/summary/demographics.php?set_pid=" + event['pc_pid'];
+            //var link = "../../interface/patient_file/summary/demographics.php?set_pid=" + event['pc_pid'];
+            let link = '<?php echo $GLOBALS['webroot'] . "/interface/patient_file/summary/demographics.php?set_pid="; ?>' + event['pc_pid'];
             var titleLink = "<a href='#'>" + event['title'] + "</a>";
             //find all event title div elements
             var patientEventTitle = element.find('.fc-title');
--- a/modules/calendar/index.php
+++ b/modules/calendar/index.php
@@ -23,13 +23,13 @@ require('includes/session.php');
 
 ?>
 <html>
-<head>  
+<head>
   <link href='full_calendar/fullcalendar.min.css' rel='stylesheet' />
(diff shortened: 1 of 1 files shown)


teryhill:
Can you try it on Chrome? it does not work for me on chrome. It works as you stated on firefox

teryhill:
It causes issues when you select any thing other than 1 day.

aethelwulffe:
top.RTop.location...at very least the naming convention is due to the old frame view of OpenEMR.
-Sorry Terry, I got 99 problems son, but Chrome ain't one!  I keep thinking that I will get it for tests one day, what with us all enjoying GSOC so much, but I have a panic attack when I go to hit the download button. ref: https://futurism.com/google-dont-be-evil/

apooravc:
@teryhill Added a commit to fix change view bug, please have a look. One other issue with this is that if you have only calendar tab opened initially and click on a title link, it will open another tab with 404 error with link as `lh-ehr/interface/interface/patient_file/summary/demographics.php` causing error. Trying to figure out why it's caused.

apooravc:
Used `let link = '<?php echo $GLOBALS['webroot'] . "/interface/patient_file/summary/demographics.php?set_pid="; ?>' + event['pc_pid'];` to fix the 404 error @teryhill

teryhill:
This works well on chrome and firefox.

aethelwulffe:
Great job, nice function

PR #1149, opened 2018-05-24
Calendar Cancellation Reasons sql fix

aethelwulffe:
Special characters were used instead of normal single quotes in sql files.  passive fail on upgrades.

Noted during testing that the database menu inserts re-run no matter if they were done or no.   I think that was intended behavior though.


aethelwulffe:
Didn't have out weekly scrum, and this is an obvious fix issue getting stale.  Merging.

PR #1158, opened 2018-05-26
Add status delete alert in patient event panel

apooravc:
Added an alert so that users can be sure when deleting patient appointments by selecting status "Delete" and clicking save in event panel. Please have a look @teryhill

### DIFF
--- a/modules/calendar/add_edit_event.php
+++ b/modules/calendar/add_edit_event.php
@@ -1844,19 +1844,41 @@ function validate(valu) {
     <?php
     }
     ?>
-    $('#form_action').val(valu);
-
-    <?php if ($repeats): ?>
-    // existing repeating events need additional prompt
-    if ($("#recurr_affect").val() == "") {
-        DisableForm();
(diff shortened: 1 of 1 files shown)


apooravc:
Adding this for tracking appointment-delete-related PRs #1116

apooravc:
Added an alert _Cannot create an Appointment with status Deleted_ to prevent user from making an event with status "Deleted".

modules/calendar/add_edit_event.php
```diff
@@ -1844,19 +1844,48 @@ function validate(valu) {
...
+                        return false;
+                    }
+                <?php endif; ?>
+                return SubmitForm();
+            }
+            //if selected "Cancel" on alert
+            return false;
+        }
+        //alert user if creating appt. with status "Delete";
+        alert('<?php echo addslashes(xl("Cannot create an Appointment with status Deleted")); ?>');
```
aethelwulffe: new appt. view model could simply hide that status option...

PR #1159, opened 2018-05-28
Move appt. time check to correct place

apooravc:
The check for appointment time was initially added (#1073) in data layer of event panel code which is executed after form or (user selected) event details are submitted by clicking on save. The check for appointment time should be within `validate` function where check for all user input related errors are. @teryhill please have a look.

### DIFF
--- a/modules/calendar/add_edit_event.php
+++ b/modules/calendar/add_edit_event.php
@@ -249,21 +249,6 @@ if ($_POST['form_action'] == "duplicate" || $_POST['form_action'] == "save")
         if ($_POST['form_ampm'] == '2' && $tmph < 12) $tmph += 12;
         $duration = abs($_POST['form_duration']); // fixes #395
     }
-    //check for selected appointment time
-    if ($GLOBALS['check_appt_time'] == 1) {
-      //before starttime gets the values from form_hour & form_minute after clicking save,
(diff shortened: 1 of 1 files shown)


teryhill:
Tested . Works Great.

modules/calendar/add_edit_event.php
```diff
@@ -1821,18 +1806,44 @@ function are_days_checked(){
...
         return false;
     }
+
     //Make sure if days_every_week is checked that at least one weekday is checked.
     if(f.days_every_week.checked && !are_days_checked()){
         alert('<?php echo xls("Must choose at least one day!"); ?>');
         return false;
     }
 
+    //check if user selected appt. time is within clinic hours
```
aethelwulffe: Wish we were keeping system time for everything 24 hour and only displaying AM/PM stuff...and only when an idiot specifically goes out of their way to demand that the system use that.  "Do you really want to use AM/PM?"  Do you Really REally want to do this?"  "OK, if you really really really want to do this, please sign the waiver form below in bright red ink, and return it to us before the ink turns rust-colored".

-anyone care to build the time.inc.php library?

aethelwulffe:
NOT tested with AM/PM.  Just making sure nothing good is broke.
Thanks @apooravc  Apoorv

PR #1161, opened 2018-05-30
GSoC 2018 : Calendar Improvements-Event based refresh

### LINKED ISSUE #824: feature request: calendar refresh on event panel exit
Possibly a GCI project; I didn't see it in earlier postings.
Can we have the calendar refresh when exiting the calendar event panel?
Seems simple enough to just do that on 'Save' button click.
As it is now, when you make any changes to the calendar you either have to wait the refresh interval or manually click on the tab's refresh icon to see the changes, and know the change was successful.  And  IMO that's one too many unnecessary clicks to remember to do.
- HT

calendarrefresh

and my acct still can't add labels, afaict

apooravc:
Fixes #824 

There are 3 buttons in event panel which require calendar refresh after their respective processes are completed. These are- Save, Delete and Create Duplicate. These buttons have an event listener `click` attached on them so that when a user clicks on one of these buttons, the flow of control reaches blocks of  `$_POST['form_action'] ==`**`value`** in `add_edit_event.php` with  **`value`** of  `save`, `delete` or `duplicate` which comes from the corresponding button which was clicked. In this commit, I've added a function `refreshCalendar()` to these blocks which is called after the Save, Delete or Create Duplicate process is completed and refreshes the Calendar tab showing the changes made.

Besides these, there are some more events where a Calendar refresh may be needed. For example:
1. Adding a new facility or new user to Calendar.
2. Editing a facility or user in the Calendar.

These events should be considered for covering the cases where Calendar tab is also open along with Facilities or Users tab.  

I'm looking for more such events and will address them in my next commit.

@teryhill @pri2si17-1997  Please have a look.


modules/calendar/add_edit_event.php
```diff
@@ -193,6 +193,23 @@ function setEventDate($start_date, $recurrence){
 
 
  }
+
+ function refreshCalendar() {
+    echo "<script type='text/javascript'>";
+      echo "var docRoot = top.$('#mainBox');
+            //Calendar frame opened during initial loading of EHR
+            var calFrame1 = docRoot.find('iframe[name=" . "lst" . "]');
```
teryhill: What if calendar is not in the 1st? it could be in 2nd.

teryhill:
@apooravc  I have been looking at the code and why can't you do the refresh in this section that way you would not need to know which window.

`if ($_POST['form_action'] != "") {
  // Close this window and refresh the calendar (or the patient_tracker) display.
  echo "<html>\n<body>\n<script language='JavaScript'>\n";
  if ($info_msg) echo " alert('" . addslashes($info_msg) . "');\n";
  echo " if (opener && !opener.closed && opener.refreshme) {\n " .
       "  opener.refreshme();\n " . // This is for standard calendar page refresh
       " } else if(window.opener.pattrk){\n " .
       "  window.opener.pattrk.submit()\n " . // This is for patient flow board page refresh
       " };\n";
  echo " window.close();\n";
  echo "</script>\n</body>\n</html>\n";
  exit();
 }`

apooravc:
@teryhill Regarding your previous question, I've looked into how Calendar is opened/accessed. There are two cases:
1. Calendar is opened in one of two initially loaded frames upon first logging in EHR. Here, the user can choose the frame (left or right) in which Calendar is to be opened or not opened at all, upon logging in. The left iframe has `name = lst` and right iframe has `name = pat`.

2. Calendar is opened in a new frame by clicking on main menu item. This iframe has `name = cal`.

I've refactored `refreshCalendar` function to meet the above cases in the second commit.

apooravc:
@teryhill I thought making an independent event handler/function would be useful if we've to refresh Calendar frame from completely different frame, like the case of add/edit - facilities/users.

teryhill:
Okay good thought. I like it.

aethelwulffe:
Naturally, this depends on the name value of the Calendar tab staying "cal" which should be safe.
-On that note, before we go too very far down this, we should really look at the menu item names, as well as do a scrum on what "target" values the menu items should have.  We have been meaning to look at this for a while, but when most of the useful stuff in the system has a target of "rep0", it becomes impossible to do certain comparisons.  We need a baseline, then of course the menu editor update can happen.  "cal" needs to be preserved, and the use of it, and possibly the target "cal".

modules/calendar/add_edit_event.php
```diff
@@ -197,16 +197,37 @@ function setEventDate($start_date, $recurrence){
  function refreshCalendar() {
     echo "<script type='text/javascript'>";
       echo "var docRoot = top.$('#mainBox');
-            //Calendar frame opened during initial loading of EHR
-            var calFrame1 = docRoot.find('iframe[name=" . "lst" . "]');
-            var calSrc1 = calFrame1.attr('src');
-            if (calSrc1 !== undefined) {
-                calFrame1.attr('src', calSrc1);
+            //get references to all possible frames for calendar
```
aethelwulffe: All *current* possible cal references.

modules/calendar/add_edit_event.php
```diff
@@ -197,16 +197,37 @@ function setEventDate($start_date, $recurrence){
...
+            var leftFrame = docRoot.find('iframe[name=" . "lst" . "]'); //initial loading of EHR
```
aethelwulffe: These loadings are with the standard default, and are configurable.  Trying to determine if that is an issue at all.

apooravc: @aethelwulffe Yes, the calendar could be loaded into either frame (left or right) or not loaded at all upon logging in. Is that what you meant? I've taken cases for both in the function.

modules/calendar/add_edit_event.php
apooravc: Will do in next commit.

modules/calendar/add_edit_event.php
aethelwulffe: So long as we keep the standard that there will never be THREE default tabs (good idea in my opinion), and so long as we do another test with Terry's "Single Page Application" global setting, I think we are all on the same iFrame.  Good job.

apooravc:
@teryhill @aethelwulffe With the third commit, the calendar would also refresh when:

1. Adding a new facility or new user to Calendar.
2. Editing a facility or user in the Calendar.

So in above cases, Calendar tab is also open along with Facilities or Users tab.

apooravc:
With this, the user area and facility area are also covered along with events area in Calendar. I think these are the only areas (cases) in Calendar which can be changed from outside or inside the Calendar and hence need a refresh to view changes made.

aethelwulffe:
Verify if you are through with this PR please.

This is great by the way, but I hope we can get more done on access restriction settings.

interface/usergroup/facilities.php
```diff
@@ -40,6 +40,44 @@
 
 $alertmsg = '';
 
+function refreshCalendar() {
```
teryhill: Would suggest that you take this function and add it to /library/formsoptions.inc.php and call it from there.

teryhill: maybe use calender.inc

apooravc:
@aethelwulffe @teryhill Added `refreshCalendar` function to library. I'm through with this PR, final testing required.

aethelwulffe:
Does this mean we can be rid of the time-based refresh?

apooravc:
@aethelwulffe Oh you mean that fixed time period one? I think yes. All changes made to calendar, be it from inside (event panel) or outside (user/facilities tab) would be shown after a refresh which will happen just after the corresponding process causing the change gets completed.

teryhill:
What if another user updates a calendar event on another computer in another area of the building. Then will your screen refresh. not until you do something , right.

naveen17797:
No, teryhill js runs on the client side, so it won't

apooravc:
@teryhill The calendar tab on the another user's computer would refresh after user's done with updates. It will refresh on the same computer.

teryhill:
So does that mean that in order to ensure the refresh we still need the back ground process that will update the calendar?

apooravc:
The refresh process here makes the calendar tab iframe's replaced with its old value, thus refreshing itself. If the user makes any changes to events, the back ground processes which do the changes to DB would run upon refreshing, and would show the changed values. Am I getting your question right?

apooravc:
@teryhill As @naveen17797 said, the another user would see his/her changes after the process on his/her computer is completed. But for a user in a different building it wouldn't refresh, so there we may need the fixed time based refresh.

teryhill:
Currently there is a "timed" process that the user can select to run every 60 seconds, 5 minutes and 10 minutes. This refreshes the calendar screen so if another user on another computer makes a change to the calendar all users see it with in the time period.

To Test I use fire fox and chrome make a change to the calendar in fire fox and the change will appear in chrome after the time interval is up.

apooravc:
@teryhill Yes, I got your question. This process works for same computer (so same user). So we may need time based refresh to let all users see changes within a fixed time period.

apooravc:
@aethelwulffe This would cover multiple users making changes with a single value check, but yeah @naveen17797 it had to be done periodically.

naveen17797:
I mean the checking function, it has to run periodically?

apooravc:
I think for same computer, same user, this refresh process would just enhance UX that it refreshed itself after completing process and for multiple users making changes, I think the fixed time refresh would suffice?

aethelwulffe:
Yes, it has to run periodically, but it only has to check a single 
value...not just a numb-fingered total reload constantly. It seems like 
a table like this could handle a lot of different update states, like 
messages and everything else.

Not saying this is the best way to do it, but without a peer-to-peer 
thing going on, minimal messages to and from the server might be the 
best way to check a lot of things.

-Yeah, I know:  We should have built all this on top of the Unreal 
Engine....:-P

teryhill:
Yep @aethelwulffe  but I think the re-write is beyond the scope of this gsoc. I think we will just for now keep the back ground check . the user can turn it off if they want to. The flowboard does the same type of refresh.

apooravc:
@aethelwulffe Got your point. Yeah, the only way to do things in real time with PHP - do it periodically.

aethelwulffe:
OK.
Where are we on this @apooravc ?
Move forward, test again and merge?  Last test was since the last change.  This is approved, but we talked about a lot of different approach concepts, and such.

apooravc:
@aethelwulffe Should test again whether Calendar is refreshed on following actions:
1. When Save or Delete or Create Duplicate button is clicked in event panel.
2. When a new facility is added to Calendar or an existing facility in Calendar is edited with Calendar tab opened.
3. When a new user is added to Calendar or an existing facility in Calendar is edited with Calendar tab opened.

This PR adds an event based refresh which covers refresh for same user making changes on same computer (and hence, avoiding the need to manually refresh or wait for fixed time based refresh) and the already present fixed time based refresh covers refresh for multiple users making changes on multiple computers.

apooravc:
@aethelwulffe Yeah, If we pass name of iframe as a parameter to this function and change name to frameRefresh then it might be used for any frame refresh and placing the call to `refreshFrame(iframe_name)` at the right place would result in an event based any specific frame refresh.

aethelwulffe:
Would you mind implementing it that way (which will truly make it a core library function) and annotate the file for what it is used for so that the next dev to touch it uses it properly?  I think that would be a nice library addition, keep things more abstracted and all that.  It may be worth looking to see if it is something that can be integrated in a general library that always goes hand-in hand (not a guarantee), but just having this function sitting out there in a known location could be a future benefit.  We want smarter functionality like this across the board.

aethelwulffe:
The above is true even if we decide to not use this for the original purpose!

teryhill:
What about adding it to calendar.inc? instead of a new file.

apooravc:
@teryhill I tried adding it to calendar.inc but the html tags were causing a problem.

teryhill:
image

This is where I put it in the file.

teryhill:
Yep it works . been testing it since the first time I mentioned it.

aethelwulffe:
OK, I plugged the function in, dropped the other file and pointer, it (still) works, nothing crashes, and everything is contained in the calendar library.
-probably the best answer here.
@apooravc think you could take a moment and do a commit that way?  I would like to put this giant discussion list behind us and get this merged.

apooravc:
@aethelwulffe Moved the function to `calendar.inc`. Tested with 3 points mentioned above, works.

aethelwulffe:
That is it dude.

apooravc:
Another event added where a calendar refresh is needed: #1175 (A calendar global is changed and saved with calendar tab also being open)

aethelwulffe:
Cool.  We can continue on in #1175

PR #1163, opened 2018-06-01
Added the ability to display the status colors

teryhill:
This will display the status colors for the appointments unless the
status is None in which case it will display it in the appointment
category color.


teryhill:
The appointments on 5-31 display the appointment status colors. The ones on 6-1 display the category colors.

image

This is what it looks like in the flow board

image

apooravc:
Getting the color of status of an event and setting it as event slot color in Calendar when global is ticked. Tested, works on my machine.

PR #1164, opened 2018-06-02
This fixed some recurring errors.

teryhill:
This fixed some recurring errors. Removed a couple of commented out
lines of code. Added logic to create out of office based on the ending
time for in office. Fixed code so that days of the week works correctly
with the check box.

### DIFF
--- a/modules/calendar/add_edit_event.php
+++ b/modules/calendar/add_edit_event.php
@@ -228,7 +228,6 @@ if ( $eid ) {
                             FROM libreehr_postcalendar_events
                               LEFT JOIN facility ON (libreehr_postcalendar_events.pc_facility = facility.id)
                               WHERE pc_eid = ?", array($eid) );
-    // if ( !$facility['pc_facility'] ) {
     if ( is_array($facility) && !$facility['pc_facility'] ) {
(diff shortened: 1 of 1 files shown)


modules/calendar/add_edit_event.php
```diff
@@ -692,6 +709,14 @@ function setEventDate($start_date, $recurrence)
 
     // done with EVENT insert/update statements
 
+      if ($_POST['form_category'] == '2' && (!$eid)) { //adding category 3 based on category 2
+        $endtime = $cat3time;
+        $starttime = $cat3time;;
```
apooravc: There's a double semi-colon here.

teryhill: Good catch. I am sending a change to remove it.

modules/calendar/add_edit_event.php
```diff
@@ -692,6 +709,14 @@ function setEventDate($start_date, $recurrence)
 
     // done with EVENT insert/update statements
 
+      if ($_POST['form_category'] == '2' && (!$eid)) { //adding category 3 based on category 2
```
aethelwulffe: Hard coded categories need protection

PR #1166, opened 2018-06-04
Adding event description after hyperlink

apooravc:
Added event description text after hyperlink name in a Calendar event slot. With this, only clicking on hyperlink would open demographics tab and clicking on anywhere else would open event panel. @teryhill Please have a look.


teryhill:
Works great.

Issue #1169, opened 2018-06-05
Added back the ability to create a patient from the calendar


teryhill:
This was in the code before the calendar was upgraded . I just fixed it so it would work.

PR #1170, opened 2018-06-05
Added back the ability to create a patient

teryhill:
Added back the ability to create a patient from the calendar. This was
in the version before we upgraded the calendar just added it back. #1169

### DIFF
--- a/modules/calendar/add_edit_event.php
+++ b/modules/calendar/add_edit_event.php
@@ -1297,7 +1297,7 @@ echo '
             // refresh and redirect the parent window
             if (!opener.closed && opener.refreshme) opener.refreshme();
             top.restoreSession();
-            opener.document.location="../../new/new.php";
+            opener.document.location="../../interface/new/new.php";
             // Close the window
             window.close();
 </script>';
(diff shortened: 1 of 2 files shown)


modules/calendar/add_edit_event.php
```diff
@@ -1297,7 +1297,7 @@ function find_available(extra) {
             // refresh and redirect the parent window
             if (!opener.closed && opener.refreshme) opener.refreshme();
             top.restoreSession();
-            opener.document.location="../../new/new.php";
+            opener.document.location="../../interface/new/new.php";
```
aethelwulffe: Can we please change the "new" name some day in the future?

modules/calendar/find_patient_popup.php
```diff
@@ -134,6 +134,19 @@
...
+<script language="Javascript">
+            // Pass the variable to parent hidden type and submit
+            opener.document.theform.resname.value = "noresult";
+            opener.document.theform.submit();
+            // Close the window
+            window.self.close();
+</script>';
+}
+?>
+<!-- ViSolve: Verify the noresult parameter -->
```
aethelwulffe: ?

aethelwulffe:
Tested now.
I think we actually killed this by intent...but I can't find any evidence of that.

Issue #1152, opened 2018-05-25
Unknown directory asset 404 action


aethelwulffe:
\LibreEHR\interface\tags_filters\models\PatientTagEntry.php:

 line  16 :            ` new Column(array('width' => '70px', 'behavior' => new ActiveStatic(array('name' => 'tag_name', 'class' => 'tag_name', 'url' => $GLOBALS['webroot'] . '/interface/crisisprep/index.php?action=dispatch!element_changed')), 'title' => 'Tag Name', 'data' => 'T.tag_name')),`
 
Anyone know what the heck "crisisprep/" is/was?  This looks like code from some custom implementation.

aethelwulffe:
A search for "crisisprep" lands me here:  https://crisisprep.mi-squared.com/openemr/interface/login/login_frame.php?site=default
-Guess we know where this came from... :shipit:

aethelwulffe:
404 Not Found
       /LibreEHR/interface/tags_filters/javascript;: 2 Time(s)

-Any ideas?  Can't figure out what formed this bad request, where or when.  Something assembled it, 'cause there is no "tags_filters/javascript".

Maybe here(this looks like "javascript:" is an empty or local function call)?:
LibreEHR\interface\tags_filters\assets\bootstrap\css\bootstrap.css:
  210    }
  211    a[href^="#"]:after,
  212:   a[href^="javascript:"]:after {
  213      content: "";
  214    }

Or one of these (Line 9 does NOT look like a function call without the colon, right)?:
LibreEHR\interface\tags_filters\views\tags_demographics.php:
    8                          <a id="tags-edit-button" onclick="top.restoreSession()"
    9:                            href="javascript;"

teryhill:
@kchapple Can you help us out here?

kchapple:
I used the data tables backen framework from another project. Should be /interface/tags_filters/...

Issue #1147, opened 2018-05-24
TAGS plugin performance issues.


aethelwulffe:
When you have thousands of patients and you set a tag for each one of them to check if the user has access to them, loading of a patient selected from the search is very unresponsive.  For a patient with 40 encounters, on a server with stupid-huge resources, rendering the demographics page takes 69 seconds.

I am not sure if the performance hit is only when tags are used in a major fashion, or if there is a significant hit with large DB's simply if the system is turned on.

Issue #1140, opened 2018-05-23
TAGS/Filters links dead


aethelwulffe:
administration>tags
Create a tag.
Click on the tag color.
The requested URL /testinstance1/interface/tags_filters/javascript; was not found on this server.

marcusbow:
This seems to be true of Administration>Patients/Tags>status link as well.

Issue #1139, opened 2018-05-23
"Plugins" system menu weakness


aethelwulffe:
"Plugins" system, which currently encompasses only tags/filters, has a severe weakness.
Anyone that has ANYTHING in their ACL settings that allows access to Administration menu, the plugin system adds the tags admin links to the menu without further regard.  Thus, this is auto-generated on the global enabling of the tags plugin. 

 This means that anyone with "Clinician" or "Physician" access can change anything they want in tags/filters.  They can actually do a lot.

ref: interface/tags_filters/start.php

PR #1171, opened 2018-06-05
Fac. ACL attempt 4 (WIP)

aethelwulffe:
Trying to implement this facility based ACL feature for the fourth time.  This is the hang up that prevents 4 clinics I volunteer for (and another I sometimes work for) from switching to LibreEHR or needed to even implement an EHR.

Probably components missing yet, and the new calendar will cause issues, and apparently this already can't be merged for some reason, but this gives a starting point to try to complete this two-year-in-the-making priority.


aethelwulffe:
Conflicts resolved.

teryhill:
Loading it to my system to test. Any setup stuff I should know about?

aethelwulffe:
Terry, I haven't even tested it this go-round.  It looks at schedule facilities, I am not sure if the "use tags" global toggle even does any thing, and I really just pushed this to try to keep my own momentum.  I have had this branch for over a week, and I just can't seem to get any work done with all the interruptions, travel and cruddy little tasks popping up all over the place.

teryhill:
Give me the theory on how it is suppose to work.

aethelwulffe:
Create more than one facility.
Create more than one provider.
Create more than one patient.
Set the demographics field "Facility" of the patients to different facilities.
Create appointments and encounters for each patient.
Schedule appointments for different patients at their respective facilities.
Create a user that only has the admin>user schedule facilities set to one facility.
Log in as that non-admin user.
User should not see patients, encounters, or scheduled items for any patients that do not have the demographics field Facility set to the facility that they are set to.

The use:  Provide group restrictions on what users can see, specifically based on what clinics, locations, or treatment programs they are assigned to.

Original use case:
Central clinic with many providers.  All users at the central clinic have access to everything.
Remote locations where these providers visit (Home Health, Assisted Living, Group homes etc...) have home care providers or on-site medical staff.  These staff members need to be able to log into the system, do notes, review meds and all that, but should only every be able to see the patients at their own facility.  These are shared patients, but the staff must not have access to all the other patients, just because the doctor that sometimes visits them also services other locations.

aethelwulffe:
Issues in reverse order:
#1152 
#1147 
#1140 
#1139 
#1110 
#1109 
#985 
#982 
#977 
#969 
#678 
#39

aethelwulffe:
Izi refactor merging will cause problems if I do everything needed here, and Izimodal will help a lot of things for this too.  
Hope we can move on the modal change PRs soon.

teryhill:
Izimodal is already in the code base.  Are there specific files that you will conflict with?

aethelwulffe:
Not really, more like the user admin modal might need tweaks to the schedule_facilities drop-down stuff.  It is a nightmare to use for this purpose at the moment.

aethelwulffe:
-Context:  It is a nightmare to use when you have dozens of providers and a hundred facilities...which is the actual use case.

aethelwulffe:
Reference #1134 and #1136 being modal window changes on related stuff.

aethelwulffe:
Ref #1169 and #1170:
Report from user testing this against the PR in #1170:

Unfortunately, the attempted try to prevent users from seeing other patients doesn't seem to work.

The new ability to go to a patient's demographics page/patient chart from the calendar seems to be bypassing my tags that are set up. I'm able to view any patient on the calendar.  This is using the master branch.



Also, getting error message

[Tue Jun 05 15:39:58.182431 2018] [:error] [pid 909] [client 207.189.2.207:59950] PHP Parse error:  syntax error, unexpected end of file in libreehr/modules/calendar/add_edit_event.php on line 2173, referer: https://libre.bowenmd.com/interface/main/tabs/main.php?url=TAB_ONE_DEFAULT


Causes a empty square where the create calendar event should appear.


Running "Search or Add Patient" generates

ERROR: query failed: SELECT *, DATE_FORMAT(DOB,'%m/%d/%Y') as DOB_TS FROM patient_data WHERE 1 = 1 AND WHERE patient_data.pid NOT IN ( 5,14,16,17,21,45,48,49,55,60,74,78,80,116,117,131,152,163,170,178,184,201,213,232,269,290,335,350,353,357,360,379,420,436,440,446,450....
...205787,205788,205854,205971,206061,206062,206063,206669,206670,206924,206957,206958,206989,207302,207304,207306,207486 ) ORDER BY lname ASC, fname ASC LIMIT 0, 100

Error: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'WHERE patient_data.pid NOT IN ( 5,14,16,17,21,45,48,49,55,60,74,78,80,116,117,1' at line 1


libreehr/interface/main/finder/patient_select.php at 218:sqlStatement


There appears to be one too many wheres in the sql query.  I've removed it in my code and it now executes.



libreehr/interface/main/finder/patient_select.php


--- patient_select.php.bak      2018-06-05 16:10:00.962594063 -0400
+++ patient_select.php  2018-06-05 16:10:35.452689014 -0400
@@ -210,7 +210,7 @@
     if ( $where ) {
       $where .= " AND ";
     }
-    $where .= " WHERE " . $patient_filter;
+    $where .= " " . $patient_filter;
   }

   $sql = "SELECT $given FROM patient_data " .

teryhill:
@aethelwulffe are these errors from Libre? I thought the FACL was a WIP not an active product.

Don't see where the code fo adding back the create patient would cause an issue. I think it is more than that PR. You can remove that code from your test setup and run it to see what changes.

aethelwulffe:
@teryhill Heck, with tags active and everything else...I dunno what all it going on.  I can get back to this tonight and hunt those down again.

interface/tags_filters/start.php
```diff
@@ -86,4 +86,93 @@ function update_tags_filters()
 }
 add_action( 'update_plugin', 'update_tags_filters' );
 
-//}
```
teryhill: You need to un-comment line 5 and move this to the end of the file so it can at least be turned on and off with the global.

teryhill:
@sam-bowen does this functionality exist currently in your version of "Openemr"?

aethelwulffe:
This PR WORKS.

teryhill:
@aethelwulffe  great then what is the problem? I am confused.

aethelwulffe:
@teryhill the error was user reported.  Not validated.

I have not tested interaction with patients via the calendar yet, moving to that now.  You need to run the sql_update, then set the globals like so :
image
in Admin/users, edit your test user to set the schedule facilities list to a subset of the facilities.
Set some patients to one facility (in dems), then others to different ones.  Check the patient list after logging out and logging in as the restricted user.  This works with an admin user as well, and the admin user can change their access, but must log out and log in to see the results.

aethelwulffe:
@teryhill I was just stating "WORKS" because I was happy with the latest test.  Woohoo and all that.

teryhill:
What did the latest test net you? What section works?

aethelwulffe:
[Tue Jun 12 18:58:48.248773 2018] [:error] [pid 10972:tid 1980] [client ::1:51609] PHP Parse error:  syntax error, unexpected end of file in C:\\xampp\\htdocs\\branchtest\\modules\\calendar\\add_edit_event.php on line 2173, referer: http://localhost/branchtest/interface/main/tabs/main.php?url=TAB_ONE_DEFAULT
When scheduling an appointment, I get an error here.

Took a long time to find, but looks like line 261 has the if update condition with an unclosed bracket.

aethelwulffe:
Can't test the appointments right now due to the above error, but I can at least verify that the patient_select functions as intended.

teryhill:
I just pushed the fix to master. No pr on that one sorry.

aethelwulffe:
No problem!
Geeze that one was killing me.  Every branch I looked at was identical, and I had not gotten to the blame history yet.

aethelwulffe:
Those time thingies have a lot of re-declarations in the add_edit file.

teryhill:
Hopefully we can redesign/rewrite it after GSoC. My head is swimming from the changes that need to be made.

aethelwulffe:
image
As you can see, while you can't use patient select to see the patient, you can get to them with the calendar, much the same as you can when a tag restricts you.  Same missing do_action hook.  The modal window for appointments loads at least.
I just got to hunt down where the hook was in the old add_edit for tags and the like.

A Oemr version where this worked is in:  https://github.com/SunCoastConnection/MI2BPUC
(private repo, but @teryhill has access to it.

teryhill:
The do_action hooks is missing in the bpuc version. Do you want to know where to put them in that version or where they are in the Libre version?

sam-bowen:
Terry, there is a lot of functionality that is being discussed so I am unclear on which functionality you are specifically  referring to but will assume you are speaking of the facility - privacy function.

The facility / privacy function in our current oemr is great.  The calendar works with the facility privacy feature.  Users are restricted from all calendar elements unless specifically given permission to see a facility or group of facilities.

Typically for facility permissions to become active the user has to log out and then log back in for the permissions to take effect.

sam-bowen:
> The do_action hooks is missing in the bpuc version. Do you want to know where to put them in that version or where they are in the Libre version?
> 

Our version works with the calendar.
The LibreHealth version does not work with the calendar.

The facility - privacy appears to be working with non-calendar related functionality but has not extensively tested by me personally.  Marcus in our organization has done more testing and is reporting positive results.

aethelwulffe:
This branch has been updated with the calendar fix. (merged master).

aethelwulffe:
@teryhill I am trying to hunt down where the hook comes in for the calendar.  The BPUC branch/fork/repo had the facility stuff put in BEFORE the tags system.  They don't have tags, and the commit that brought the stuff in was pretty messy;  hard to track what was going on.
  I am digging in again.  If we fix this, then it should also fix it for the tags.
  Next question will be the flowboard.

sam-bowen:
How much of a slow do you think is going to result from using the tags?  So far the implemented tags seems to be major drag on performance.  Sam Bowen, MD

aethelwulffe:
Yeah, that is a known thing.  Nonetheless, I put in a thingie so they can be turned on (they are off by default).
  We are discussing ripping out the tags code, and probably the plugin engine completely until something functional actually uses it.  I made an attempt to consolidate the components, but they are very difficult to track down and uncouple from the standard libraries.  Once it is working, I will see about getting the facacl stuff implemented as simple functions called when/where they are needed as conditional statements.

aethelwulffe:
@sam-bowen we won't be using the TAGS at all.  The performance hit is why we went back to this simpler approach.

sam-bowen:
How is the code doing now at enforcing the facility privacy rules for your test database?

aethelwulffe:
Working tonight to try to find a good place to make the function call within the new calendar.  Some help from javascript code-monkeys would help a lot.

apooravc:
@aethelwulffe I'd like to help with Calendar. Most of the calls in fullCalendar happen from `modules/calendar/index.php` such as `eventClick`, `eventRender`.

aethelwulffe:
PLEASE!  Apoorv, I am seriously suffering here.  This is a feature that should have existed in the EHR since OpenEMR started in 1998.  I am harassed from all angles working on too many things, and this is a problem that I can never seem to get solved.  You would not believe that number of clinics this would be useful to.  Unfortunately, all of them seem to be rural shoestring budget type places that can't afford anything fancy, so they try to use our software!

apooravc:
That's great to hear. Regarding this issue, the functions added in `interface/tags_filters/start.php` have to be called from within Calendar but when they should be called? Which events or user actions should call these functions?

teryhill:
@apooravc when you code this in make sure you put it on one of the gloabls that @aethelwulffe wants to use so it can be turned off for the users that don't need it.

aethelwulffe:
@apooravc Invited you as a collaborator so you can see the repo above.  It is some very old code.

The original commit that added the feature:
https://github.com/SunCoastConnection/MI2BPUC/commit/cb0e7dcf5c4d02d87e1694e60d37f97e8c03901b

In the code where (part) of the feature was implemented to function in the old calendar:
https://github.com/SunCoastConnection/MI2BPUC/blob/a14b05ed715233a9f6851ff738f0115c3e6eb1e4/interface/main/calendar/modules/PostCalendar/pnuserapi.php#L1000

@teryhill the feature does not need an additional global check.  When do_action() fires, the start.php handles the inclusion/exclusion of the functions.  That said, this can be implemented in any manner.  We can even rip the functions out and add them in a much more simple way outside of the "plugin" thing.  It is not maintained, the Tags are not recommended for use due to performance issues, and it is shotgun code.  It is a great concept, and it was cleverly implemented, just not easy to utilize.  I believe that if you write a module for the application, you should be able to spam hooks (documented in the code) wherever you want, and have that PR merged without argument.  The code itself need not have an engine disbursing it all over the application though.  It just needs to be in an isolated directory, and have a registration function to include it.  Menu items and configuration can be run at registration, in the exact same manner as installing an encounter form (or maybe a little slicker than that).

teryhill:
@aethelwulffe what progress have you made? 

If I understand the issue , "You need the patients to "Not" be displayed on the calendar unless the facility 
 That is in their record in patient_data is selected and that facility has to be one in the user table that is assigned to that user" is this correct. I am trying to get a work item so we can maybe do a surgical strike.

teryhill:
@apooravc I think this could be done in get_provider_events.

In old code:
C:\AlphaFiles\bidenis\REPOS\MI2BPUC\interface\main\calendar\modules\PostCalendar\pnuserapi.php:
998 "a.pc_eventDate <= '$end')) ";
999
1000: $patient_filter = do_action( 'filter_patient_select_pnuserapi', $_SESSION['authUser'] );
1001 if ( $patient_filter ) {
1002 $sql .= " AND ".$patient_filter;

sam-bowen:
Having used this a lot in the old oemr code base, I don't mind "admin getting burned".  I think a better solution is to require selection of a facility when each new patient is being added to the patient_data table.

Thank you @apooravc  for looking at this.  We really need this feature in our distributed multi-facility (110 facilities) practice.

Sam Bowen, MD

apooravc:
@aethelwulffe On top of my list now. Working on it.

aethelwulffe:
Let me know what you need from me.  100% access is all I can offer here.
-gotta look into why the e-signature stuff is messing up now (argh).

apooravc:
@aethelwulffe Wanted to check if I understand the issue correctly. So, there are multiple users and multiple facilities and if a user logs into EHR and Calendar opens up, we want that user to see only those appointments which are at facilities that the user has access to, right? Is this the only goal or is there something else too?

aethelwulffe:
pseudo code:

if patient_data.facility NOT IN (SELECT (field array) FROM 
schedule_facilities WHERE user = $user){User should not be able to get 
to that patient or even see their name anywhere in the system;}

sam-bowen:
Each individual may have 1 or more facilities assigned.  Some of the users (practice employees) will need access to all of the facilities.  Sam Bowen, MD

aethelwulffe:
Individual meaning "user" above.
-eventually we want to turn the patient.data.facility into an exploded array so it is possible to assign them to more than one facility, but that is page2.

apooravc:
Opened a PR as an extension of this PR to cover Calendar related aspects of facility ACL.

aethelwulffe:
OK, big post coming up...I really care about this piece! :)

I think you are on to something.  My guess is that you need not check the patient_data table at all, just the facility that the even is scheduled to, then only display those events.  

I could be wrong, but it looks like your code parses the whole patient_data table to get a list of PID's to check against.  That would work, but is probably very slow.

Perhaps it is not iterating through the whole list of PID's and only the patients that are displayed.  If so, then that is fine, and better in some cases than if it is only looking at the facility they were scheduled at (and no more checks than would otherwise happen.  Tell me what you think.  I know that the TAGS stuff makes pretty heavy queries, and becomes unusable when there are a lot of patient encounters.  Pretty sure the original version of this just checked the event facility.

Another nice feature would be to display a shaded out box where there is a non-visible appointment, so as not to cause double booking based on thinking it is an open schedule.

In Mental Health, and in Behavioral Health for foster children care or group homes, the "Facility" is often a service program (like "Counseling" or "Case Management").  In those situations we are using "Service Programs" instead of "facility", but the service programs still map to a "Facility" in a technical way, so this is still OK...it is just that the child (or sibling, or parent) might be in multiple service programs.  Access control has to change a lot, is very complex, and very very important in this.  More fine grain control would be very useful, but only if there is also a very simple to use version (like based on single facility as we are doing here) that is good for the other 90% of the cases and isn't touchy to work with.

On your number 1 above (and number 2) I think I have a good idea what the facility toggle user interface should look like, and I think it should basically replace all user configuration items except the simple modal window user creation.
  This includes the "User facility specific settings" which may not be applicable anymore (it was for an old billing requirement some ten years out of date and is NOT related to what we are doing here) and the admin edit of the address book.
  There should be somewhere that has ALL user information, and displays an accordion view of everything based on what type (user vs. contact vs. referring provider) they are.
  That is outside the purview of calendar development though, and within my capacity to work on, so I will concentrate there while you forge ahead with the GSOC items (not that you have not done a ton already)

teryhill:
@apooravc in the get provider events do a check of the "User" allowed facilities (Scheduled Facilities). Then check the facility in the $fetchedEvents array and if they don't match "continue" in the foreach loop

teryhill:
@apooravc  once you get them not displaying we will look at this requirement

 "Another nice feature would be to display a shaded out box where there is a non-visible appointment, so as not to cause double booking based on thinking it is an open schedule."

apooravc:
Since, the action hook `do_action('filter_patient_select_pnuserapi', $_SESSION['authUser'])` returns `" pd.facility IN ( user_schedule_facility_id_string ) "` which has a list of scheduled/allowed facilities for that logged in user, I initially thought to compare an event's facility using `$event['pc_facility']` with this list of allowed facilities (let's say this focus-to-facility option) but since a facility column was also added in patient_data and there was a `pd.facility` in SQL string, I went with this focus-to-patient option in my PR which first extracts a table of PIDs (from patient_data or pd) where facility is in allowed facility string and then makes an array of these (allowed) PIDs to check against any event's PID.

Since, a patient can now be given a facility (say facility_A) in patient's demographics facesheet, is this case possible where that same patient has events at a couple of other facilities (some of which are not scheduled/allowed to current logged in user)? Because then according to this focus-to-patient algo, that patient is allowed to user (assuming facility_A is in scheduled facilities of user) meaning user can also see patient's events at those facilities which are not in scheduled facilities of user.

I think that's why comparing event's facility directly with scheduled/allowed facilities of user (focus-to-facility option) would be better. I would do it like this- first extracting facilities (from facility table) allowed to user using `" pd.facility IN ( user_schedule_facility_id_string ) "` (have to remove pd.facility from string), make an array of such facilities and this array would be compared with an event's facility in `foreach` loop, sounds right? @aethelwulffe @teryhill Would queries with facilities be better than queries with patients in terms of speed?

teryhill:
@apooravc yes focus-to-facility. Hold off for right now on the patient data facility. I will give this some more thought but the goal for right now is to display or hide the patient based on the scheduled facility matching the "scheduled facilities" in the user field.

My initial thought on the patient_data_facility is that they should not be allowed to be scheduled at any facility except the one they are assigned to. This would be accomplished in the "find_patient.php" program and I think it already contains the "Hooks" 

First is to accomplish the focus-to-facility.

sam-bowen:
I am not as convinced that a grayed out box is going to work.  The problem s that we frequently double book appointments on purpose.  I personally do this because I use a 13.3 inch screen laptop and need two or three appointments per line to preserve real estate screen.  We use these more as appointment lists instead of appointment times.  I know our practice is different from others and input from other others of this feature would be appreciated.  Sam Bowen, MD

teryhill:
@aethelwulffe is this a need for the behavioral folks?

"Another nice feature would be to display a shaded out box where there is a non-visible appointment, so as not to cause double booking based on thinking it is an open schedule."

 It would seam to me that if the user does not have access to that facility , then the worry for "Overbooking" does not exist, because they are scheduling for the facility they only have access to.

sam-bowen:
This issue as I see it, the Clerk A will be only be able to see Facility A schedule.  This clerk with then see a list of appointments that is either grayed out or not but will only fill in the non-grayed out schedule.  I think this will encourage a list one appointment wide.  When what we need is two or three columns wide.  Maybe the solution is to have a feature to turn 1, 2, or 3 columns?

i would prefer to get the basic facility-privacy feature working first before worrying about columns.

Sam Bowen, MD

sam-bowen:
The overbooking is something that I use intentionally as a feature.

sam-bowen:
A number of family practices use this as a feature to have their patients show up in small groups, A practice that is referred to as "shingling". 3 patients on the hour, 2 patients at ten after, and 1 patient 20 minutes after the hour.  The pattern repeats every thirty minutes.

Sam Bowen, MD

teryhill:
@sam-bowen  Every provider I know uses the "Overbooking"  Feature. 8^) > We will try to enhance that feature.

aethelwulffe:
@apooravc  Apoor, I think you are on the right track.  Looking at the event facility solves most of the edge cases of use, and is the lightest approach as well.  We don't want them to see anything not happening at their facility, and the rest of the goings-on are taken care of by the other placed hooks.  There may be other reports (and the report builder) that need to take this feature into account, but this is the simplest approach to victory.
-----------------------Off Topic--------------------------
-On the rest of the calendar, even just reducing (specifying) the font to use for the provider name labels would help.  Really needs a horizontal overflow (which can't really be done anymore I think), AND/or better yet, a horizontal view switch.  With even ten providers in the view, it is useless.  An option could be to provide a paged view (LIMIT) that can be configured, then you can state how many providers should be on your view etc... and page around instead of using an x-scroll.  x-scroll is really only convenient when it is replicated at the top of the view as well as the bottom, and has sticky label rows.  Those all suck up space.  A view scroll arrow (vs slider bar) could be implemented, but all of that is like video-game level viewport interface controls.  I would love to have a mouse_right_down panning control in something like that.
-just musing....

apooravc:
@aethelwulffe The accommodation of large no. of providers is also on my list [here](https://forums.librehealth.io/t/project-improving-the-calendar-system-in-ehr/2342/3). I'll have a look at it.

apooravc:
Above commit shifts code (and hence, facility ACL feature) from this branch to #1189.

aethelwulffe:
@sam-bowen The real code is now at https://github.com/LibreHealthIO/lh-ehr/pull/1189 #1189.  Closing this.

PR #1172, opened 2018-06-08
Fix #1162 Not able to change fields in globals

### LINKED ISSUE #1162: Not able to change fields in globals.
I cannot change some of the fields in globals. I checked it on the demo site and I get the same issue. For example I am changing calendar end time and it will not save the new entry.

image

I think it is an issue with the field validation.

apooravc:
Fixes #1162 

Some of the global fields in Administration > Calendar section were not saving changes. Corresponding updates were not taking place in DB. After examining where flow of control went while program was running, it was found that solving the issue required a couple of syntax changes in the _elseif_ chain in validation block from line `258` to `309`.  @teryhill Please have a look.


teryhill:
@aethelwulffe  and @tmccormi  I tested this and it works. I am going to merge it to test @naveen17797 's updater.

PR #1175, opened 2018-06-10
Admin based calendar refresh

apooravc:
Making Calendar refresh itself whenever a change to it is done, either from inside or outside Calendar. This covers the case when a Calendar global in `Admin > Globals > Calendar` is changed and saved with Calendar tab also being open along with  _Edit Global Settings_ tab. @teryhill Please have a look.


aethelwulffe:
I think this also runs the refresh when ANYTHING is changed in globals, which should be fine.

PR #1134, opened 2018-05-20
Refactoring Facilities for iziModal

apooravc:
Added iziModal to Add Facility and Edit Facility. Some issues still coming up:
1. Edit Facility modal opens upon clicking on a facility link repeatedly. Issue associated with fetching of dynamic links.
2. Edit Facility modal isn't getting closed after clicking Cancel.

One way to see the changes _locally_ would be as follows:
1) Clone my repo via `git clone https://github.com/apooravc/LibreEHR.git`
2) After cloning, only the master branch would be available locally. Enter `git checkout -b iziModal-facilities origin/iziModal-facilities` to get the branch _iziModal-facilities_ locally.
3) The changes could then be seen on facilities page.

@teryhill @muarachmann Please have a look.

PR #1176, opened 2018-06-10
GSoC 2018 : Calendar Improvements-Audit feature for Appointments.

### LINKED ISSUE #529: Audit features for appointments.
The request from the user is to see the users that have touched the appointments, They want to know who rescheduled an appointment and when. They want this information to be accessed from the demographics screen.

I have added the original user(person who created the appointment) and the users who changed the status  to the Patient Flow Board report .

apooravc:
Fixes #529 

Initial structure/interface commit for _audit feature_ for appointments. Three files for interface are added in first commit; these are select_patient.php (tab), track_appointments.php (tab) and appointment_log.php (modal). Working on back-end process for recurring, non-recurring appointments and interface upgrades wherever needed. @teryhill @aethelwulffe @pri2si17-1997 Please add relevant tags if applicable.


apooravc:
The audit feature interface with sample data could be viewed in two ways:
1. Through Demographics tab > Track Appointments link.
2. Through Patient/Client sub menu > Track Appointments option.

A _Track Appointments_ tab would open in either of above two ways and clicking on a row in this tab would open the _log window_.

teryhill:
@apooravc any progress on the conflict with the modal close and updater?

apooravc:
@teryhill I think I found the code (function) which was causing that issue.

There are two types of calling iziModal:
1) When we do not pass a dynamic link for it to open. Eg: `iframeURL: "facilities_add.php"`.
2) When we pass a dynamic link for it to open. Eg: `iframeURL: "facility_admin.php?fid=" + link`, where `link` is a variable.

In second case, we use `initIziLink(link)` function in which there is a function which executes whenever such an iziModal with dynamic links is closed:

`onClosed: function () {`
                      `setTimeout(function () {`
                          `parent.$(".fa-refresh").click();`
                      ` }, 300);`
`}`

So, the log window modal was a case of an iziModal with dynamic links and so was another facilities edit modal which I worked upon earlier in #1134. I tested it with updater and closing it showed same results as log window in this case. So, I think closing of any such iziModal with dynamic links using above function in `onClosed` property would open updater.

This function refreshes all tabs that are open when such modal is closed after a delay.

teryhill:
Great Job figuring it out !! How should iziModal be used to prevent the issue?

apooravc:
@teryhill One way would be removing that function but when I was working with @muarachmann on iziModal refactoring, he said that it was necessary to refresh tab after iziModal closes otherwise upon re-opening some other `link`, it opens data of previous `link`, which I tested to be true.

teryhill:
Ask the group and see if they have any thoughts. Is the issue just the updater? or will this be a bigger issue?

apooravc:
I will ask on group. I'm still looking for a connection b/w tabs getting refreshed event and updater modal opening event. I'm not sure but I think issue is with use of iziModal in updater. Looking into it.

apooravc:
@teryhill Got it, if you inspect element on updater-icon, you can see the `<i class="fa fa-refresh"></i>` which is used for updater's icon from font awesome library. Now, the tab refresh icon also has `<span class="fa fa-refresh tab-button" data-bind="click: tabRefresh" title="Refresh"></span>`. They both have common class `fa-refresh` because they both have same refresh type icon. 

So, when we close iziModal, that above function with property `onClosed` finds all such elements on screen with `fa-refresh` class which in this case includes all refresh icons of tabs which are open as well as updater icon and then clicks on all of it using `parent.$(".fa-refresh").click();`.

Eventually we are not only directing all tabs to refresh upon closing of iziModal but also updater to open up!

teryhill:
Chance the updater to something else and see if that changes the outcome.

apooravc:
I did, I removed the `fa-refresh` class from updater icon and it removed its icon and upon closing of iziModal, it didn't open up. 

One way of preventing this would be to use another class to identify all such refresh icon on tabs like `class = "tab-refresh-icon"` (adding in tab icon code) and using `parent.$(".tab-refresh-icon").click()` instead of `parent.$(".fa-refresh").click();`.

aethelwulffe:
Not a review of the code, just some label recommendations for context usage:
1.  If you have a title of  "Track Appointments" and you have a a date, you can just call the date column "Date" vs. Appt. Date.
2.  "Appointment Track" at the level you dive to, can just list "Moved On" or Created, and does not need the appt. in that.  We know the context, and titles are there for a reason.  You only use line-item identifiers when you need to pull data raw out of a database (id, field_id, field_value) sort of things.  Otherwise, the context is obvious, and the result is cleaner.

In your log, make sure you 1.  Really need to center and not left-right/right-left justify, and that you look at it with a very long user name.
Finally, PLEASE DO NOT USE mm/dd/yy or dd/mm/yy dates.  Ever.  😉   YYYY/mm/dd everywhere we go, unless a triple-checked "Do you really want to use confusing regional date scheme with your vital patient records?" has been asked many times.  Europe likes day-month-year.  US likes month-day-year.  Geeks, computers, military, federal agencies and anyone with any foresight uses numerically sorting yyyy-mm-dd.  In medical usage, the "Good Documentation Practices" mandates by most professional organizations at least mention that as an aside.

apooravc:
PR #1194 fixing conflict b/w iziModal and updater.

aethelwulffe:
Conflict on that darn menu popped up here.
-It would either be in the JSON or in the DB sql, one of the two!

apooravc:
@teryhill Attempted to cover non-recurring appointments in latest commit. Please have a look.

Track Appointments tab displays data using server-side processing with DataTables framework with source at track_appointments_ajax.php. 

Track appointments tab:
track-appt

Log window:
log-window

Two main queries are used to get data.
Query and resultant table used in Track Appointments tab:
main-table

Query and resultant table used in Log window:
log-table

interface/main/finder/select_patient.php
```diff
@@ -0,0 +1,179 @@
+<?php
+// Copyright (C) 2012 Rod Roark <rod@sunsetsystems.com>
```
aethelwulffe: Dunno why this file is showing up as new in this branch.

interface/patient_file/history/track_appointments.php
```diff
@@ -0,0 +1,249 @@
+<?php
```
aethelwulffe: Same here.

aethelwulffe:
I guess I should pull this down for a test, yeah?  I will get on that today.

interface/main/finder/select_patient.php
apooravc: Used Patient Finder as Select Patient in this PR with only difference being that former leads to demographics tab of a patient and latter leads to track appointments tab of a patient. Is there a need to change the header info.?

interface/patient_file/history/track_appointments.php
apooravc: This is a new file and is interface for track appointments tab. Will add header info.

teryhill:
@apooravc on the status change, can you put the description in the message instead of the symbol?

teryhill:
image

Make it say reminder done in addition to the *

apooravc:
Latest commits add:
1. "changed by user..." in case of status change and status description in logs.
2. A global `$GLOBALS['gbl_appt_list_page_size']` for no. of appts. displayed per page in appt. list.
3. A range filtering in case of appointment dates.

apooravc:
Last commit fixes an order issue in Log window. The order of change records is important and addition of a new `JOIN list_options ON option_id = status` in second last commit caused resulting table for Log window to be in wrong order as shown below:

Before addition:
correct

After addition:
incorrect

There are three things to understand what happens in Log window background process:
1. If a date changes, a new id (and patient_tracker_id) is added as a row in patient_tracker (and in patient_tracker_element) tables.
2. If a time changes, a new id (and patient_tracker_id) is added as a row in patient_tracker (and in patient_tracker_element) tables.
3. If status changes, no new id is added as a row in patient_tracker table but a row is added with same id as before the change in patient_tracker_element table (under patient_tracker_id column) with seq column be `+= 1` for that id.

A sample case with patient_tracker_element table (see **42** and **53** in patient_tracker_id column) is shown:
why

This means that date and time changes can be displayed in correct order using id of patient_tracker and then status changes can be displayed in correct order using seq (as illustrated by yellow markers in above image) of patient_tracker_element, both in ASC order. @teryhill Please have a look.

apooravc:
@teryhill Added time format and license headers. Please have a look.

aethelwulffe:
Retesting at 17:00 today.  Gonna apply this on top of the izimodal fix I tested this morning (and without).

teryhill:
@aethelwulffe I am good with a commit on this and then we can see what needs to be changed/fixed else where.

aethelwulffe:
Had to switch my testing data.
Updater worked good here too with the latest changes manually added.  Merging.

PR #1179, opened 2018-06-14
Set Calendar time zone 

apooravc:
This PR adds a global `calendar_timezone` which the user could use to select his/her time zone for Calendar date/time functions. [date_default_timezone_get](http://php.net/manual/en/function.date-default-timezone-get.php) (used as default time zone for Calendar) gets the value of `date.timezone` in php.ini file if it is valid/supported (valid time zones given [here](http://php.net/manual/en/timezones.php)) otherwise defaults to UTC. The user selected time zone from Main globals or User specific globals (given more preference when different values) is set as `date.timezone` configuration option value using [ini_set](http://php.net/manual/en/function.ini-set.php). 

`date` function is used to fill entries in `date` & `start_datetime` columns in patient_tracker and patient_tracker_element tables respectively. @teryhill Yet to add other time zones ([list here](http://php.net/manual/en/timezones.php)). Should I add all of which are given there? Or go for a particular set covering major regions in time zones?

### DIFF
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -129,6 +129,7 @@ $USER_SPECIFIC_GLOBALS = array('default_tab_1',
                                'print_next_appointment_on_ledger',
                                'calendar_view_type',
                                'calendar_refresh_freq',
+                               'calendar_timezone',
                                'check_appt_time',
                                'event_color',
                                'pat_trkr_timer',
@@ -1523,7 +1524,7 @@ $GLOBALS_METADATA = array(
        'bool',                          // data type
        '1',                             // default
       xl('Use the Appointment Status Colors in the Calendar Instead of the Appointment Category Colors.')
-    ),    
+    ),
 
     'calendar_refresh_freq' => array(
       xl('Calendar Refresh Frequency'),
@@ -1535,6 +1536,19 @@ $GLOBALS_METADATA = array(
       ),
        '360000',                     // default
       xl('How often the calendar automatically refetches events.')
+    ),
+
+    'calendar_timezone' => array(
+      xl('Calendar Time Zone'),
+      array(
+        date_default_timezone_get() => xl('Default - php.ini value'),
+        'Asia/Dubai' => xl('Dubai'),
+        'Asia/Kolkata' => xl('Kolkata'),
+        'Asia/Karachi' => xl('Karachi'),
+        'Asia/Pyongyang' => xl('Pyongyang'),
+      ),
+       date_default_timezone_get(),    // defaults to php.ini "date.timezone" value if set valid otherwise UTC
+      xl('Set calendar time zone.')
     ),
 
     'calendar_provider_view_type' => array(
--- a/library/globals.inc.php
+++ b/library/globals.inc.php
@@ -83,6 +83,28 @@ if (stristr(PHP_OS, 'WIN')) {
   $backup_log_dir      = '/tmp';
 }
 
+//------------------------------------
+// for creating $zone_opt_arr at line 1565
+
+// concatenating '!' to avoid Default option from not coming in list
+// when php.ini value is same as an item on zone_list
+$zone_opt_arr = array( '!' . date_default_timezone_get() => xl('Default - php.ini value') );
+
+// iterating to get time zones in form of, for eg:
+// 'Asia/Kolkata' => xl('Kolkata'),
+$zone_list = timezone_identifiers_list();
+foreach ($zone_list as $item) {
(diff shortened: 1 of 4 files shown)


teryhill:
look at the demo files at full calendar. get-timezones.php, get-events.php and utils.php.

library/globals.inc.php
```diff
@@ -1537,6 +1538,19 @@
       xl('How often the calendar automatically refetches events.')
     ),
 
+    'calendar_timezone' => array(
+      xl('Calendar Time Zone'),
+      array(
+        date_default_timezone_get() => xl('Default - php.ini value'),
+        'Asia/Dubai' => xl('Dubai'),
+        'Asia/Kolkata' => xl('Kolkata'),
```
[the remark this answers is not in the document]
aethelwulffe: I knew that! :)

apooravc:
In the third commit:

1) Moved global check code from `library/patient_tracker.inc.php` to `interface/globals.php` to cover time zones all over EHR.

2) Added reference to `globals.php` from `patient_tracker.inc.php` file.

Note about `ini_set` function: "Sets the value of the given configuration option. The configuration option will keep this new value during the _script's execution_, and will be restored at the script's ending." Referring here to those scripts which will use `date` functions which depend on time zones. Needs to be tested wherever time is involved. @teryhill Please have a look.

aethelwulffe:
Good idea.  This stuff has been on todo lists for a while.  One place for time stuff, and only one place.

apooravc:
@teryhill Also, this PR was first intended to be used to set time zone for Calendar only but now it can be applied to whole of EHR. So, should I let global option for setting the time zone remain under Calendar sub-section or should I shift it to somewhere else (and if shifting, then where)?

teryhill:
Move it to under the Systems tab

teryhill:
Also look at moving your function from globals.inc.php to edit_globals.php.

apooravc:
@teryhill The code creating `$zone_opt_arr` right? Is it causing errors?

teryhill:
@apooravc it is not causing errors. It is just "Cleaner" to have it outside of globals.

library/patient_tracker.inc.php
```diff
@@ -19,6 +19,7 @@
 *
 */
 require_once(dirname(__FILE__) . '/appointments.inc.php');
+require_once("../interface/globals.php");
```
teryhill: is this needed?

apooravc: Yes, the code in globals.php uses `ini_set` set the selected time zone in Globals as time zone for the script in which globals.php is included. So, wherever date/time functions, this should be included in script containing them.

teryhill:
It is called in a program that includes this one.

apooravc:
@teryhill You mean that globals.php is called in edit_globals? Yes, but edit_globals only lets a user save his/her global option which could be then accessed by `$GLOBALS['calendar_timezone']`. Suppose we're saving an appointment, then `patient_tracker.inc.php` is executed too and without including globals.php here, the time zone for `date` function inside `manage_tracker_status` would not be saved according to user saved option.

teryhill:
The time zone information is in globals.php, correct. Globals php is a main program that is loaded with the main code . Look at main.php, it is loaded there, look at patient_tracker.php , it is loaded there. patient_tracker.inc is used by patient tracker. Globals is called by the "Top" programs.

apooravc:
Changes done in last two commits:

1. Removed `include globals.php` from patient_tracker.inc.php as it is included in `main.php` itself.
2. Changed global name from `calendar_timezone` to `ehr_timezone` as it now covers time zone all over EHR.
3. Moved time zone option list from Calendar to System tab in Admin > Globals and also moved code from `globals.inc.php` to `edit_globals.php`.

@teryhill Please have a look.

teryhill:
This option is a user specific item also @aethelwulffe so if the user is in another time zone they can set there system for the different zone. Needs to have the / put back in so the locations are easier to determine.

apooravc:
@teryhill Added back `/` to options so that user may find it easier to navigate to a time zone. One more thing I see is that after moving global from Calendar to System tab, the time zone option list now only appears in Admin > Globals > System and not also in User Preferences.

teryhill:
You need to add the "Systems" tag to the users.

image

apooravc:
@teryhill Thought so, on it.

apooravc:
@teryhill Added System tab to Edit User Settings. Please have a look.

apooravc:
@aethelwulffe These changes are final.

Issue #479, opened 2017-03-08
Log review interface.


aethelwulffe:
The Administration->Logs feature, which lets you do searches in the logs by user etc...  should not do anything except display the interface when you navigate to it.  Currently, it executes queries, sometimes crashing marginal systems with a lot of data in the logs without the user ever having selected any search criteria.  Often times, this is due simply to accidentally selecting the wrong item, getting logs and having to wait until the system starts responding again.

navendu29:
Can I work on this Usability feature ?

aethelwulffe:
Yes you may, of course.  To assign you, I will need to add you to the collaborator list.

aethelwulffe:
Are you a GSOC hopeful, or do you just want to be added to the standard team invitation?

navendu29:
Yes, I intend to apply for GSOC .

aethelwulffe:
Then you may accept the invitation I sent you to that team, and welcome aboard!

navendu29:
Thank you for the invitation.

navendu29:
What type of interface do u want? 
Should the interface only contain the log information from the log tabel ?

aethelwulffe:
Ha!  I don't want an interface.  The community does.  That means you get to decide and design it, in light of it's common usage case....
  So, what is this log table used for?  

- It is used to track activity by users, sort of a "blame" track that can be used to identify a series of events.
- A source for extracting every query that has been made and use that do things like recreate the data after someone has deleted it.  On this piece, it would be nice if the log.comment section kept the queries clean and accessible instead of prepending them with other strings.
- It can be used to find where some types of upgrade or imports have happened for troubleshooting purposes.
Now, in regard to the above, and the fact that at the moment the log interface has a very small footprint, it would be nice if it got improved without turning it into a huge project unto itself, with a whole bunch of assets that only it uses.  While the criteria selection and display are a bit clunky, the main issue is about the fact that the page takes forever to load because it tries to do things right when you click on the menu item...sometimes taking up massive resources and freezing the server...which is the main objective of the issue.

navendu29:
I am getting a trouble in implementing this feature. I think i need to study more regarding this for which i would need some time so , if anyone else wants to work on this feature he/she may proceed .

tmccormi:
A pull request is coming from @growlingflea that is the baseline.  It's an Activity Log viewer we created for Dr Bob Hoyt.  Uses DataTables.

Looks like this
1) Default  TODAY
[...]

PR #1180, opened 2018-06-14
Report libre ehr user Activity 

growlingflea:
Funding by Dr Bob Hoyt for NHANES education project, Reference to issue #479 

This  reports the amount of time that a user has been logged in.  The time is calculated by either going from login to logout (forced or not) or in the case the user closes the browser and doesn't logout, the time from login to the last activity before the following login.  

The report allows the administrator to enter a start time and a end time, it also allows the user to filter the data table using a search field.

Here is an example of the summary
image

Here is an example of the logins.
image


tmccormi:
The bulk of new feature includes updated version of datatables (which Find Patient uses) and will be the basis of a clinic research tool to replace Reports->Clients->Clinical.

This can also be the basis of a better logviewer tool.

teryhill:
Why are the assets in /library and not /assets? There is a version of them in assets.

tmccormi:
I think this should be a standard report in interface/reports/ IMHO - datatables is meant to be a shared resource (it should be in assets/ for sure)  and be a upgrade of what is already there (assume the current version of datatables is in assets already.

aethelwulffe:
yeah, datatables is there, and if this stuff actually requires all 191 files to do that with (care was taken to only bring in the minimum for other purposes) then fine.  It is a shared asset clearly, and even if it is not shared, then it is still third party and belongs in /assets.
I like the idea of dumping the whole rest of the shebang into one directory under /interface/reports, provided it has been pared down of purely extraneous stuff we will never use.  My thoughts with that vs. /modules is how it all fits into the Report Generator cleanup (already happened...and keeps me from going there and messing with anything during GSOC).

  Personally I think it would live more happily in modules as a self-contained thing that does not really compare to the rest of that content, but either argument is reasonable.  That said, I believe @Trodrige should have the opinion that actually matters.

aethelwulffe:
IMO
250,000 lines of code added for this report, so it bears looking at.

I understand that this was commissioned by a third party, developed outside of the public project, and then kindly contributed.  That is great and awesome, but the ultimate use case != footprint.
I also understand that the export bits are elaborate, and account for a lot of the code.  Refactoring to use existing assets might be too much trouble.  As a matter of fact, there may be no interest at all in devoting the time needed to bring it to project standards.  This is understandable.  We have had other things in the same class we were simply not able to bring in, though we are grateful.

To get this moved in:  If the originator can put in the effort, that is fine.  If not, we can keep this PR and work the branch when we can to pare it down and get it contained.  If it really winds up using all those assets, all the .css, all the images and everything and is still a big horking thing, *and* we can justify it to someday tie it to log admin, timesheets or something really useful, then it should be a module...than is optional in contribs...for now.

1.  All the ajax stuff and all that should go into a one-piece directory with all the rest used by this.  It isn't core functionality, and we just spent a couple of years in cleanup.
2.  If it is added to the contribs repo as a module, that is great, but if it is otherwise bringing in things like pdf generators (major project was devoted into cleaning that up already) then it should really be a "Hey, gimmie this bit extra thing.
-Not making any use of any existing dev norms we are trying to support, so while not ungrateful, we still gotta look this in the mouth a bit.

tmccormi:
Agreed that it needs to be trimmed and refactored into assets (and or a module), but the idea here, that I have talked about a lot, is that we could be using Datatables for more the the patient finder, and I shared examples of other work we did using them.  This is meant to be a basis for a new standard for simple pre-created reports that have a lot of easy to implement functionality.

growlingflea:
OK.  I made the changes and rebased with master.  Let me know if there is anything else I should do.  I have plans on showing a breakdown in activity in the future.

teryhill:
Thanks @growlingflea . @aethelwulffe the assets are now under assets. I did a quick look over the code and will test it tonight.

interface/reports/username_report.php
```diff
@@ -0,0 +1,251 @@
...
+ * LICENSE: This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://opensource.org/licenses/gpl-license.php>;.
```
teryhill: @growlingflea can this be re-licensed under MPL

interface/reports/username_report.php
```diff
@@ -0,0 +1,251 @@
...
+ * Copyright (c) 2018 Growlingflea Software <daniel@growlingflea.com>
+ * File adapted for user activity log.
+ * @package ppc-db
```
teryhill: Should be LibreHealth EHR

interface/reports/username_report.php
```diff
@@ -0,0 +1,251 @@
...
+ * @author  Tony
```
teryhill: You

interface/reports/username_report.php
```diff
@@ -0,0 +1,251 @@
...
+@import "<?php echo $GLOBALS['webroot'] ?>/assets/js/datatables/media/css/demo_page.css";
+@import "<?php echo $GLOBALS['webroot'] ?>/assets/js/datatables/media/css/demo_table.css";
+@import "<?php echo $GLOBALS['webroot'] ?>/assets/js/css/jquery-ui-1-12-1/jquery-ui.css";
+<!-- @import "<?php echo $GLOBALS['webroot'] ?>/library/css/jquery.tooltip.css"; -->
+.mytopdiv { float: left; margin-right: 1em; }
+</style>
+
+<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/assets/js/datatables/media/js/jquery.js"></script>
+<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/assets/js/datatables/media/js/jquery.dataTables.min.js"></script>
+<script type="text/javascript" src="<?php echo $GLOBALS['webroot'] ?>/library/js/jquery-ui-1.8.21.custom.min.js"></script>
```
teryhill: Is this version required or will 3.1.1 work

interface/reports/username_report.php
```diff
@@ -0,0 +1,251 @@
...
+
+
+
+
+
+</script>
+</head>
+<body class="body_top formtable">&nbsp;&nbsp;
+<form action="./username_report.php" method="post">
+<label><input value="Show Session Details" type="submit" id="details_selector" name="show_all" id="show_all"><?php ?></label>
```
teryhill: is <?php ?> needed at the end of the line. Should "Value" be translated Both of these lines

interface/reports/username_report.php
```diff
@@ -0,0 +1,251 @@
...
+    $from_date = date("Y-m-d");
+
+} else {
+    $from_date = fixDate($_POST['form_from_date'], date('Y-m-d'));
+    $to_date = fixDate($_POST['form_to_date'], date('Y-m-d'));
+}
+
+if ( !$_POST['form_to_date']) {
+    // If a specific patient, default to 2 years ago.
+    $to_date = date("Y-m-d");
```
teryhill: Can you make the dates follow the system setting for dates  "Date Display Format" in globals

interface/reports/username_report.php
```diff
@@ -0,0 +1,251 @@
...
+?>
+
+    <input type='text' name='form_from_date' id='form_from_date' size='10'
+           value='<?= $from_date ?>' >
+
+    <input type='text' name='form_to_date' id='form_to_date' size='10'
+           value='<?= $to_date ?>' >
+
+    <input hidden id = 'summary' value = '<?= $_POST['show_summary']  ?>'>
+    <input hidden id = 'details' value = '<?= $_POST['show_all']  ?>'>
```
teryhill: Using the short php tags in the above statements.  Is it needed?

interface/reports/username_report.php
```diff
@@ -0,0 +1,251 @@
...
+			<th><?php echo xla('Session Time'); ?></th>
+		</tr>
+
+	</thead>
+	<tbody id="users_list">
+	</tbody>
+</table>
+</body>
+<link rel="stylesheet" href="../../library/css/jquery.datetimepicker.css">
+<script type="text/javascript" src="../../library/js/jquery.datetimepicker.full.min.js"></script>
```
teryhill: datetimepicker is actually in the assets directory use jquery-datetimepicker. the other on is a hold out from the old code.

interface/reports/username_report.php
```diff
@@ -0,0 +1,251 @@
...
+
+<script>
+    $(function() {
+        $("#form_from_date").datetimepicker({
+            timepicker: false,
+            format: "Y-m-d"
+        });
+        $("#form_to_date").datetimepicker({
+            timepicker: false,
+            format: "Y-m-d"
```
teryhill: again could you follow the system setting for dates.

library/ajax/username_report_ajax.php
```diff
@@ -0,0 +1,193 @@
+<?php
+/**
+ * library/ajax/claims_viewer_ajax.php Ajax file to process imported X12 filesgstat
```
teryhill: is this correct?

library/ajax/username_report_ajax.php
```diff
@@ -0,0 +1,193 @@
...
+ * LICENSE: This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; either version 3
+ * of the License, or (at your option) any later version.
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://opensource.org/licenses/gpl-license.php>;.
```
teryhill: Can you use MPL?

library/ajax/username_report_ajax.php
```diff
@@ -0,0 +1,193 @@
...
+    if($testing)
+        return " Limit 0,10";
+    else return "";
+
+}
+
+//This function will find the date-time of the last
+function getLastSessionActivityTime($user, $date){
+    $query = "Select date from log where user = '$user' and date < '$date' order by date desc limit 0,1";
+    $result = sqlStatement($query);
```
teryhill: sql binding?

library/ajax/username_report_ajax.php
```diff
@@ -0,0 +1,193 @@
...
+    $row = sqlFetchArray($result);
+    return $row['date'];
+
+}
+
+function getSessionTime($user, $date){
+	//get the datetime of the next login.
+	$query = "Select min(date) as date from log where date > '$date' and event = 'login' and user = '$user'";
+
+    $result = sqlStatement($query);
```
teryhill: sql binding?

library/ajax/username_report_ajax.php
```diff
@@ -0,0 +1,193 @@
...
+function getSumSessionsByUser($user, $from_date = '', $to_date = ''){
+    $sum = 0;
+    //get the list of dates and put it in an array
+    $query = "select date from log where event ='login' and user = '$user' ";
+        if($to_date != '')
+            $query .= " and date >= '$from_date' ";
+        if($from_date != '')
+            $query .= " and date <= '$to_date' ";
+
+        $result = sqlStatement($query);
```
teryhill: sql Binding?

library/ajax/username_report_ajax.php
```diff
@@ -0,0 +1,193 @@
...
+    if(isset($_POST['from_date']) && strlen($_POST['from_date'] > 9)){
+        $qstring .= " and date > '{$_POST['from_date']}' ";
+    }
+
+    if(isset($_POST['to_date']) && strlen($_POST['to_date'] > 9)){
+        $qstring .= " and date < '{$_POST['to_date']} 23:59:59' ";
+    }
+
+    $qstring .= ifTestingTrue($testing);
+	$result = sqlStatement($qstring);
```
teryhill: Sql Binding?

interface/reports/username_report.php
tmccormi: Yes

growlingflea: Fixed

interface/reports/username_report.php
growlingflea: Fixed

interface/reports/username_report.php
growlingflea: Fixed

interface/reports/username_report.php
growlingflea: Fixed

interface/reports/username_report.php
growlingflea: Fixed

interface/reports/username_report.php
growlingflea: Fixed

interface/reports/username_report.php
growlingflea: Fixed

interface/reports/username_report.php
growlingflea: Fixed

interface/reports/username_report.php
growlingflea: Fixed

library/ajax/username_report_ajax.php
growlingflea: Fixed

library/ajax/username_report_ajax.php
growlingflea: Fixed

library/ajax/username_report_ajax.php
growlingflea: Fixed

library/ajax/username_report_ajax.php
growlingflea: Fixed

library/ajax/username_report_ajax.php
growlingflea: Fixed

growlingflea:
OK.  Ready for a re-review.  Sorry for the delay in fixing this as I've been pretty busy.  More reporting is on the way.

growlingflea:
@teryhill @aethelwulffe Should I close this pull request then recreate one?

tmccormi:
No need to redo the P/R

aethelwulffe:
Hi Daniel, sorry took so long.  ISP just tried to put me out of business by being a lazy evil monopoly.

-Trying to see a commit message for something new.  Says just one commit.

PR #1183, opened 2018-06-17
Managing vacation/holidays in bulk

apooravc:
This first commit adds the interface for this feature where user can enter holidays, vacations in bulk without going to the dates on the calendar to do it. Calendar now has an _add_ button as shown below:
add

Clicking on this button would open a dialog modal having two tab views/forms:

Add Vacation form
vac

Functionality:
In this tab, user can select a range of consecutive dates of vacation for a selected provider from _Start Date_ to _End Date_. Upon submitting this form, vacation slot would be booked for specified dates and specified provider.

Add Holiday form
holid

Functionality:
In this tab, user can select and _add_ disjoint dates of holiday (which will be shown under _Selected dates_ column) for a particular clinic/facility. Upon submitting this form, holiday slot would be booked for specified dates and all providers related with that clinic/facility.

@teryhill Please have a look.

### DIFF
--- a/modules/calendar/add.php
+++ b/modules/calendar/add.php
@@ -20,7 +20,7 @@ if ($_POST['form_action'] == "vacation_submit") {
   $end_date = $_POST['endDate'];
   $fetchedEvents = fetchAllEvents($start_date, $end_date); // returns an array of events b/w start and end dates
   if (!empty($fetchedEvents)) {
-    // don't make requested event slots
+    // don't make requested event slots,
     // alert user and close "Add Vacation/Holiday" dialog box
     $alert_close_box = '<script type="text/javascript">
                           alert("There are already some events b/w requested start and end dates. Please try again.");
@@ -91,16 +91,16 @@ if ($_POST['form_action'] == "vacation_submit") {
 
 // 2. Clinic Holiday dates
 if ($_POST['form_action'] == "holiday_submit") {
-  // check if there is an appt. (either recurring or non-recurring) on any of selected holiday dates
+  // check if there is an appt. (either recurring or non-recurring) on at least one of selected holiday dates
   $dateAndType = json_decode($_POST['jsonData'], true);  // convert JSON string to an associative array
   foreach ($dateAndType as $date => $type) {
     // check each selected date
     $fetchedEvents = fetchAllEvents($date, $date); // returns an array of events on $date or including $date (in case of recurring appts.)
     if (!empty($fetchedEvents)) {
-      // don't make requested event slots if there's at least 1 conflicting date
+      // don't make requested event slots,
       // alert user and close "Add Vacation/Holiday" dialog box
       $alert_close_box = '<script type="text/javascript">
-                            alert("There are already some events b/w requested start and end dates. Please try again.");
+                            alert("There are already some events on at least one of the selected dates. Please try again.");
                             var addDialogBox = top.$(".dialogIframe"); // select dialog box element
                             var windowCloseBtn = addDialogBox.find(".closeDlgIframe"); // find close button element
                             windowCloseBtn.trigger("click"); // simulate "click" event on button
--- a/modules/calendar/add.php
+++ b/modules/calendar/add.php
@@ -6,6 +6,7 @@ require_once("$srcdir/patient.inc");
 require_once("$srcdir/formatting.inc.php");
 require_once("$srcdir/calendar.inc");
 require_once("$srcdir/encounter_events.inc.php");
+require_once("$srcdir/appointments.inc.php");
 
 $dateFormat = DateFormatRead();
 call_required_libraries(array("jquery-min-3-1-1","bootstrap","datepicker"));
@@ -14,6 +15,23 @@ call_required_libraries(array("jquery-min-3-1-1","bootstrap","datepicker"));
 
 // 1. Provider Vacation dates
 if ($_POST['form_action'] == "vacation_submit") {
+  // check if there is an appt. (either recurring or non-recurring) b/w (or on) start & end dates of vacation
+  $start_date = $_POST['startDate'];
+  $end_date = $_POST['endDate'];
+  $fetchedEvents = fetchAllEvents($start_date, $end_date); // returns an array of events b/w start and end dates
+  if (!empty($fetchedEvents)) {
+    // don't make requested event slots
+    // alert user and close "Add Vacation/Holiday" dialog box
+    $alert_close_box = '<script type="text/javascript">
+                          alert("There are already some events b/w requested start and end dates. Please try again.");
+                          var addDialogBox = top.$(".dialogIframe"); // select dialog box element
+                          var windowCloseBtn = addDialogBox.find(".closeDlgIframe"); // find close button element
+                          windowCloseBtn.trigger("click"); // simulate "click" event on button
+                        </script>';
+    echo $alert_close_box;
+    exit();
+  }
+
   $event_date = $_POST['startDate'];  // Vacation starts at this date
   $event_end_date = $_POST['endDate'];  // Vacation ends at this date
   $starttime = "{$GLOBALS['schedule_start']}:00:00";
@@ -68,10 +86,30 @@ if ($_POST['form_action'] == "vacation_submit") {
   // Insert events - Vacation event starts at some date and repeats till end date,
   // covering Calendar slots over entire clinic hours under selected provider.
   InsertEvent($args);
+  refreshCalendar(); // after "Add vacation" process is complete
 }
 
 // 2. Clinic Holiday dates
 if ($_POST['form_action'] == "holiday_submit") {
+  // check if there is an appt. (either recurring or non-recurring) on any of selected holiday dates
+  $dateAndType = json_decode($_POST['jsonData'], true);  // convert JSON string to an associative array
+  foreach ($dateAndType as $date => $type) {
+    // check each selected date
+    $fetchedEvents = fetchAllEvents($date, $date); // returns an array of events on $date or including $date (in case of recurring appts.)
+    if (!empty($fetchedEvents)) {
+      // don't make requested event slots if there's at least 1 conflicting date
+      // alert user and close "Add Vacation/Holiday" dialog box
+      $alert_close_box = '<script type="text/javascript">
+                            alert("There are already some events b/w requested start and end dates. Please try again.");
+                            var addDialogBox = top.$(".dialogIframe"); // select dialog box element
+                            var windowCloseBtn = addDialogBox.find(".closeDlgIframe"); // find close button element
+                            windowCloseBtn.trigger("click"); // simulate "click" event on button
+                          </script>';
+      echo $alert_close_box;
+      exit();
+    }
+  }
+
   function createHolidayEvent($key, $value, $prov_id) {
     // $key is event date as a YYYY-MM-DD string
     // $value is day type string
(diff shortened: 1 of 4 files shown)


teryhill:
I am guessing that this is just a UI since it did not put anything on the calendar. 

The Holiday will apply to the entire clinic , it should also allow 1/2 days. There needs to be a check on both functions to make sure no appointment is scheduled for any of the requested dates.

aethelwulffe:
Can you do the sanity check for existing appointments that have not been canceled/moved?  Should be good otherwise.

apooravc:
@aethelwulffe You're referring to the check which Terry mentioned right? Check to make sure no appointment is scheduled for any of the requested dates,, before forming Vacation/Holiday slot on them?

aethelwulffe:
Yes.  That thing.  Should be at least a warning.  I can totally see this scrumming up a well meaning admin, while causing the front desk folks to want to kill them....and some pretty upset patients.  At very least, it is a nice additional tool for ways to FIND blocks of time to schedule the holiday.  We have checks like that already I believe.

teryhill:
@apooravc  it put the vacation on the scheduled still not closing the popup, you are getting there.

apooravc:
Last three commits added back-end code to create event slots according to Provider Vacation dates, radio buttons to Clinic holidays for allowing half days.

Clinic Holiday preview:
holiday-form

Creating events using Provider Vacation:
vacation

Yet to address/fix:
1. Clinic holiday code. (Trying to pass dates and day types)
2. Closing of Add Vacation/Holiday window after submission.

@teryhill Please have a look.

modules/calendar/add.php
```diff
@@ -0,0 +1,354 @@
...
+    $args['form_prefcat'] = "0";  // None
+    $args['locationspec'] = $locationspec;
+
+    $def_facility = sqlFetchArray(sqlStatement( "SELECT u.facility_id
+                                                 FROM users u
+                                                 LEFT JOIN facility f on (u.facility_id = f.id)
+                                                 WHERE u.id = ?", array($_SESSION['authId']) ));
+    // default_user_facility denotes id of default facility of logged in user
+    $default_user_facility = $def_facility['facility_id'];
+    // Taking facility & billing facility as default facility of logged in user
```
apooravc: Since provider is on a vacation, I assumed this. Is this assumption fine?

teryhill: Yes that is fine . They should only be working with the facility that they have access to.

apooravc:
@teryhill Added code for Clinic Holidays case. It takes a date (key) & day (value) type (full day/half day) and then for each provider, creates a holiday event.

teryhill:
Tested and it works Great @apooravc . Can you go a step further and check the categories to see if they are a clinic option and put them in a drop down?

image

teryhill:
Yes but make it a drop down. If the user wants another category say "Blue Grass Day" they can and close the clinic for the event.

teryhill:
@apooravc hold on and let me think that thru,

teryhill:
@aethelwulffe @tmccormi I am thinking this thru and really instead of managing Holidays we are actually managing clinic closure. So instead on putting it on the schedule as holidays Should we use "Closed" instead and instead of "Add" on the calendar what about "provider/facility". With the intent to move all the provider options from where they are currently to here. Then ACL can be applied easier.

Thoughts?

teryhill:
What about Lunch? I think you are allowing the "Lunch" to produce an error.

apooravc:
@teryhill I tested sanity check with event category as "Lunch", it was working. Could you explain that user case and error? I noticed that I should also add provider id in case of checking adding vacation in `fetchAllEvents($start_date, $end_date)` because it is possible that provider B has some events b/w vacation range dates for provider A and this check will not let user add vacations in such case.

teryhill:
The check for provider needs to be done in both instances. I an getting the alert when I try to add holidays because the in office and out of office tags. @apooravc

apooravc:
@teryhill Isn't a holiday independent of provider, like if I make a holiday event on a date, it will be for all providers and it should check if no provider has any event scheduled which conflicts with holiday event on that day? So, what type of case would require inclusion of a provider check in Add Holiday?

teryhill:
@apooravc you are currently checking to see if any events are scheduled on the holiday date. The issue is Lunch , In office and out of office show up as events so you cannot schedule a holiday . The check needs to be is there any patient appointments on the day. PID != 0

apooravc:
Latest commit adds provider checks in both Add Vacation and Add Holidays options. With this:
1. In case of former, it restricts check to selected provider for which vacation is added.
2. In case of latter, it checks if on at least one of selected holiday date, at least one event is patient event then it will not add holidays for selected dates. If all events are provider events, it will add holidays.

@teryhill Please have a look,

apooravc:
Latest commit adds:
1. Ability to make holiday events by adding a specific facility.
2. Ability to delete a particular set of holidays by adding corresponding date and facility.

@teryhill Please have a look.

aethelwulffe:
I am going to apply this without and on top of some other PR commits to test.  Scheduled for 16:00 today.

modules/calendar/add.php
```diff
@@ -0,0 +1,548 @@
...
+// 2. Clinic Holiday dates
+if ($_POST['form_action'] == "holiday_submit") {
+  if ($_POST['holiday_action'] == "Delete Holiday") {
+    // user requests to delete a set of holidays
+    $dateAndType = json_decode($_POST['jsonData'], true);  // convert JSON string to an associative array
+    // this associative array contains elements in form of "date" => array("day type", "facility id")
+    foreach ($dateAndType as $date => $value) {
+      // for each set of date and facility, delete holiday event under all providers
+      // no need to iterate through providers in this case
+      sqlStatement("DELETE FROM libreehr_postcalendar_events WHERE pc_eventDate = ? AND pc_facility = ?", array($date, $value[1]));
```
teryhill: You need to add the pc_catid here or else you might delete appointments that are there.

modules/calendar/add.php
```diff
@@ -0,0 +1,548 @@
...
+    // check if there is an appt. (either recurring or non-recurring) on at least one of selected holiday dates
+    $dateAndType = json_decode($_POST['jsonData'], true);  // convert JSON string to an associative array
+    // this associative array contains elements in form of "date" => array("day type", "facility id")
+    foreach ($dateAndType as $date => $value) {
+      // check each selected date
+      $fetchedEvents = fetchAllEvents($date, $date); // returns an array of events on $date or including $date (in case of recurring appts.)
+      if (!empty($fetchedEvents)) {
+        foreach ($fetchedEvents as $event) {
+          // check if event is a patient event
+          if (event['pc_pid'] > 0) {
```
teryhill: Small bug is if you are adding for a specific facility and there is an appointment in another facility you get the message. Facility checking needs to be done.

modules/calendar/add.php
```diff
@@ -0,0 +1,548 @@
...
+        $args['facility'] = $value[1];
+        $args['billing_facility'] = $value[1];
+      // Insert events - A holiday event is made for each selected date,
+      // covering Calendar slots according to selected day type for that date with selected facility under all providers.
+      InsertEvent($args);
+    }
+
+    // reached here, means no conflicting dates, make requested events
+    foreach ($dateAndType as $date => $value) {
+      $providers = getProviderInfo();
```
teryhill: There is no facility check so the Holiday is applied to every facility, even if you select a facility it is still applied to all.

modules/calendar/add.php
```diff
@@ -0,0 +1,548 @@
...
+        $args['recurrspec'] = $recurrspec;
+        $args['starttime'] = $starttime;
+        $args['endtime'] = $endtime;
+        $args['form_allday'] = 0;
+        $args['form_apptstatus'] = "-";  // None
+        $args['form_prefcat'] = "0";  // None
+        $args['locationspec'] = $locationspec;
+        // Taking facility & billing facility as selected facility of holiday
+        $args['facility'] = $value[1];
+        $args['billing_facility'] = $value[1];
```
teryhill: Billing facility in non provider events is important and could be different from the facility. It does not matter with provider events just want you to know about it .

modules/calendar/add.php
```diff
@@ -0,0 +1,548 @@
...
+        <div class="col-sm-4">
+          <input type="radio" id="radio-choice-2" name="holiday-actions" value="Delete Holiday" class="holiday-action">
+          <label for="radio-choice-2">Delete Holiday(s)</label>
+        </div>
+      </div>
+      <div class="form-group">
+        <label class="control-label col-sm-2" for="clinic_id">Clinic</label>
+        <div class="col-sm-10">
+          <select class="form-control" id="clinic_id" name="clinic_id">
+            <option value="0" data-text="Undefined">Choose Clinic/Facility</option>
```
teryhill: An option for "All" facilities would be nice to have.

modules/calendar/add.php
apooravc: @teryhill If a user adds a holiday for a particular date, and if an appointment is on that date even if in another facility, shouldn't it be prevented from adding so as to not overlap it?

modules/calendar/add.php
apooravc: @teryhill Facility id is stored in `$value` corresponding to `$date`. I tested with different facilities and holiday was made w.r.t to selected facilities. It wasn't applied to all, am I doing something wrong here?

modules/calendar/add.php
apooravc: @teryhill So, if user makes a holiday with a date, day type and facility option "All", what should be stored in DB in facility id? I mean how should it work?

modules/calendar/index.php
```diff
@@ -138,6 +138,7 @@
       var title_agenda2 = '<?php echo xlt('2 day'); ?>';
       var title_agenda = '<?php echo xlt('1 day'); ?>';
       var title_search = '<?php echo xlt('search'); ?>';
+      var title_add = '<?php echo xlt('add'); ?>';
```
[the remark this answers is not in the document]
apooravc: Calling it "Provider" would be good as it is also the tab name in event panel.

modules/calendar/add.php
teryhill: What is w.r.t?

apooravc: with respect to

teryhill: I have a provider that is only at one facility and other providers at another facility. If I select the facility that has one provider and make a holiday it makes it for "All" providers no matter which facility is selected.

modules/calendar/add.php
teryhill: If one facility needs to close it should not be prevented by another facilities appointments. They are separate. You can wait until customers bring up issues if you want.

modules/calendar/add.php
teryhill: The facilities listed in the facility table. You could wait on this also and see if the customers want it.

modules/calendar/add.php
apooravc: Oh, so you are talking about it being applied to all providers when we make a holiday event. That's the default behavior because facility is closed no matter which provider.

teryhill: What if they use this to close a facility because the facility is being repaired?

apooravc: To add under only those providers which are at that facility selected for holiday, I think we have to take help of getUsersFacility.

apooravc: @teryhill Will change adding under all providers to under only those providers specific to that selected facility.

modules/calendar/add.php
teryhill: Set up 2 facilities, add the facility to a user as the default. Add an appointment to that user's calendar. On the second facility add a holiday for the same day the appointment is on.

apooravc: @teryhill It did the alert. So, overlapping of events in case when date is same, provider is same but facility is different should be fine? And it should only alert when holiday's and appt.'s date, provider and facility are same?

apooravc: I think checking with date and facility is enough because if facility is closed, there can't be that appt. in any case, right?

teryhill: if ($event['pc_pid'] != 0) {
              if($event['pc_facility'] == $value[1]) {

apooravc:
@teryhill Did some changes:
1. Added a check b/w holiday's facility and event's facility on same date.
2. Holiday would now add under providers which are specific to that facility of holiday.
3. Added translation identifiers.
4. Added holiday category in holiday deletion.

Please have a look.

modules/calendar/index.php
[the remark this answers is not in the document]
aethelwulffe: Either one is fine,  and a reference to NPI is not international.  Provider/Facility is most completely descriptive, but long.

aethelwulffe:
REstesting now.

aethelwulffe:
Nevermind.  I got to kill my error reporting to test this.  Putting it a couple of spots back in the queue.

aethelwulffe:
@teryhill Please go over this to check if all your points are met with the appointment/calendar schedule of holidays.  Lots of work just sitting here.

aethelwulffe:
These are all approved.  Three months without second approval or protest.  Merging.  Everyone, this is now up for general testing in the main branch.  Let me know if there are issues.  I did my test on this LOOONG ago.

PR #1186, opened 2018-06-19
moved time zone

teryhill:
Apooravc had the function for time zone in the globals.inc.php I moved
it to edit_global.php @apooravc  you can combine this with your pr. #1179


teryhill:
@apooravc the only thing I did not do was read the one from php.ini and show it as the default.

teryhill:
@apooravc is going to incorporate this into his pr closing this one

PR #1187, opened 2018-06-20
sql_upgrade move/hide/menu

aethelwulffe:
Hiding sql_upgrade, requiring auth, adding to admin menu, preventing direct access.

This may not be a real security issue, but it certainly makes things easier for the user to upgrade.

### DIFF
--- a/sites/default/menu_data.json
+++ b/sites/default/menu_data.json
@@ -491,6 +491,14 @@
             "requirement" : 0,
             "acl_req" : [ "admin", "users" ]
         }, { 
+            "label" : "Upgrade Database",
+            "menu_id" : "adm0",
+            "target" : "adm",
+            "url" : "/sql/sql_upgrade.php",
+            "children" : [ ],
+            "requirement" : 0,
+            "acl_req" : [ "admin", "super" ]
+        }, { 
             "label" : "Native Data Loads",
             "menu_id" : "adm0",
             "target" : "adm",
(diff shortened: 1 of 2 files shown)


tmccormi:
Concept is good.  Is there some way to warn the user that the sql is out of date? We used to do that the database version.php compared to the table value.  An alert at login would be nice

aethelwulffe:
concept is flawed.
basically, we need the old admin.php with a login.  I just upgraded a bunch of code, then tried to use this.  The upgrade caused fatal errors without the database installed for each site in the installation, so I had to add back in the old Wild Wild West version to get the job done (without resorting to turning the upgrade queries into a patch).  I am going to push the old version back to the code in this branch.  There is at least a possibility it might get used this way.  Pretty sure that the new upgrade tool should be the answer here.  Still got to check out the multi-site version of that.

aethelwulffe:
There is still the sql_patch stuff and all that rot.  some of it looks at version one way, some another way.  I don't think either one properly adjusts database versions.
-Certainly need that version warning...maybe on the login page, yes.  The login page *almost always* displays after an update.
  This could all be just legacy poo after the installer comes around, but the DB part of that is not yet done.

tmccormi:
I don't think that's a bad idea, just make it so admin.php requires admin login to function.

aethelwulffe:
Tinkering with it now.  Can't step on current install stuff, can't give site admin to every site within every site, can't make a new login that might get forgotten and lock someone out of it, onandon.
I think there are a couple of clean possibilities though, and those include the "out of date" warning, but that is going to be the new setup responsibility...so, I will do this as a contribs thing just so it is out there for possible use.

PR #1173, opened 2018-06-09
Commented the unused code

RAJAGOPALAN-GANGADHARAN:
The part of code which focuses on adding new patients is now useless so i have commented it.

PR #1188, opened 2018-06-22
Update menu patient add/search link

apooravc:
Trying to open "Search or Add Patient" tab by going main menu > Patient/Client > Add New/Search results in following blank tab to open:
empty

It happens because link to this tab isn't updated in menu_data.json while changing tab's file from `interface/new/new.php` to `interface/new/new_comprehensive.php` in #1173. @aethelwulffe Please have a look.

### DIFF
--- a/sites/default/menu_data.json
+++ b/sites/default/menu_data.json
@@ -1,11 +1,11 @@
-[ { 
+[ {
     "label" : "Calendar",
     "menu_id" : "cal0",
     "target" : "cal",
     "url" : "/interface/main/main_info.php",
     "children" : [ ],
     "requirement" : 0
-}, { 
+}, {
     "label" : "Flow Board",
     "menu_id" : "pfb0",
     "target" : "flb",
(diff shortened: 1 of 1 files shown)


muarachmann:
was the merge from @RAJAGOPALAN-GANGADHARAN  done if so then you are right @apooravc there was unused code and @RAJAGOPALAN-GANGADHARAN did a pretty good clean up. Will check that

sites/default/menu_data.json
```diff
@@ -1,73 +1,73 @@
...
         "url" : "/interface/main/finder/dynamic_finder.php",
         "children" : [ ],
         "requirement" : 0
-    }, { 
+    }, {
         "label" : "Add New/Search",
         "menu_id" : "new0",
         "target" : "pat",
-        "url" : "/interface/new/new.php",
+        "url" : "/interface/new/new_comprehensive.php",
```
muarachmann: i guess this was the key thing to do here approving these changes looks good to me

apooravc:
Yes @muarachmann it was merged.

sites/default/menu_data.json
aethelwulffe: Basically, you have to migrate the menu changes to any /sites directories you are using.  This is always the case for menu upgrades, which is why Terry wanted to the menu in the database in the first place.  Upgrading each site file is a pain.  That means that since we don't even allow the normal use of the /default site anymore (read only essentially), you always have to manually patch the sites directories in use.
  Menu admin could use a button that says "update from default" or something like that to grab the new menus too.  Or even a save/load function.

aethelwulffe: I could have sworn that he had the change in the default site.  It was in all the roles stuff though.  I admit I did it manually because I was not in default.  Approving mad merging.

aethelwulffe:
There were a lot of whitespace changes in this.  I suppose we have to get used to seeing those.

muarachmann:
yeah was just for clean code i love it too.

aethelwulffe:
The menu updater tool changes that stuff every time you use it.  That is the format we should try to match.  I try to use the updater tool to achieve that result instead of manually updating.

Issue #1177, opened 2018-06-11
Large Databases are an issue. Demo database with 10000 patients Created using Arts Data generation program


teryhill:
We have been working with a large demo database. If you run a billing screen with no date of service and status of unbilled, with the background refresh of the calendar set to 60 seconds. The system will hang up, sometimes losing the MySQL connection. We have max execution set to 600 and max input vars set to 3000.

aethelwulffe:
This is an old issue.  Should require a date range, or specify "do you really really want to do this".

aethelwulffe:
Should definitely fix this!

AfiMorgan:
I turned on slow-query-logging and set the time limit to 2seconds.  
Went in to the cmd line and started watching with`tail -f query.log`
Then went into billing reports, unchecked the "data of service" and update the list.
I saw this query on line 1178 in interface/billing/billing_reports.php run over 60k times:
` $resMoneyGot = sqlStatement("SELECT pay_amount as PatientPay,date(post_time) as date FROM ar_activity where ".
        "pid = ? and encounter = ? and payer_type=0 and account_code='PCP'",
        array($iter['enc_pid'],$iter['enc_encounter']));`
Now we need to figure out the best way to fix it...

apooravc:
A bit similarity with #774 regarding large database.

aethelwulffe:
Long demographics page loading is an issue as well.  Possibly related to billing popups, meds/issues lists or any number of things.

aethelwulffe:
@apooravc @muarachmann @teryhill @Trodrige @naveen17797  and everyone else:
Link to a tested and working patient data generator version:
http://suncoastconnection.com/hosted/Downloads/PrintaPatient_A7_2_2.zip

PR #1189, opened 2018-06-22
Cover Calendar for facility ACL

apooravc:
@aethelwulffe @teryhill Opened this PR to cover Calendar related changes for facility ACL (#1171). 

Attempted a fix, please have a look. This is how it goes:
A user logs in with `facility_acl` global checked --> user is restricted/assigned to a particular group of schedule facilities --> filter patients with with facility in that group --> make an array of id (pid) of these patients --> while getting events, before displaying an event on Calendar, check if id (pc_pid) of that event is in array --> if it is not, then `continue;` without getting that event (and displaying that event). So, events of a patient with facility not in schedule facilities of logged in user are not shown to that user.

Is this covering all possible cases?


aethelwulffe:
I wish we had a way to simply restrict the calendar view on the basis of facility, vs. building a complete array out of all pid's and iterating through them.  

I am pretty certain that when you schedule an event, a facility is required.  If you can display the appointments just based on the facility (not billing facility) of the event, then check that against the schedule_facilities, I believe you need not actually check the patient list.
Correct, yes?

apooravc:
Yes, that is also an option and a better one. I went with this focus-to-patient option for using the newly added facility in patient_data and also that the action hook returned `pd.facility ...`, so as to make SQL queries with patient_data table seemed fit.

teryhill:
@aethelwulffe does bpuc use just one billing facility? I am thinking if they do we could remove it from the appointment screen.

apooravc:
Changed from **focus-to-patient** to **focus-to-facility** so as to check event's facility directly with schedule facilities of a logged in user.

teryhill:
@apooravc  I will test it as soon as I can but it will be a while. Maybe @aethelwulffe  can get to it before me.

This should keep the patients off the calendar if the user does not have access to that facility.

aethelwulffe:
The user should still also be restricted to scheduling for the restricted facilities, and I am not even sure if that feature works. 
  I have some testing to do today.  Thanks Apoor!
-I assume I need this on top of the other PR, but how about you pull that stuff into THIS PR, and I close the other?

aethelwulffe:
@teryhill THEY don't use more than one group NPI (now...I think...) but l have others that certainly do.  ARSP alone has four group NPI's, each for totally different types of programs, with different providers, and different supervisors.
-The schedule_facilities should be looked at when building either the "facility" or billing facility" drop-downs.
@apooravc  Only slightly off topic again: Just so there is no terminology mix-ups:

- A facility can be a "Place of Service", a "Billing Location", or a "Primary Business Entity" or any combination of the above.
- Simple clinics will likely have one facility with that one location being everything.
- A clinic that has doctors that go to see patients in a hospital as well as in their clinic will have their clinic as everything, with hospitals as "Place of Service" facilities in their system.
- A "Place of Service" boils down to a POS code, which is a two digit number like 11=office 6=home,21 and 22 being inpatient and outpatient hospital settings and the like.  Normally you supply that location's Tax ID and other things with it if they exist (for hospitals or other businesses).
- A Billing Location is where the money is sent to.  The "Group NPI" or Group National Provider Identifier is supplied to outgoing claims from this, as is the address and Tax ID involved in the transaction.
- A Primary Business Entity is purely the legal business entities information.  It *might* not have anything to do with either the above, so it must be specified, but it usually is also one or more of the above.

There are lots more to all this, and I should prepare a developer primer for them (and add it to the user docs), both as an explanation, as well as a point of comparison so that folks can use, inactivate, or modify/expand the base paradigm for use in other countries.  I know India is 70% self-pay, and there is, frankly, a lot of data systems knowledge in India for processing outsourced claims, but it should *cleanly* adapt to wherever it is use.

apooravc:
@aethelwulffe There are two globals related to this schedule facilities and Calendar:

1) Restrict users (providers) to Facilities: If checked, the facilities selected in Schedule Facilities of Edit User Admin (of a user) are the only ones to appear as an option in choosing Facilities while making an event in event panel with that user (as provider). Implemented in #1137.

2) Restrict user access by Facilities: If checked, a user logged in to EHR could access information of only those facilities which comes under that user's Schedule Facilities (Edit User Admin).

Yes, this PR is an extension of the (and based on) original Fac. ACL Attempt PR. Should we keep these two separate for further developments, if required?

aethelwulffe:
No, you should take over the PR for the feature in my opinion, executing it entirely at your discretion.  This is where it should happen, though there is no reason not to keep the other thread open for documentation continuity *if you see a need*.  I should have discussed much of that in an issue and not the PR.  It is a lot cleaner to develop a new feature in a single branch and not...the other way!
  Say the word and I will close the other PR.

apooravc:
@aethelwulffe This third commit shifts code from PR **1171** to this branch.

aethelwulffe:
WooHoo.
Pulling tests now.

aethelwulffe:
@sam-bowen Try this!

aethelwulffe:
OK,
A great many things seem to do exactly what I expect.
Other than the fact that logging out as one user and logging in as another screws up the tab label display and some other things if you don't clear cache (not related) I only see one minor boo-boo in the calendar itself:
With the global set that is supposed to restrict users to *scheduling* appointments at only their allowed facility, the user still gets a full list of facilities to set an appointment to.

Associated issue is that we still have the problem where with a large number of facilities, the modal window disappears off the bottom of the list.  You have to zoom out to see it.  I imagine that we have an izimodal fix for this right now (that I will look for, retest, and insist on attention).

aethelwulffe:
OK...  More puzzle, but defined results from above.:
Facility 1 is set as a billing facility, and that shows up when adding an appointment, in addition to the facility that the PROVIDER is restricted to, not the user.  That can...make sense I suppose, but that allows the user to schedule for other facilities, as anyone can see any provider.  This makes my previous observations inaccurate.

In ANY CASE however, Facility 3 (with ID 3) shows up no matter what.  It seems it is not possible to restrict access to facility ID 3.
DELETING facility 3 made the issue go away completely without seeming to break anything...but that is not a real option for production.
Adding facility 3 back causes the problem to re-appear.

I verified that the providers in question did not have Facility 3 in their list.

So, aside from the facility 3 thingie, and the facility admin modal window being goofy, everything is great.

aethelwulffe:
1506            if (in_array($facrow['id'], $ufid) || $facrow['id'] == $e2f || $facrow['id'] == 3) {
 1507              //id 3 is for "Your Clinic Name Here"
-in add_edit_event.
-No idea why ID 3 is included as hard-code.  Normally, even though for some weird reason we start with ID 3 in the default database (WHY????)
like so:
`INSERT INTO `facility` VALUES (3, 'Your Clinic Name Here', 'Your Clinic Name Here', '000-000-0000', '000-000-0000', '', '', '', '', '', '', NULL, NULL, 1, 1, 0, NULL, '', '', '', '', '','#99FFFF','0', '');`

We should change that initial facility ID to 1 for starters.  Next, normally you EDIT this facility, not leave it "Your Clinic Name Here", and that in no way guarantees it is any particular special facility.
-So, I believe just getting rid of that || $factro['id] == 3 should work.  Ja?

sam-bowen:
I never actually figured out why Bowen Primary & Urgent Care, PA is "facility 3".  We did have previous names Unifour Internal Medicine, then BPUC, then Apollo Healthcare, then back to BPUC.  I suspect it has something to do with human error in terms of setting up the database, such as 2 prior attempts at setting things up in 2003 that did not work out so well.

Samuel T. Bowen, MD

sam-bowen:
I'm just guessing, but what I remember is that either the MySQL auto increment function or the oemr auto increment function was very difficult to cause to restart counting from 1.  Two failed attempts resulted in two lines being deleted, then the auto increment insisted on starting at 3.

This is the only explanation I can offer.  The hard coding of Facility 3, seems like Williams Medical Technology doing something easy, instead doing it correctly.

Sam Bowen, MD

sam-bowen:
Of course the other possibility, Facility 3 was likely originally hard coded because it predated any idea of more than one facility being necessary.  It also pre-dates the multi-language code.  The initial conversion to the multi-language could have caused this.

Samuel T. Bowen, MD

sam-bowen:
So, what I am saying, I guess is, changing Bowen Primary & Urgent Care, PA = Facility 1, sounds like a reasonable idea, but it makes me nervous, based on a long history of painful negative economic stresses.

Samuel T. Bowen, MD

teryhill:
@aethelwulffe the facility 3 stuff needs to be checked thru the code before it i removed. There are (ia would guess) a number of spots where it is hard coded. The code needs a facility to run out of the box. The issue you are having with facility can be fixed by removing the check in line 1506 of the add_edit_event program.

You can test this by removing this portion of the or statement  || $facrow['id'] == 3),
Try that and see if that makes the facility 3 thingie go away.

teryhill:
@sam-bowen I vote for making the small change above and leaving your long standing data alone (for now)

aethelwulffe:
@sam-bowen Yeah, this has nothing specifically to do with your data.  
This hard-coded facility 3 thing was done by one of our team fairly recently.  It is a new thing.
Testing an install run with initial facility ID=1 on the default facility.  
Personally, I have never seen code that specified a facility ID other than in database.sql and in this facility check in the calendar that was just recently merged.

-but I am looking.

aethelwulffe:
Kid Tested, Mother don't care...
image
Seems totally fine to me.  This is against the PR I made changing the default facility ID.

aethelwulffe:
NOt finding any (other) hard-coded terms against the facility table so far...searched for quite a few.

teryhill:
I have not found any instances of it being hard coded. Looks like the change can be done.

aethelwulffe:
I am also not finding any use for "defaults.sql"
-Adding that redaction to my db change PR once my really old code check finishes up here to see what it really was used for...
-and it looks like it was part of a very very old demo data set...or something.  No real results in OE4.0 either.

aethelwulffe:
@apooravc if you can retract the OR facility.id = 3 condition, I think this stuff can roll.

apooravc:
@aethelwulffe This `|| $facrow['id'] == 3)` was added to include text _Your Clinic Name Here_ in facility drop down in event panel but as you said it is edited normally, I'll be removing it with next commit.

teryhill:
@apooravc  this should be changed in add_edit_event.php 
"With the global set that is supposed to restrict users to scheduling appointments at only their allowed facility, the user still gets a full list of facilities to set an appointment to."

Make sure that the facility dropdown respects the global flag.

teryhill:
@aethelwulffe how close is this PR to meeting the requirements.

aethelwulffe:
It meets the (related) minimum requirements, and does not seem to stop on anything else.  It is a good feature branch.
-Apoor already knows the associated admin UI stuff that will make use easier (user admin UI for schedule_facilities), and there are other PR's that address associated issues like editing facilities.

I think this one is in the box as an initial implementation.

apooravc:
@teryhill @aethelwulffe Regarding this:

"With the global set that is supposed to restrict users to scheduling appointments at only their allowed facility, the user still gets a full list of facilities to set an appointment to."

Are we talking about this **Restrict Users to Facilities** global/checkbox? 

So, when a user is making an appointment with a provider, should the facility drop down list facilities restricted to that provider slot upon which user clicked or that logged in user? I think this _scheduling appointments at only their allowed facility_ means according to the logged in user.

In #1137, I changed a line governing this - `$facils = getUserFacilities($_SESSION['authId']);` (which lists according to logged in user) to $facils = getUserFacilities($userid); (which lists according to provider slot upon which that user clicked).

aethelwulffe:
Actually, both approaches sort of make sense, but the supposed (and probably best) behavior is that when anyone tries to schedule an appointment, they only get a drop-down that matches their own user settings.  -Currently, this PR has it as the *provider's* settings.  "Access/use according to the the logged-in user" is the best concept in my opinion.

We want to give the restricted user the experience that they are totally isolated to their little part of the kingdom...as much as possible.

The feature requesters are used to jumping through hoops and juggling things like this to get them to work right, so we would not hear too many complaints with this as-is, but changing the logic to fit the above would be a good thing.

apooravc:
@aethelwulffe Added commit for **facility lists according to user**. Now, if we restrict some user to facilities A, B, and C in Schedule Facilities in Edit User Admin...that user when logged in, while making an appt. with any provider, should only have facilities A, B, and C in facilities drop down (along with default facility of that user which is also allowed/scheduled).

aethelwulffe:
This feature actually helps us with a future certification as well (on at least five points).

apooravc:
@aethelwulffe A change is added in this commit which changes default facility in facility lists from according to provider (event slot) to according to logged in user, fixing following:

If there are four facilities 1, 2, 3 and 4 and two users A and B. Suppose, A has default facility as 1 and schedule as 1, 4 and B has default facility 2 and schedule as 2, 3. Then, while creating an event while logged in as A, facility drop down shows default as 1 only upon clicking on provider A's slots and as 2 on provider B's slots (meaning default acc. to provider), but it should only show default as 1 on all providers' slots, while logged in as A.

Just changing `$userid` to `$_SESSION['authId']` in default facility `$e2f` block.

sam-bowen:
We installed the new code changes.  We are just now starting to test on our database.  Current impression is that response time of the demographic is still taking 25 seconds to complete the query and display the results.  Other queries such patient finder (very rapid), pulling up individual encounters, starting speech_dictation forms are much more like normal response times.  We will continue testing.

Does this issue need its own thread to bring a larger audience of developers?

Sam Bowen, MD

aethelwulffe:
These changes look good.
@sam-bowen I suggest telling the kids to turn OFF the tags feature global and directly observe the performance differences.  If this was already turned off, we need to log the length/speed of the queries.  They should also have the CDR engine AND translation engine turned off for the base test.

aethelwulffe:
@apooravc @teryhill Dr. Bowen is attempting to convert his clinic from old heavily customized openemr code to LibreEHR.  There have been...challenges.  His database is a good one for finding performance issues.

sam-bowen:
Is this something that we could run a query and turn these off for ALL users and set them to OFF as default?

Sam Bowen, MD

aethelwulffe:
Those are just globals.  They don't have to be individually turned off.  Just disable the whole TAGS system.
-While thinking about it, I better check and make sure I had re-enabled a global toggle in this code...

sam-bowen:
There are 6 globals to "Translate...." do these all need to be turned off?
Sam Bowen, MD

aethelwulffe:
Yes, I did.  Look for a feature global called "Enable the TAGS/FILTERS feature" and kill it.  The tag configurations can just sit there (for the test install at least).
The translations can all be turned off, but the "other languages allowed" kills everything.  Just turn them all off though....

sam-bowen:
"Enable Role-based menu system" should be turned on?

sam-bowen:
Leave "other languages allowed" on?

sam-bowen:
Default Language English (Standard)

All Languages Allowed
Enable the TAGS/FILTERS feature
Allow Debugging Language	
Translate Layouts	
Translate Lists	
Translate Access Control Groups	
Translate Patient Note Titles	
Translate Document Categories	
Translate Appointment Categories

Are all off

Activate CCR/CCD Reporting

is on

Demographic page response to load is 23 seconds.

Sam Bowen, MD

aethelwulffe:
Can Marcus set up long query logging for like...5 second intervals and then load some dems and send me the logs?
-On topic, @apooravc @teryhill it appears that with this morning's changes, we are at a publishable point.  Performance issues are related to the process of loading and displaying the demographics layouts.  We will investigate those further to find the time-hog(s).

aethelwulffe:
We have added tons of "cool features" without worrying about overhead.

aethelwulffe:
From my side, it looks like they are not actually slow queries...starting an issue for this.

aethelwulffe:
Ref #1177 already an issue in progress.

modules/calendar/add_edit_event.php
```diff
@@ -939,8 +939,8 @@ function setEventDate($start_date, $recurrence)
 
     //-------------------------------------
     //(CHEMED)
-    //Set default facility for a new event based on the given 'userid'
-    if ($userid) {
+    //Set default facility for a new event based on logged in user, not provider (event slot)
```
aethelwulffe: Ooops.  Looks like maybe the array element session variable should be authUserId and not authId.

apooravc: Did a console.log test using `<?php echo "<script>console.log('{$_SESSION['authId']}')</script>"; ?>` &
    `<?php echo "<script>console.log('{$_SESSION['authUserId']}')</script>"; ?>`, the latter is blank while former corresponds to an user id.

aethelwulffe: AGreed...I am casting around looking for problems, because we can't get any appointments to display on my branch test or Marcus Bowens'.
Periodically receiving the "error occurred while fetching appointments."  Looking into everything I can right now.

aethelwulffe:
Also, with $userid we can get a drop-down of facilities, but can't with authID.
Looking looking....

aethelwulffe:
I lied.
Apparently updating globals so many times (like twice in a row) crashed mysql.
restarted and cleared cache, and can now see the calendar.   ARgh!

modules/calendar/add_edit_event.php
teryhill: @aethelwulffe that error is usually caused by the get_provider_events.php not the add_edit_event.php

aethelwulffe:
Getting confused due to too many odd behaviors. 
Global settings related to facility access and restriction are off, everything clear, can see appointments and schedule them.

Going to turn on just facility ACL, restart everything, then test again.

aethelwulffe:
Results:  just tried to update globals....hit 151 connections to the database, and everything froze up.
-going to kill a bunch of processes and try again.

aethelwulffe:
Killed mysql, logged back in, and can see appointments.

apooravc:
@aethelwulffe Could you also delete all tuples from users_facility table before testing, you might have to set schedule facilities of users again but recent changes from changing facility lists from provider to logged in user might cause wrong results otherwise in drop down because of old relationships in those rows (according to provider).

aethelwulffe:
I hear you.
I am going to build a brand new installation and database and play there.
With the application just sitting there, the maximum connections hit the limit and caused queries to fail.   
  Just updated globals again to set the facility acl flag on.  This time it finished updating leaving like four available connections open.  Restrict users to facilities is still off.  Going to kill the threads and try to keep testing this.

sam-bowen:
I had a memory that floated back into my consciousness from years ago.  I remember having this performance issue related to software intended to collect data for meaningful use.  I turned it off in order to get the performance up to where we could use the software.  

Is there an engine collecting data for meaningful use and can I turn it off?

Sam Bowen, MD

aethelwulffe:
That stuff is all mostly dead... replaced by the .but shutting down most 
of the logging certainly helps.

-this issue is likely the data quality.  I am running tests etc... to 
move this forward.  Once we have a clean result with clean data again, 
we can have this feature moved into the code base.  There is a 
performance issue...issue for documenting the performance issues.  There 
are far fewer issues for some large installations than others.  The 
difference is obviously in either the quality or the content of the data 
(or the server variables).

modules/calendar/includes/get_provider_events.php
```diff
@@ -24,7 +24,35 @@
...
+  $facilities_allowed_to_user = array();  // array containing facility ids which logged in user is allowed to see
+
+  // making of $facilities_allowed_to_user
+  if ($facility_filter) {
+    // $facility_filter = " pd.facility IN ( user_schedule_facility_id_string ) " from tags_filters/start.php
+    $facility_filter = substr($facility_filter, strpos($facility_filter, 'IN'));
+    $facility_filter = ' facility.id ' . $facility_filter;
+    $sql = "SELECT facility.id AS fid
+            FROM facility
+            WHERE" . $facility_filter; // $facility_filter = " facility.id IN ( user_schedule_facility_id_string ) "
```
aethelwulffe: Boo-Boo here.  should say 
            `WHERE" . $facility_filter;"; // $facility_filter = " facility.id IN ( user_schedule_facility_id_string ) "`

aethelwulffe: Only took me 14 hours to find a simple syntax error... 😜

aethelwulffe: Correction:  
    $sql = "SELECT facility.id AS fid FROM facility WHERE" . $facility_filter.";"; // $facility_filter = " facility.id IN ( user_schedule_facility_id_string ) "

teryhill: That looks like it would stand a chance of working the other one did not.

So did that fix the error?

aethelwulffe: NO.
The error log output debug string echoing the $sql reads:
get_provider_events.php query string: SELECT facility.id AS fid FROM facility WHERE  facility.id  pd.facility = '-1' ;

The concatenated value must be not right...

aethelwulffe: Oh geezz...maybe that was NOT supposed to be a comment!

teryhill: No that is ok, Do you have a facility assigned to the user?

aethelwulffe: Yes.
The error_log($sql) output (and subsequent syntax error generated) shows that it amounts to a bad query.
WHERE facility.id pd.facility = '-1'  isn't good.  There is no join there.  The function that it pulls that from with all the substrings and such is a bit beyond me this late at night.  I guess I should track them down to try to help Apoor when he wakes up to this mess of typing I have been generating.

teryhill: From what I have figured out the actual thing that is being returned by the code is

 pd.facility = '-1'

Which indicated there is no facility information getting rendered from the do action thingie.

function tf_filter_patient_select_pnuserapi2( $username )
{
    $facilitiesToShow = get_facilities_to_show( $username );
    $where = " pd.facility = '-1' ";
    if ( count( $facilitiesToShow ) ) {
        $facilityString = implode(",", $facilitiesToShow);
        $where = " pd.facility IN ( $facilityString ) ";
    }
    return $where;
}
add_action( 'filter_patient_select_pnuserapi', 'tf_filter_patient_select_pnuserapi2' );

aethelwulffe: I guess the function call to start.php isn't returning an expected value.  I swear I think I have the database set up to use the feature.  This thing has been a pain for years now.  It was a pain in the old version due to non-optimal configuration and hinkey stuff, and it is a pain now.  Apoor will probably figure this out in two minutes.

apooravc: Putting `error_log($sql);` after this sql statement resulted in following logs:
1. When logged in as a user with no mention in `users_facility` table, i.e., no allowed facilities:
`SQL Error with statement:query failed: SELECT facility.id AS fid
            FROM facility
            WHERE facility.id  pd.facility = '-1' --You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'pd.facility = '-1'' at line 3==>C:\wamp64\www\pro\lh-ehr\modules\calendar\includes\get_provider_events.php at 41:sqlStatement`

2. When logged in as a user with a few mention in users_facility table, i.e., some (here 2) allowed facilities:
`SELECT facility.id AS fid
            FROM facility
            WHERE facility.id IN ( 5,4,5 )`

I think the code in start.php is working properly and returning the required string of allowed facilities and the issue may be resulting from the case when it returns `$where = " pd.facility = '-1' ";` because of having no allowed facilities corresponding to logged in user in users_facility table. 

And I haven't included any handling of this case in my code resulting in sql error. I'll add it in my next commit. @aethelwulffe When logged in as a particular user, please check if that user has at least 1 corresponding facility in users_facility table and see if correct sql (with facility id string) is logged. 

In case of no allowed facilities for logged in user (when there is no mention in users_facility table), I think there is no point of performing any sql query because `$facilities_allowed_to_user` would be empty even if we perform query by changing `pd.facility = '-1'` to `facility.id = '-1'`. And all events in Calendar would not be fetched/displayed.

apooravc:
Putting `error_log($sql);` after this sql statement resulted in following logs:
1. When logged in as a user with no mention in `users_facility` table, i.e., no allowed facilities:
`SQL Error with statement:query failed: SELECT facility.id AS fid
            FROM facility
            WHERE facility.id  pd.facility = '-1' --You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'pd.facility = '-1'' at line 3==>C:\wamp64\www\pro\lh-ehr\modules\calendar\includes\get_provider_events.php at 41:sqlStatement`

2. When logged in as a user with a few mention in users_facility table, i.e., some (here 2) allowed facilities:
`SELECT facility.id AS fid
            FROM facility
            WHERE facility.id IN ( 5,4,5 )`

I think the code in start.php is working properly and returning the required string of allowed facilities and the issue may be resulting from the case when it returns `$where = " pd.facility = '-1' ";` because of having no allowed facilities corresponding to logged in user in users_facility table. 

And I haven't included any handling of this case in my code resulting in sql error. I'll add it in my next commit. @aethelwulffe When logged in as a particular user, please check if that user has at least 1 corresponding facility in users_facility table and see if correct sql (with facility id string) is logged. 

In case of no allowed facilities for logged in user (when there is no mention in users_facility table), I think there is no point of performing any sql query because `$facilities_allowed_to_user` would be empty even if we perform query by changing `pd.facility = '-1'` to `facility.id = '-1'`. And all events in Calendar would not be fetched/displayed.

interface/tags_filters/start.php
```diff
@@ -85,5 +85,97 @@ function update_tags_filters()
...
+function get_facilities_to_show( $username )
+{
+    // User facility is stored in users_facility table as facility_id
+    $sql = "SELECT UF.facility_id, U.facility_id AS default_facility
+        FROM users U
+        JOIN users_facility UF
+        ON UF.table_id = U.id
+        WHERE
+            U.username = ? AND
+            UF.tablename = ?";
```
apooravc: @aethelwulffe @teryhill I think a `;` is not required for a single query statement inside string. I think I haven't seen it inside string anywhere in code base.

aethelwulffe:
Hi Apoorv!
I am rested and now sane I think.
Pulling down the "no facility" case changes and having another go!
-Thank you so much!  I have been telling people for years that this solution is "just around the corner".  Now it really is.

aethelwulffe:
OK,
What I am experiencing.
Setup: 
 100 patients
 two facilities (FAC3 with 99 patients and Fac4 with 1 patient)
three providers (Using "TheArtist FormerlyKnownAsArt" = provider, calendar, admin).

Test case results:

1.  Demonstrate fake patient Marion P. Campbell, assigned to FAC3.
2.  Demonstrate fake patient Anthony H. Aguilar, assigned to Fac4.
3.  Demonstrate Globals settings showing enabling both facility restrictions
4.  Lots of processes launched with global updates when calendar open.  No failure warning for appointments.
5.  Setting TheArtist FormerlyKnownAsArt exclusively to Fac4.
6.  Using the dynamic patient finder to list accessible patients (only one result).
7.  With Aguilar selected as active patient, clicking in The Artist's schedule, automatically selects the patient.  Created appointment.
8.  Cleared active patient.  Create another encounter and searching for a patient (1 result only, correct).
9.  No calendar results displayed for either utilized appointment creation method.
10.  Deactivated the global toggles for FACACL and FAC selection restrictions.  Restrict to facility done first (no change in behavior), then both off.
11.  With both features off, patient appointments show.

Code is today's latest commit.
image

aethelwulffe:
Remarks:
Appointment retrieval errors only cropped up during globals updates, not constantly.  That is good.
When creating an appointment, the available billing facilities show up no matter what.  That is good.

apooravc:
@aethelwulffe When you're logged in as TheArtist FormerlyKnownAsArt, are there any rows in users_facility table corresponding to your user id?

modules/calendar/includes/get_provider_events.php
```diff
@@ -24,7 +24,44 @@
...
 $fetchedEvents = fetchAllEvents($_POST['start'], $_POST['end']);
 
+// facility ACL check
+if ($GLOBALS['facility_acl'] == 1) {
+  $facility_filter = do_action('filter_patient_select_pnuserapi', $_SESSION['authUser']);
+  // do_action returns either
+  // 1. $where = " pd.facility IN ( $facilityString ) "; i.e. schedule/allowed facility ids in a SQL string or
+  // 2. $where = " pd.facility = '-1' "; when logged in user has no corresponding facilities in users_facility table i.e. no allowed facilities
+
+  $facilities_allowed_to_user = array();  // array to contain facility ids which logged in user is allowed to see
```
apooravc: @aethelwulffe I think the reason for not displaying any appointment is that in all cases, they are getting `continue;`d. Please move this array definition to somewhere before line 27 so that it's available in both if blocks and check again.

aethelwulffe: Moved line.  Debug out SELECT facility.id AS fid\r\n            FROM facility\r\n            WHERE facility.id IN ( 4,4 )
actual settings:
image

No display of appointments.

aethelwulffe: Multi-select for schedule facilities not working with the feature off.  Does not properly echo changes when on.  Adds additional facilities when there are a total of 3 (adding 2)

teryhill: In the user code, right. May be there needs to be another chapter to this novel.

apooravc:
@aethelwulffe Found something. If you check my latest commit, you'd see that I've added some custom error logs to check where flow of control is going. If you do not include the change I made in appointments.inc.php, you may find your error log after refreshing calendar like this:

Note: For my tests I used 2 patients, patient 1 with facility as fac 1 (id = 6) and patient 2 with facility as fac 2 (id = 4). USER 1 has schedule facility as fac 1 and USER 2 has schedule facility as fac 2 in users_facility table. With feature turned off, I made two events - EVENT 1 at fac 1 with patient 1 and EVENT 2 at fac 2 with patient 2.

blank

`error_log($event['pc_facility']);` was logging blank line and going backwards (`$fetchedEvents` --> `fetchAllEvents()` --> `fetchEvents()` --> `$query`), I found that `pc_facility` is not `SELECT`ed and is therefore not in returned appointments array and so not in `$fetchedEvents` or `$event`. After adding it you may find your error log after refreshing calendar like this:

1. When logged in as USER 1
not blank

2. When logged in as USER 2
not blank 2

**Continued** ==> Do not display event / **Not Continued** ==> Display event

modules/calendar/includes/get_provider_events.php
apooravc: Multi-select for schedule facilities has another issue I think, when there you're logged in as a user and there is no row in users_facility table, the multi-select still shows user's default facility as **greyed** or selected but there's no mention in that table.

It's because of `<option <?php echo in_array($frow['id'], $ufid) || $frow['id'] == $iter['facility_id'] ? "selected" : null ?> value="<?php echo $frow['id'] ?>"><?php echo htmlspecialchars($frow['name']) ?></option>` where $iter['facility_id'] is id of default facility of selected user.

aethelwulffe:
Heh, have not gotten that far yet.
Changed all my calendar events from facility id 1 to facility id 3 yesterday, and the thing crashed.  I cleared the logs today and got 42 megs of error logs in less than a minute.
image
I will start over with a database that has no appointments and start again.

teryhill:
@aethelwulffe turn off the notices and just get real errors for right now.

apooravc:
@aethelwulffe Were the events displayed on Calendar correctly according to facility check?

aethelwulffe:
Closing this for a squashed update.

Issue #1190, opened 2018-06-23
Getting error when statement print type is Modern/Images


teryhill:
Getting the following error

[22-Jun-2018 14:43:47 America/Kentucky/Louisville] PHP Fatal error:  Uncaught Exception: TCPDF ERROR: Could not include font definition file: courier in C:\xampp71\htdocs\LibreEHR\modules\html2pdf\vendor\tecnickcom\tcpdf\tcpdf.php:2927
Stack trace:
#0 C:\xampp71\htdocs\LibreEHR\modules\html2pdf\vendor\tecnickcom\tcpdf\tcpdf.php(4275): TCPDF->Error('Could not inclu...')
#1 C:\xampp71\htdocs\LibreEHR\modules\html2pdf\vendor\tecnickcom\tcpdf\tcpdf.php(4436): TCPDF->AddFont('courier', '', '', true)
#2 C:\xampp71\htdocs\LibreEHR\modules\html2pdf\vendor\setasign\fpdi\fpdf_tpl.php(303): TCPDF->SetFont('courier', '', 10)
#3 C:\xampp71\htdocs\LibreEHR\modules\html2pdf\_class\parsingCss.class.php(340): FPDF_TPL->SetFont('courier', '', 10)
#4 C:\xampp71\htdocs\LibreEHR\modules\html2pdf\html2pdf.class.php(4239): HTML2PDF_parsingCss->fontSet()
#5 C:\xampp71\htdocs\LibreEHR\modules\html2pdf\html2pdf.class.php(1275): HTML2PDF->_tag_open_PRE(Array)
#6 C:\xampp71\htdocs\LibreEHR\modules\html2pdf\html2pdf.class.php(1244): HTML2PDF->_executeAction(Array)
#7 C:\xampp71\htdocs\LibreEHR\modules\htm in C:\xampp71\htdocs\LibreEHR\modules\html2pdf\vendor\tecnickcom\tcpdf\tcpdf.php on line 2927

apooravc:
@teryhill Could be closed now.

PR #1191, opened 2018-06-23
Restoring missing fonts

teryhill:
Restoring missing fonts and files. #1190


aethelwulffe:
Hoped we could pare that down...but here we go.  No sense waiting to merge.

PR #1122, opened 2018-05-15
Updater for LibreEHR using github API

naveen17797:
screenshot from 2018-05-15 16-36-47

@nileshprasad137 @tmccormi @teryhill @aethelwulffe 

first commit : 
started making the updater, placed the icon in the right bottom corner of the application as floating icon.

PR #1194, opened 2018-06-24
Fix iziModal - updater opening

apooravc:
If you inspect element on @naveen17797 's updater-icon, you can see the `<i class="fa fa-refresh"></i>` which is used for updater button's icon from font awesome library. Now, the tab refresh icon element is `<span class="fa fa-refresh tab-button" data-bind="click: tabRefresh" title="Refresh"></span>`. They both have common class fa-refresh because they both have same refresh type icon.

So, when we close iziModal, the function with property `onClosed` below finds all such elements on screen with `fa-refresh` class which in this case includes all refresh icons of tabs which are open as well as updater icon and then clicks on all of it using parent.$(".fa-refresh").click();. Eventually, we are not only directing all tabs to refresh upon closing of iziModal but **also updater to open up**, which is the current issue.

This PR adds a `class = "tab-refresh-icon"` to refresh icon element of tabs, so that by using `parent.$(".tab-refresh-icon").click()` instead of `parent.$(".fa-refresh").click()` in iziModal code (in `initIziLink(link)` function) below:

`onClosed: function () {`

                      setTimeout(function () {
                          parent.$(".fa-refresh").click();
                       }, 300);

`}`

would refresh all open tabs upon closing of iziModal and also not trigger opening of updater. This `initIziLink(link)` function is used whenever dynamic links are required in iziModal and `onClosed` property is used to refresh all open tabs upon closing of iziModal which is necessary in case of dynamic links. @muarachmann 

@teryhill I tested this after changing class (in `parent.$(".fa-refresh").click();`) in case of both log window in audit feature and also edit facilities iziModal, works. Please have a look.

### DIFF
--- a/interface/main/tabs/templates/tabs_template.php
+++ b/interface/main/tabs/templates/tabs_template.php
@@ -30,7 +30,7 @@
             <div class="body_title tabControls">
                 <span class="tabSpan bgcolor2">
                     <span  data-bind="text: title, click: tabClicked, css: {tabHidden: !visible()}"></span>
-                    <span class="fa fa-refresh tab-button" data-bind="click: tabRefresh" title="Refresh"></span>
+                    <span class="fa fa-refresh tab-refresh-icon tab-button" data-bind="click: tabRefresh" title="Refresh"></span>
                     <!--ko if:!locked() -->
                         <span class="fa fa-unlock tab-button"  data-bind="click: tabLockToggle" title="Unlock"></span>
                     <!-- /ko -->


muarachmann:
yop didnt see this coming in the future having a look @apooravc

muarachmann:
@apooravc  could you give us a sample case when this occurs if possible series of events. Cause generally on a refresh tab click(any) updater is to open but thats not the case

aethelwulffe:
Kinda hard to merge this one into @naveen17797 the main PR unless naveen implements it himself.

aethelwulffe:
*Maybe if we did more test result reviews and got more things merged, we could integrate stuff and not be stomping on each-other's stuff?*

naveen17797:
@aethelwulffe @teryhill we can merge the pr #1122 , i can do the bug fixes if any found.

teryhill:
@naveen17797 @tmccormi is going to review it today then it will be merged.

apooravc:
Since this and facility iziModal merged now, after @naveen17797 updater and @muarachmann user iziModal are also merged, we may change `parent.$(".fa-refresh").click()` to `parent.$(".tab-refresh-icon").click()` at required places to avoid the iziModal-updater conflict.

naveen17797:
If i change the font awesome icon to image icon will this fix all conflicts?

apooravc:
You may @naveen17797 Since, there's a possibility of conflict happening again, it'd be better to not use a general class like `.fa-refresh` to refresh tabs and instead use a custom reference to it.

aethelwulffe:
Heh...I actually like even naming files for whatever feature they are for.  I wish our clinical forms looked for '%_new.php' instead of having dozens of files named "new.php".

aethelwulffe:
I think this discussion is done for the conflict avoidance/resolution.  Closing.

PR #1196, opened 2018-06-27
Statement download name change

teryhill:
Statement download name change from the old file to libreehr

### DIFF
--- a/sites/default/statement.inc.php
+++ b/sites/default/statement.inc.php
@@ -24,7 +24,7 @@
 
 // The location/name of a temporary file to hold printable statements.
 // May want to alter these names to allow multi-site installs out-of-the-box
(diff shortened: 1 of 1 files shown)


aethelwulffe:
I will change that one, more like remove it completely.

sites/default/statement.inc.php
```diff
@@ -34,7 +34,7 @@
  *  2.  Branded Statement, whose core is build from 1., the original statement, using HTML2PDF.
  *
  *      To customize 2., add your practice location/images/practice_logo.gif
- *      In the base/default install this is located at '/openemr/sites/default/images/practice_logo.gif',
+ *      In the base/default install this is located at '/libreehr/sites/default/images/practice_logo.gif',
```
aethelwulffe: that is practice_logo.png now.

PR #1198, opened 2018-07-01
Add appointment's tooltip display styles

apooravc:
Added 4 display styles for an appointment's tooltip, like an appointment's slot text has. Also, in case of provider events, there wasn't a check for including comment, if there is one, in event's tooltip which is added in this commit. 

@teryhill Please have a look. I also noticed that the event's category+comment becomes a part of hyperlink along with name, probably because of including them in `$e['title']` instead of `$e['e_info']` in switch cases in get_provider_event.php:
text

Could you test if it's appearing like above on your machine as well? Including status, name in hyperlink or in `$e['title']` and rest - event's category+comment in `$e['e_info']` would separate them and all of them would be governed as per Appointment Display Style global.


aethelwulffe:
I will pull this today.

aethelwulffe:
We are looking to see what can be come up with for implementing tooltips in cool and useful ways.

apooravc:
@teryhill @aethelwulffe There's a tooltip plugin [qTip2](http://qtip2.com/demos) which has a cool layout as shown below:
qtip

So, the details of an appointment could be represented in above layout with image of patient on left side and description of appointment with its details - category, status, comment, facility on right side. Should we use it or implement a layout like that internally?

teryhill:
Can you bring the asset in first and let us see how large it is? Any other places that you can think to use it?

apooravc:
@teryhill qTip's CSS file is 2KB and JS file is 26KB and are added with last commit. It can be used to display simple info. texts (as in Globals), HTML (with images and links in Calendar appts.) and also supports AJAX which may be used if we want to display some dynamic medical info./data from web. We may use this as a default tooltip plugin across EHR or improve layout of current tooltip used in Calendar.

Also, this commit changes appt. display format from:
text
to:
apptd
so that only status and name are displayed in the hyperlink.

apooravc:
@teryhill qTip's css file is 2 KB and js file is 26 KB and are added with last commit. It can be used as a default tooltip plugin across EHR for displaying simple info. texts (as in Globals), HTML with images, links (in Calendar) and it can also be used to fetch dynamic medical info./data from web via AJAX. Or we may improve the current tooltip layout used in Calendar.

Also, appt. display is changed from:
text
to:
apptd
so that only status and patient's name is in hyperlink, according to globals.

apooravc:
Latest commit replaces native Calendar tooltip with qTip. 
Previews:
1. Patient event tooltip with profile image:
pic1

2. Patient event tooltip without profile image:
pic2

3. Provider event tooltip:
pic3

In case of appointment (patient event) tooltips, content depends on _Appointment Tooltip Display Style_ global. @teryhill Please have a look.

aethelwulffe:
We have, or have had qtip in the code before.
This PR is pretty light.  Testing now for basic operation.

With fullcalendar project, have you inquired with them as to using qtip in their project (and why they didn't)?

aethelwulffe:
I think the old qtip was in the ZEND stuff, so nothing to dump.
In my minimal use-case test, this works nice.  I have the outstanding question based on the fullcalendar project, but otherwise this seems real good to me.

aethelwulffe:
Would someone test this and do a review Pretty Please?

aethelwulffe:
Fixed conflicts here.  Can I get a second review on this?

PR #1199, opened 2018-07-03
Event drag-resize-copy feature

apooravc:
Initial commit adds the ability to drag and resize events as well as the ability to save the changes made and update corresponding parameters in DB. @teryhill Please have a look.


apooravc:
I'm working on event copy/paste part. Also, I'm using a simple string right now ("query executed") in case of query execution being successful. PHP logs error when query is failed and I'm trying to make an if condition on what it logs so that "query executed" or "query failed" can be sent as response. Any thoughts on how to check what it logs?

modules/calendar/drag_copy_event.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+require_once("../../interface/globals.php");
+require_once("$srcdir/htmlspecialchars.inc.php");
+
+if ($_POST['action'] == "drag") {
+  // event is dragged - update date, start time, end time and provider id
+  sqlStatement("UPDAT libreehr_postcalendar_events
```
apooravc: @teryhill No "E" in "UPDATE" here.

teryhill:
@apooravc this works great, The only issue I see is the recurring appointments. We need to figure out how to break the one that is being moved or changed. Prompt them like in add_edit_events when a recurring appointment is changed.

aethelwulffe:
Update, Updat, Upgrade, Uppgraydd.  whatever.

Recurring appointments are always a pain.

aethelwulffe:
Resize does not "stick".  This is real useful.  What is your status @apooravc ?

apooravc:
Changes done in latest commit:
1. Added ability to copy & paste events. To do so, press control key and then drag an event - it will be copied to a clone event which could be dragged to a particular slot on Calendar. For pasting it there, drop clone event and then leave control key. If pressing control key is not making a clone, refresh Calendar tab and try again.

2. Added manage tracker status in event-drag feature because we're changing date or start time in case of patient events, which need to be tracked.

3. Removed start time from event-resize because only end time and duration are changed.

@teryhill Please have a look.

apooravc:
Changes done in latest commit:
1. Added ability to copy & paste events. To do so, press control key and then drag an event - it will be copied to a clone event which could be dragged to a particular slot on Calendar. For pasting it there, drop clone event and then leave control key. If pressing control key is not making a clone, refresh Calendar tab and try again.

2. Added manage tracker status in event-drag feature because we're changing date or start time in case of patient events, which need to be tracked.

3. Removed start time from event-resize because only end time and duration are changed.

@teryhill Please have a look.

apooravc:
@teryhill I was looking to find a solution for changing start time on resize and found that this is an open issue in [FullCalendar repo](https://github.com/fullcalendar/fullcalendar/issues/374). Seems like it's not yet a feature in FullCalendar itself.

teryhill:
I noticed my google calendar does not do it either, That will be on hold until.

apooravc:
@aethelwulffe added license headers in last commit.

teryhill:
@apooravc Could you make the paste a right click. Just have the menu be right click.

apooravc:
@teryhill I initially tried to make paste option come up in right clicking an empty day cell just like in case of events and copy option. The latter works by using [eventRender](https://fullcalendar.io/docs/eventRender) callback which attaches `on("contextmenu", function(e) {}` (triggered after clicking right and before default menu appears) to all events when they are rendering. Since `eventRender` gives you event details, it can be used to pass the details of event which was right clicked and where it was clicked (to show custom context menu there) inside that function.

Now, [dayRender](https://fullcalendar.io/docs/dayRender) also gives you info. you need while creating a clone event at a day cell (date, start time, resource of day cell where we want to paste) and I could place `on("contextmenu", function(e) {}` inside it so that all day cells get attached with it and pass day cell details like in event copy but, `dayRender` [only gets called for **all day** slots in agenda views](https://github.com/fullcalendar/fullcalendar/issues/3715#issuecomment-306895834) and not for all day cells. This right click callback for all day cells is an [open issue](https://github.com/fullcalendar/fullcalendar/issues/2748) for now. That's why I used left click callback on day cell `select` and since, it was already used to open event panel, I made an option for creating event along with paste.

Any ideas on what could be done instead of this?

apooravc:
In 1 day agenda view, making `allDaySlot: true,`, 2 providers in Calendar:
allday
and following `dayRender` test code:
`dayRender: function(date, cell) {`
   `console.log("all day slot");`
`},`
gave in the console:
console

teryhill:
@apooravc I followed the links in that issue to pr#13 Triggering issues fix. See what you think about that solution.

apooravc:
I tested latest [release](https://github.com/mherrmann/fullcalendar-rightclick/releases) v1.9 of [fullcalendar-rightclick](https://github.com/mherrmann/fullcalendar-rightclick) with this code:
`dayRightclick: function(date, jsEvent, view) {`
            `console.log("day right click");`
            `console.log(date);`
            `// Prevent browser context menu:`
            `return false;`
`},` 
but, got this error:
hitspan. 
This is an open [issue](https://github.com/mherrmann/fullcalendar-rightclick/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+getHitSpan) for now. Going through that issue, I found that `getHitSpan` was removed in v3.5 of FullCalendar but, there is also a solution in that issue which I think could be applied to FullCalendar v3.0+, testing it.

modules/calendar/fullcalendar-rightclick-1.9/fullcalendar-rightclick.js
```diff
@@ -0,0 +1,68 @@
...
+ * fullcalendar-rightclick v1.9
+ * Docs & License: https://github.com/mherrmann/fullcalendar-rightclick
+ * (c) 2015 Michael Herrmann
+ */
+
+ // Note: This file has been modified from v1.9 to support dayRightclick
+ //       in versions of FullCalendar used in EHR. Please see comment at
+ //       https://github.com/LibreHealthIO/lh-ehr/pull/1199/#issuecomment-412256963 and
+ //       https://github.com/mherrmann/fullcalendar-rightclick/issues/14 which is the issue used in modification
+
```
apooravc: @teryhill Please do see this note.

apooravc:
@teryhill Added fullcalendar-rightclick support for day cells and also restored left-click on day cell to open event panel action. Now, user can right-click on an event to copy and right-click on a day cell to paste. Please have a look.

aethelwulffe:
Up for re-test.  Sorry guys, I have been hurting an awful lot lately, and unable to use a keyboard much.

apooravc:
@aethelwulffe Sorry to hear, please take your time.

aethelwulffe:
@teryhill  waiting on a second approval.  Expanding the events seems to work good.

PR #1201, opened 2018-07-06
GSoC 2018 : Calendar Improvements-Repeat parameters in Calendar Admin

### LINKED ISSUE #767: Calendar Admin - can't set repeating factors for Provider event categories
Like it says, the Calendar Admin tool will let you create a new category but it doesn't accept repeat parameters in either Patient or Provider categories.
pc-caladminnorepeat
pc-provcatnorepeat
You have to go to the calendar, click on the time block (or appt) header and edit the event, adding the repeat factors.

apooravc:
Fixes #767 

Repeat parameters are added in Calendar Admin so that a set of repeat parameters settings can be stored with respect to an existing category or new category and can be accessed as default settings of repeat parameters corresponding to the selected category in patient or provider event panel.

Preview:
rp

Work done:
1) Added repeat parameters in calendar admin (admin.php) as they were in event panel (add_edit_event.php).
2) Made some changes in admin.php and admin_helper.php so that they store and display settings as selected by user.
3) Made some changes in add_edit_event.php and appointments.inc.php so that event panel displays default settings in repeat parameters in case of new event corresponding to selected category.

@teryhill Please have a look.

### DIFF
--- a/modules/calendar/add_edit_event.php
+++ b/modules/calendar/add_edit_event.php
@@ -1033,7 +1033,8 @@ td { font-size:0.8em; }
  // var rectypes  = new Array();
 <?php
  // Read the event categories, generate their options list, and get
- // the default event duration from them if this is a new event.
+ // the default event duration & default recurring parameters corresponding to selected category
+ // from them if this is a new event.
  $cattype=0;
  if($_GET['prov']==true){
   $cattype=1;
@@ -1046,6 +1047,7 @@ td { font-size:0.8em; }
   $thisduration = $row['pc_alldayevent'] ? 1440 : round($row['pc_duration'] / 60);
  }
  while ($crow = sqlFetchArray($cres)) {
+  // loop through all rows of table libreehr_postcalendar_categories
   $duration = round($crow['pc_duration'] / 60);
     if ($crow['pc_end_all_day']) {
         $duration = 1440;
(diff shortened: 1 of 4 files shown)


apooravc:
@teryhill Added a few things to fix some issues. Expected behavior:
1. In case of new events, set recurring parameters according to values set in admin upon opening of event panel (default category) or when category changes.
2. In case of existing events, set recurring parameters according to values set in table libreehr_postcalendar_events upon opening of event panel (selected category) which is already the case but set recurring parameters according to values set in admin when category changes.

teryhill:
@tmccormi can you see if this is the way you want this to work?

aethelwulffe:
OK.  Looked at this, and it seems to:
1.  Work without crashing.
2.  Solve the use-case outlined by htuck.

modules/calendar/add_edit_event.php
```diff
@@ -55,10 +55,22 @@
...
+ $date          = $_GET['date'];        // this for new events
+ $userid        = $_GET['userid'];      // this for new events
+ if ($_GET['catid']) {
+    // when existing event
+    $default_catid = $_GET['catid'];
+ } else {
+  // when new event
+  if ($_GET['prov']) {
+    // selecting provider tab
+    $default_catid = '2'; // In Office as default category
```
aethelwulffe: We really should link these to variables saved...probably in the category table where a row is marked as the clinic default....or maybe in the user prefs table for default category.  This would avoid issues that happen when all the default categories are changed (or an existing database is used) that has totally different indexes.

apooravc:
@aethelwulffe added license headers in last commit.

aethelwulffe:
I have not been looking at the attribution on files lately.  Ooops.  Good catch. 
-Folks, this is ready for further review.

modules/calendar/add_edit_event.php
aethelwulffe: I *think* there is somewhere else that sets a calendar category as default, at least at the calendar level...or there used to be.  Maybe in the practice/calendar/Category creation UI?  Probably best to look at the table for a value.

apooravc:
@tmccormi Please have a look.

aethelwulffe:
@teryhill @tmccormi @Trodrige @pri2si17-1997 @Perseus and everyone else.  Please check this.

aethelwulffe:
Fixed conflicts.  Second test by someone else please!

aethelwulffe:
Apparently there is an issue here.

PR #1205, opened 2018-07-13
Provider scroll in Calendar's agenda views

apooravc:
Continuing #1203, initial commit adds a horizontal scroll bar for provider scroll in agenda views of Calendar. It works by making Calendar slot table overflow its container. One issue was disappearance of scroll bar because of height of `.fc-scroller` and had to add a function to make sure it's visible in all possible states.

@teryhill Please have a look.

### DIFF
--- a/modules/calendar/index.php
+++ b/modules/calendar/index.php
@@ -51,7 +51,8 @@ require('includes/session.php');
     }
 
     .fc-agenda-view > table {
-        width: 1200px;
+        width: var(--col-width, 1200px);
+        overflow-wrap: break-word;
     }
   </style>
 </head>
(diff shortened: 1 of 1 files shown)


teryhill:
The only issue I see is when there is no scroll needed we lost screen area.
before

image

After

image

apooravc:
@teryhill did the calendar shrink in size in after image? Was the scroll bar not visible in before image? Both of these images are taken after applying this PR's changes, right?

teryhill:
The before image is before the pr was applied. the after is after it was applied. They are both the same screen size. Yes the calendar shrank after the pr was applied.

apooravc:
@teryhill I think we should also test this with a large no. of providers to see if they are (Provider names) still getting crowded even after adding the scroll. About the shrinking, I tested on a new branch and it was working fine. I was able to shrink it by changing width to 600px in style block below:
`.fc-agenda-view > table {
       width: 1200px;  /* was fine with this value as in PR */
}`

Could you check in inspect element tab if its value is 1200px or something else? This one:
fc-scroll

teryhill:
@apooravc  

image

aethelwulffe:
I tested one branch today so far, and I will do the "tons o' providers" visual and function later today.

apooravc:
@teryhill @aethelwulffe Latest commit adds a function to prevent crowding of columns based on number of providers in agenda views of Calendar.

apooravc:
@aethelwulffe Tested this with a 1000 providers, display was fine in agenda views,

aethelwulffe:
Outstanding.  On my list for today now that I am going to be able to suck up the pain long enough to get something done.  1000 providers sounds awesome!

PR #1208, opened 2018-07-18
fix for folders with dot causing bug in updater

naveen17797:
fixes a bug which causes folders to appear as files in updater download module

### DIFF
--- a/updater/lib/updater_functions.php
+++ b/updater/lib/updater_functions.php
@@ -205,8 +205,10 @@ function replaceFile($filename, $original_name, $status, $old_name) {
 	if ($status == "added") {
 		//since Added there is no need to unlink the old file,it does not exist
 		checkDir($original_name);
-		copy($filename, $original_name);
-		echo $status;
+		echo $original_name;
+		echo "<br/><br/>";
+		//copy($filename, $original_name);
+		
 	}
 	if ($status == "modified") {
 		unlink($original_name);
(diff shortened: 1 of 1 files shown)


naveen17797:
still not working, need to figure the error caused when applying #1180 , dont merge or test it for now

aethelwulffe:
This is actually interesting, but Otay! :)

naveen17797:
ready for testing, @teryhill @aethelwulffe 
finally found what caused that.

aethelwulffe:
I will run this test and post my findings from Monday's round of tests this afternoon.

aethelwulffe:
I didn't get back to this.  Sorry.
Test was sat for updater.
What specific directory names have dots?  We should not really allow those anyway.  Our asset dir should use the naming convention for js stuff (commonly has dots) as hyphens.

naveen17797:
actually there is a directory with dot in PR #1180 
 assets/js/Buttons-1.5.1
this make the updater think that it was a file, since pathinfo($filename, PATHINFO_EXTENSION) returned a extension of that folder(which i didnt expect, i thought it would only return extesnion when it is a file), so i altered the logic,
you can test by going to developer mode
choose #1180  and applying it @aethelwulffe

aethelwulffe:
Our naming convention (remarked when that PR was put out) is to replace 
dots with hyphens or underscores.  headers.inc.php kinda demands that 
for at least one of the functions.

naveen17797:
finally tutorial for the updater
https://www.youtube.com/watch?v=f42BiTp75Z8

aethelwulffe:
PERFECT!

PR #1209, opened 2018-07-22
Provider at multiple facilities.

apooravc:
The facility select drop-down in Calendar's sidebar uses `getUserFacilities` function and table users_facility to get scheduled facilities of logged in user. Without global _Restrict Users to Facilities_ turned on, this feature won't work properly. If we are logged in as Admin, and we select two facilities A and B as schedule ones for Admin, then only these two would be shown in that drop-down along with "All facilities". 

When "All facilities" is selected, which is always by default, it shows all providers in sidebar and after selecting a particular facility from drop-down, only providers having that facility as schedule ones are shown in sidebar and on Calendar. This select option was not shown when logged in user had only one facility as scheduled one. Fixed this issue and now, if we select facility A and B as schedule ones for another provider, then this other provider would be shown in sidebar in both cases - when we select facility A and also when we select facility B in drop-down.

Therefore, now we can have a provider at multiple facilities. If the global isn't turned on, then this will work w.r.t to default facility and not scheduled facility and hence, a particular provider would only be shown in sidebar when that provider's default facility is selected from drop-down, limiting that provider to be at one facility only. Before testing, please make sure to delete all rows in users_facility and re-establish relations b/w providers and schedule facilities (similar to the one described in this PR).

@teryhill Please have a look.

### DIFF
--- a/modules/calendar/index.php
+++ b/modules/calendar/index.php
@@ -60,15 +60,12 @@ require('includes/session.php');
         $facilities = getFacilities();
       } else {
         $facilities = getUserFacilities($_SESSION['authId']); // from users_facility
-        if (count($facilities) == 1)
-          $_SESSION['pc_facility'] = key($facilities);
       }
 
-      if (count($facilities) > 1) {
+      if (count($facilities) > 0) {
         echo "   <select name='pc_facility' id='pc_facility' >\n";
         if ( !$_SESSION['pc_facility'] ) $selected = "selected='selected'";
         echo "    <option value='0' $selected>"  .xl('All Facilities'). "</option>\n";
-
         foreach ($facilities as $fa) {
             $selected = ( $_SESSION['pc_facility'] == $fa['id']) ? "selected" : "" ;
             echo "    <option style=background-color:".htmlspecialchars($fa['color'],ENT_QUOTES)." value='" .htmlspecialchars($fa['id'],ENT_QUOTES). "' $selected>"  .htmlspecialchars($fa['name'],ENT_QUOTES). "</option>\n";


aethelwulffe:
This is good.

I do have an idea that may help us with the complexity as to what a "facility" is though.

I think setting things up so that facilities have a hierarchy would help.  
A clinic may have multiple facilities.  These are not necessarily different physical locations.  They could be associated with Group Identifiers (Group NPI's in the US an others) that annotate different treatment programs.  They could also be just place of service code differences.  "Home" is a facility (POS code 06 I think).  Inpatient and outpatient are codes 21 and 22 also, but those might not be distinctions with a difference for excluding/including a particular provider.  They are logistical differences perhaps, but probably don't change a provider access.  It might be good to have the option to have things somehow be "sub-facilities", so you wind up with different POS codes or alternate Facility NPI numbers, but you don't have a laundry list of different facilities.

teryhill:
Before we jump to hierarchy let's get the changes that are being done in the code base and tested and deployed.

aethelwulffe:
Ja, Ja.  Deleting some facility rows and testing now....
-Just a thought bro.

teryhill:
@apooravc  I tested with out this change and it appears to be working. So is this change needed?

apooravc:
@teryhill I only did a minor change in this PR which is the case when logged in user has only one schedule facility and facility select option would not appear in sidebar. Also, I think that global not being checked was the reason it wasn't working before. Was the global checked when you tested?

teryhill:
No I did not have it checked.

apooravc:
@teryhill I tested this case: selected two facilities as schedule ones for a provider and when I selected each of them in select option, that provider was there in sidebar in both cases but only when global was checked and it appeared in default one only when global was not checked.

apooravc:
@teryhill Closing this. Seems like that global needs to be checked for many features in Calendar to work properly.

PR #1216, opened 2018-07-25
changed the +'s to .'s

teryhill:
Was getting warnings in error log with php 7


aethelwulffe:
Normal operators.  This is good.

PR #1220, opened 2018-08-04
Patch : fixing updater logic

naveen17797:
logic mistake got fixed
- WIP : to improve the performance

### DIFF
--- a/updater/updater.php
+++ b/updater/updater.php
@@ -113,7 +113,7 @@ else {
 		//it means the token is expired, so ask for re-entry
 		$loader->set_template_file("updater_screen_three");
 		$updater_token_string = "Your $updater_host User Access Token Was Expired, Re-enter it now";
-		$loader->assign("HOST",xlt($updater_host_string));
+		$loader->assign("HOST",xlt($updater_token_string));
 		$loader->output();
 	}
 	elseif (isTokenValid($updater_token)) {
@@ -122,7 +122,7 @@ else {
 		$user_name = $owner_arr['login'];
 		$upcoming_prs = getUpdaterSetting("upcoming_prs");
 		if (getUpdaterSetting("github_current") != "empty_setting") {
-		$pull_request_number = getUpdaterSetting("github_current");
+			$pull_request_number = getUpdaterSetting("github_current");
 		}
 		else {
 		//load settings from json file, this pr number refers to the pr number at which the updater gets merged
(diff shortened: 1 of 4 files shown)


naveen17797:
@teryhill, @aethelwulffe , @tmccormi  
This patch is ready for review

updater/updater.php
```diff
@@ -113,19 +113,40 @@
 		//it means the token is expired, so ask for re-entry
 		$loader->set_template_file("updater_screen_three");
 		$updater_token_string = "Your $updater_host User Access Token Was Expired, Re-enter it now";
-		$loader->assign("HOST",xlt($updater_host_string));
+		$loader->assign("HOST",xlt($updater_token_string));
```
tmccormi: Translation of this string should occur in the construction of the string on line 115.
ie: $updater_token_string = xlt("Your") .' '. $updater_host .' '. xlt("User Access Token Was Expired, Re-enter it now")

Otherwise a translation into another language will not work.

tmccormi: Even that is iffy for RTL languages and some other constructs.  but be do what we can.

updater/updater.php
```diff
@@ -146,19 +167,18 @@
...
 					$extension = pathinfo($ke['filename'], PATHINFO_EXTENSION);	
 					array_push($files_need_to_be_downloaded, $sha);	
 				}
-			}
 			$count_files = count($files_need_to_be_downloaded);
 			$loader->assign("STATUS", xlt("status"));
-			$loader->assign("UPDATER_START", "<b>$cpr_title</b><br/><h6>$cpr_body</h6>");
+			$loader->assign("UPDATER_START", "<b>$cpr_title(#$cpr_number)</b><br/><h6>$cpr_body</h6>");
 			$loader->assign("COUNT_FILES", $count_files);
 			$loader->assign("PR_NUMBER", xlt("Write Feedback For #".$pull_request_number));
```
tmccormi: Same here.  translate tool will not work with mixed variable data

naveen17797: working on it @tmccormi  :relaxed:

aethelwulffe:
@naveen17797 @muarachmann There are translation engine issues here, but the translation engine is really totally broken anyway.
Translations with variables should be turned into statements like:
xlt("Your variable name is:").$variable
instead of,
xlt("Your ").$variable.xlt("name is.")
HOWEVER, translations MUST be a separate issue to make them work at least a little bit, then we can hold up stuff like this.
What do you folks think about merging this?

PR #1221, opened 2018-08-04
WIP : Db backup Module for the updater

naveen17797:
dont test it, since there is a bug in updater fixed in #1220, it need to be merged to test this


aethelwulffe:
@naveen17797 Status report please?

PR #1195, opened 2018-06-26
GSoC 2018: Setup Improvements Continue(libreehr)

muarachmann:
Rewriting the scripts lost some files in the process of rebasing , deleting and moving repos thank God i can find my way out. 
A cleaner branch and rapid commits like never before on their way

PR #1207, opened 2018-07-18
Izi Refactoring(Administration menu)

muarachmann:
@aethelwulffe here is a cleaner branch to clarify the #1136  will push them gradually

PR #1222, opened 2018-08-06
GSOC 2018 Setup improvements #1195 (rebased and squashed)

tmccormi:
This is based on PR #1195.

I rebased it to current dev tip master and squashed all the commits so it would be easier to review and merge cleanly,


muarachmann:
@tony do i commit here or on the other PR #1195 ? i have some few modifications to add just a little

tmccormi:
@muarachmann You can keep using your own branch, this is just for ease of review and testing.   I'll do this again when we merge for real into the main line.

aethelwulffe:
Thanks Tony.

aethelwulffe:
@muarachmann 
We want to merge this if you think you can handle the URL adjustments and such....  Please?
You can run this back into a branch of your own too, so you get the marks on the scoreboard when we merge your stuff.  Gotta be looking to your CV portfolio.

modules/setup/includes/settings.inc.php
```diff
@@ -0,0 +1,62 @@
...
+    $billingDirectory2 = $LIBRE_SITE_DIR."/era";
+
+    $billingLogDirectory = $LIBRE_SITE_DIR."/logs";
+    $lettersDirectory = $LIBRE_SITE_DIR."/letter_templates";
+    $gaclWritableDirectory = dirname(__FILE__)."/../../../gacl/admin/templates_c";
+
+
+    // Record name of php-gacl installation files
+    $gaclSetupScript1 = dirname(__FILE__) . "/../../../gacl/setup.php";
+    $gaclSetupScript2 = dirname(__FILE__) . "/../../../acl_setup.php";
```
aethelwulffe: 7.1 forbidden (7.2?)
Also, if you have visibility of globals, there are several you can use for the root path.

modules/setup/includes/shared.inc.php
```diff
@@ -0,0 +1,30 @@
...
+// *** remote file inclusion, check for strange characters in $_GET keys
+// *** all keys with "/", "\", ":" or "%-0-0" are blocked, so it becomes virtually impossible
+// *** to inject other pages or websites
+    foreach($_GET as $get_key => $get_value){
+        if(is_string($get_value) && (preg_match("/\//", $get_value) || preg_match("/\[\\\]/", $get_value) || preg_match("/:/", $get_value) || preg_match("/%00/", $get_value))){
+            if(isset($_GET[$get_key])) unset($_GET[$get_key]);
+                die("A hacking attempt has been detected. For security reasons, we're blocking any code execution. <a href='index.php'>Click here</a>");
+                }
+            }
+?>
```
aethelwulffe: Normalize line endings please.

modules/setup/libs/fonts/fontawesome/fontawesome-webfont.svg
```diff
@@ -0,0 +1,2671 @@
+<?xml version="1.0" standalone="no"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd" >
```
aethelwulffe: fontawesome and the like are already in /fonts if you want to use them.

aethelwulffe:
@muarachmann I would like to get this in the code base before I finish up doing the xlt() and xl() translation function DB tables.

muarachmann:
roger that @aethelwulffe i pushed some commits on #1195 and that is good to go left with the urls of course. I can handle that. I guess the suplimentary hours i took during gsoc/exams is weighing me a little just came back from the hospital. Please i am done with izi_modal take a look here #1207.

muarachmann:
@pri2si17-1997 , @tmccormi , @aethelwulffe I need a device to handle multiple sites login. Here are my thoughts when a user launches the software that is navigating to localhost/LibreEHR for example, the setup checks if there is a previous installation (i mean a 1 in the sqlconfig) not just a folder/dir. If it finds one, it shows the following screen.
screenshot from 2018-08-12 11-36-24

Now depending on the number of sites you have installed still talking about the 1 in all the sites for example (default > sqlconfig and site2 > sqlconfig). It should propose to you a method(dropdown would be suitable) for you to login into the site of your choice. Now if you manually place a 0 in the config variable in any of those sites then that site should not be available though we developers might know the db still exists and the stuffs in there. I would love a device to read alll the sqlconfig files of all the folders placed in the sites directory so as to read the value of the config variable. If you chose a force install it will go the normal way. When that is done we can point the index of of site to something like header("location": modules/setup/) and from there all will be cool till the login. Please your ideas thanks

aethelwulffe:
The site login argument will be URL?site=mysite
You parse the sites directory for available subdirs that represent various sites, and you drop the "default" from the list.  Default is just...there.

muarachmann:
okay

aethelwulffe:
Did that help?
argument selection === (directories under sites-default)

Note: patient portal globals strings should create a default based on the current install and site too.  This has nothing to do with anything here, but very similar.

muarachmann:
I will see how to generate the looping. Another way could be to write to a file at each install a key valu pair something of the sort and on a destroy it removes the stuff

aethelwulffe:
@muarachmann status on this?  I think this needs to be closed?

muarachmann:
yes this can be closed will continue from the other

aethelwulffe:
OK...so closing it...

Issue #1229, opened 2018-08-23
Tag's code runs always impacting performance of certain screens 


teryhill:
The tags system is not controlled via a globals switch to turn it on and off, This causes performance hits on large systems. Started with the most obvious screen demographics.

tmccormi:
We have performance update coming for tags that we just deployed to a customer today ... just FYI

teryhill:
Does it include the ability to turn it off, I know Ken did not think the collapsing of it on the demographics screen was needed?

tmccormi:
It's a module.  Delete the directory and it's "off".  Option to turn it off is ok too, I guess but might not be needed in the near future if the reason is performance.

teryhill:
If you delete the module what happens when do_action gets called

tmccormi:
nothing, if there is no code the action just moves on.  That't the whole purpure of the do_action hooks

tmccormi:
@teryhill I think your change and Kens have both been merged now.  Closing this.

PR #1230, opened 2018-08-23
Started coding the ability to completely

teryhill:
Started coding the ability to completely turn off tags. #1229


teryhill:
While the impact is not felt on the smaller systems it is seen on the larger ones.

aethelwulffe:
This is very necessary.  Large DB performance stinks.  Someone please review so we can merge.

PR #1232, opened 2018-08-27
Improve performance of patient finder with tags on large DB

kchapple:
On a large DB, the join on the  tf_patient_tags pivot table caused big performance penalty, so I added a migration to insert an index to pid col of tf_patient_tags pivot table.

### DIFF
--- a/interface/tags_filters/migrations/Migration_2018_08_22_14_48_11.php
+++ b/interface/tags_filters/migrations/Migration_2018_08_22_14_48_11.php
@@ -1,0 +1,132 @@
+<?php
+
+use \Framework\Plugin\Migration\Migration;
+
+class Migration_2018_08_22_14_48_11 extends Migration
+{
+    public function up()
+    {
+        $directive = <<<MIGRATION
+--
+--  Comment Meta Language Constructs:
+--
+--  #IfNotTable
+--    argument: table_name
+--    behavior: if the table_name does not exist,  the block will be executed
+
+--  #IfTable
+--    argument: table_name
+--    behavior: if the table_name does exist, the block will be executed
+
+--  #IfMissingColumn
+--    arguments: table_name colname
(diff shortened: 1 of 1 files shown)


interface/tags_filters/migrations/Migration_2018_08_22_14_48_11.php
```diff
@@ -0,0 +1,132 @@
...
+--    behavior:  If the index does not exist, it will be created
+
+--  #EndIf
+--    all blocks are terminated with and #EndIf statement.
+
+--
+-- Indexes for table `tf_patients_tags`
+--
+
+ALTER TABLE `tf_patients_tags`
```
kchapple: This is really the only code. Migrations are only run once. There is no need for conditional logic in this case.

interface/tags_filters/migrations/Migration_2018_08_22_14_48_11.php
```diff
@@ -0,0 +1,132 @@
...
+--    arguments: table_name colname
+--    behavior:  If the index does not exist, it will be created
+
+--  #EndIf
+--    all blocks are terminated with and #EndIf statement.
+
+--
+-- Indexes for table `tf_patients_tags`
+--
+
```
kchapple: This is really the only code. Migrations are only run once. There is no need for conditional logic in this case.

aethelwulffe: Does the migration require any user input to run?

kchapple:
No, these migrations are run automatically. Essentially, the first request made to the server, the migration will be run. The system keeps track of the migrations that are run so they are only run once. 

Ken
kenchapple.com
facebook.com/kenchapplemusic
@ken_chapple

Check out my latest CD, Crossing the Black here: https://open.spotify.com/album/1fM4v6hJ1275fywjaREzNP?si=ohHDH2vV

PR #1234, opened 2018-09-03
This reports the amount of time that a user has been logged in. The t…

growlingflea:
…ime is calculated by either going from

login to logout (forced or not) or in the case the user closes the browser and doesn't logout, the time from login to the last
activity before the following login. Newest version of Datatables as been added to assets.


growlingflea:
Found another issue made a quick fix

PR #1236, opened 2018-09-03
    User Activity Report pulls data from the log tables and reports t…

growlingflea:
…he total time each user has been logged in.  Time is either calcualted

    from Log-in to Log-out or from log-in    to last recorded activity.  Option to see a total summary or a time breakdown.
    Datatables gives the user ability to search by username in break down.    Ability to print to PDF, csv, or Excel


growlingflea:
Reviewing I found another bug.  This should be working correctly with all the suggested edits.  More reporting to come.

teryhill:
Daniel is there a reason for closing and then opening new pr's with what appears to be the same code?

growlingflea:
Lets hold off on this PR.  I actually found a better way to do this.  I'm working on it now.

Issue #1228, opened 2018-08-17
Review of Systems and Review of Systems Checks forms not working


2208Abhinav:
1)Go to patient
2)Open any encounter
3)Under **Clinical** you will find the forms named **Review of Systmes** and **Review of Systems Checks**
![issue](https://user-images.githubusercontent.com/30211121/44255493-e8515700-a223-11e8-9557-5227646707d8.jpg)

When you fill any of these two forms and click Save nothing happens.

muarachmann:
@2208Abhinav , @aethelwulffe looking keenly at this here is the code being outputted but i cant see the action the form submits. cause 1.) there is no submit button in the form 2) there is no javascript/jquery that attempts to handle the submit like say `$("#save").click(fucntion(e){ prevent form from submiting and do some stuffs first.});` reason why that button cant give anything. Take a look at the screenshot i got 
![kazam_screenshot_00024](https://user-images.githubusercontent.com/16350814/44295209-b9ae9d00-a272-11e8-8cc9-b0d1b8ddb2da.png)

Taking a closer look will ping you if ive found something

muarachmann:
All what i see is this at the head `<form method=post action="/ehr/interface/forms/reviewofs/save.php?mode=new" name="my_form" onsubmit="return top.restoreSession()">` of which i doubt if that form is actually being submitted

muarachmann:
I dont know if what i have done is correct(i mean i was to check for validations etc). Here is what i got when the `<input id="save" type="button" class="cp-submit">`  was given a `type="submit"`. It works well and submits. Just that when nothing is cheked or placed it still submits. @aethelwulffe , @teryhill please clarify this to know the necessary form validation to be done before a submit is done to get this issue e.g i think it should check if at least one check box is being checked before it can submit successfully. My screenshots

![screenshot from 2018-08-17 20-19-24](https://user-images.githubusercontent.com/16350814/44295326-bf0ce700-a274-11e8-80f4-60e01aba1f4d.png)

![screenshot from 2018-08-17 20-19-18](https://user-images.githubusercontent.com/16350814/44295330-c633f500-a274-11e8-988f-1f74fb478232.png)

aethelwulffe:
[...]

PR #1239, opened 2018-09-07
fix Review of Systems form submit

muarachmann:


muarachmann:
Fix for #1228. @teryhill , @tmccormi i left some comments on this issue. I need a go ahead for the validations necessary or should i come up with one??. with that @2208Abhinav can continue with his report generation too.
screenshot from 2018-08-17 19-43-30

and here the review. But i dont like the idea of it saving the form even when no box is being checked. I think that should be a validation

screenshot from 2018-08-17 20-19-18

aethelwulffe:
1.  Needs spacing between output of things like "FEVER:yes"
2.  Mind checking out https://github.com/LibreHealthIO/lh-ehr-contribs/tree/Button_ros and building it up as a replacement form so we can drop this *and* the other ROS form?

aethelwulffe:
Yes, I believe so.  I have not used this form in years and years....wait....I have never really used this form...  :)

tmccormi:
@muarachmann sorry, way behind.  I'm going to test this tonight and approve it if I agree with @aethelwulffe (which I'm pretty sure I will)

PR #1242, opened 2018-09-18
Enhance passprase update with form validations/ izi notification

muarachmann:
This commit does some checks like:
- Checks if the passphrase field is empty
- Checks if the old password and new are same(denies this)
- Provides notification via toast to the user as show below.
- Tracks status of ajax request that is if successfull or not and displays appropriately to the user. 
This can be seen below

screenshot from 2018-09-18 16-35-53
screenshot from 2018-09-18 16-36-18

### DIFF
--- a/interface/main/tabs/templates/user_data_template.php
+++ b/interface/main/tabs/templates/user_data_template.php
@@ -10,7 +10,7 @@
     $notification = $unreadMessages;
 ?>
 
-<link rel="stylesheet" type="text/css" href="../../css/userdata.css"/>
+<link rel="stylesheet" type="text/css" href="css/userdata.css"/>
 <script type="text/html" id="user-data-template">
     <!-- ko with: user -->
     <nav class="nav navbar-nav navbar-right" style="margin-right: 10px !important;">
--- a/interface/usergroup/facilities.php
+++ b/interface/usergroup/facilities.php
@@ -161,7 +161,7 @@ $(document).ready(function(){
         iframeURL: "facility_admin.php?fid=" + link,
         onClosed: function () {
           setTimeout(function () {
-            parent.$(".fa-refresh").click();
+            parent.$(".tab-refresh-icon").click();
           }, 300);
         }
       });


muarachmann:
The second commit fixes the BatchCom title tab. Fix for #1243 
screenshot from 2018-09-18 17-28-12

tmccormi:
Don't mix pulls with different purposes.  Please split this into two separate pull requests.

muarachmann:
okay @tony i thought since it was a one line commit i could just squeeze it here

tmccormi:
Also make sure that any changes to the rules for Password security are optional and configurable in the Globals, there are already options for various rules there.

muarachmann:
reverting changes thoo

muarachmann:
@tmccormi  didnt change the rules. Here is what i did. I noticed the message for incorrect password was being displayed into a div. I changed that and instead used a toast to notify the user. Futhermore since this was an ajax request i just outputed json format to tract the status and message. Also did some validations client side before actually sending the ajax request. Like if the field is empty it shouldnt submit the form of which it does , if old password == new password it should either submit the form

muarachmann:
@tmccormi removed the batchCom title stuff. This should be only for passphrase now please check

interface/usergroup/user_info.php
```diff
@@ -12,32 +12,67 @@
...
     <?php
-      call_required_libraries(['bootstrap', 'jquery-min-1-9-1']);
+      call_required_libraries(array("jquery-min-3-1-1","bootstrap","font-awesome", "iziModalToast"));
       ?>
     <script language='JavaScript'>
       //Validating password and display message if password field is empty - starts
       var webroot='<?php echo $webroot?>';
       function update_password()
       {
-          top.restoreSession();
```
tmccormi: Pretty sure this top.restoreSession() is required and it be removed from you side

muarachmann:
IMO it doesn't seem to be a pb taking off. I remember asking @terry its use
and he said it was to be taken off soon or later. And I have tried to trace
its usage too from the past to no avail

tmccormi:
It is used to keep the selected / active patient data in sync between pages.  IT was put there (and everywhere) originally by the OpenEMR project because patient data was getting scrambled between pages.  If that is not an issue then fine, @teryhill  and @aethelwulffe need to approve this.

tmccormi:
This is the original post from OpenEMR explaining it:
```
PHP Sessions and Browser Windows
Before August of 2007 OpenEMR worked poorly, even dangerously, when used in multiple top-level browser windows on the same machine. The underlying problem is that most web browsers cannot support separate cookie-based sessions in this case, because all windows share the same cookie storage area.

Then a JavaScript-based solution was implemented as follows:

interface/login/login.php was modified to delete the session cookie when a new login occurs. This ensures that a unique session ID is generated for each login (we assume you start each new browser window session with a login).
A JavaScript function restoreSession() was created via the script library/restoreSession.php. This script is included in every top-level window. When called, the function restores the session cookie's value to the session ID that was supplied when the window was created.
Many other scripts (about 200!) where changed to call "top.restoreSession()" wherever a server-side script is invoked. For example, "onclick='top.restoreSession()'" was added to every "<a href=...>" tag.
The effect of all this is to make the top-level window the storage area for the session cookie. So by creating multiple top-level browser windows and logging in separately to each, you can have multiple OpenEMR sessions concurrently on your desktop. This is a valuable feature for busy users who must juggle multiple tasks at once.

Naturally this impacts future development. You must include a JavaScript call to top.restoreSession() wherever you invoke a PHP script that requires current session data (which is most of them). The most common ways of doing this are by including the onclick handler as described above, and by including "onsubmit='return top.restoreSession()'" in <form> tags.
```

muarachmann:
yop you're right filling that in back :+1:  thanks

tmccormi:
@teryhill need a 2nd ...

teryhill:
Sorry @tmccormi I have been hammered the last few weeks, Right now I am having issues with the rain and my internet. I will review this as soon as I can.

interface/usergroup/user_info.php
```diff
@@ -12,32 +12,68 @@
...
+                icon: "fa fa-warning",
+                message: 'Password field cannot be empty'
+            });
+        }else {
+
+            if(curPass === newPass){
+                iziToast.warning({
+                    title: 'Warning',
+                    icon: "fa fa-warning",
+                    message: 'Old password cannot be same with new one'
```
teryhill: Pass Phrase

interface/usergroup/user_info.php
```diff
@@ -12,32 +12,68 @@
...
+        var curPass  = $("input[name='curPass']").val();
+        var newPass  = $("input[name='newPass']").val();
+        var newPass2 = $("input[name='newPass2']").val();
+
+        //check if the password field is not null
+        if(curPass === '' || curPass === null || newPass === '' || newPass === null || newPass2 === '' || newPass2 === null){
+            iziToast.warning({
+                title: 'Warning',
+                icon: "fa fa-warning",
+                message: 'Password field cannot be empty'
```
teryhill: Pass Phrase

interface/usergroup/user_info.php
```diff
@@ -29,15 +29,15 @@ function update_password()
...
 
             if(curPass === newPass){
                 iziToast.warning({
                     title: 'Warning',
                     icon: "fa fa-warning",
-                    message: 'Old password cannot be same with new one'
+                    message: 'Old Pass Phrase cannot be same with new one'
                 });
             }else {
                 console.log("asdasd");
```
tmccormi: Oh,  I did catch this one ... this is a junk message.  Please fix to something that is meaningful.

muarachmann: Hahahaha i did that during debugging alsways fun of that. Removing that right away

tmccormi:
Closing this.  Repo unknown.   I assume a new PR is coming or has come already

PR #1248, opened 2018-10-03
Final? version of Facility access control

aethelwulffe:
Yet another PR in this theme.  I will link all precursor versions.


aethelwulffe:
Once again, this feature aged until it died...
-Something wrong ever since the master merge.

Issue #1249, opened 2018-10-04
Calendar Fatal error in master branch commit 24e2ce0


tmccormi:
Getting a fatal error related to izmodal on my Ubuntu 16.04LTS
![calendar-notworking](https://user-images.githubusercontent.com/285835/46497539-ba25d780-c7cf-11e8-9617-6a3b9a49b8c4.png)

tmccormi:
There is no "css" directory in the interface directory ... (and there shouldn't be)
GET http://localhost/lh-ehr-dev/interface/css/userdata.css net::ERR_ABORTED 404 (Not Found)

code:
```
interface/main/tabs/templates/user_data_template.php
13:<link rel="stylesheet" type="text/css" href="../../css/userdata.css"/>
```
Seems like the relative path is getting borked somewhere.

tmccormi:
@aethelwulffe you were the last person to touch that particular line of code... :-)

tmccormi:
find . -name userdata.css
./interface/main/tabs/css/userdata.css

aethelwulffe:
So you are saying the move of templates to /sites.  Gotcha.  Checking.

aethelwulffe:
Actually, never noticed this stuff.  Guess I should look and see why it would seem like I touched it in the past.  Don't recall.

aethelwulffe:
So, originally when I noticed the error, the line was .../../css/userdata.css
I looked at it, and said "oops, someone gots too many dot thingies!  Delete one, good to go!
Wrong.

aethelwulffe:
that was #1150

aethelwulffe:
Third dot had been introduced in https://github.com/LibreHealthIO/lh-ehr/commit/8f381b1cea811d07abf830849bc84067d8222883#diff-37bf90ba2e40d4f1442aa9d79a04a24a

#653 sort of indicates that this was a boo-boo in the tabs code from day one....???

tmccormi:
Path is fixed the real problem (with calendar) appears to be in the module drag and drop code by @apooravc 

![index-token](https://user-images.githubusercontent.com/285835/46503302-54414c00-c7df-11e8-97f9-42153b8f0db5.png)

tmccormi:
I don't see anything obvious in this bit of code
![drag_copy](https://user-images.githubusercontent.com/285835/46503628-34f6ee80-c7e0-11e8-9171-505571109a72.png)

tmccormi:
[...]

PR #1250, opened 2018-10-04
repair relative link

aethelwulffe:
issue #1249

### DIFF
--- a/interface/main/tabs/templates/user_data_template.php
+++ b/interface/main/tabs/templates/user_data_template.php
@@ -10,7 +10,7 @@
     $notification = $unreadMessages;
 ?>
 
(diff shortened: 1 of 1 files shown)


tmccormi:
Nope...  now the path is closer but still not correct
userdata_css

http://localhost/lh-ehr-dev/interface/main/css/userdata.css

aethelwulffe:
Yeah, I see.
Checking to see why the hell not.

aethelwulffe:
K...should work.  Can't test right now.

tmccormi:
not quite, however if you drop the leading slash it clears that particular error.    The calendar is still borked, but we are down to 1 error now...
index-token

tmccormi:
I've fixed the css path.  Looking at the index.php issue

Issue #1246, opened 2018-09-27
Calendar opens on medication outcome select (Resolved)


muarachmann:
**What issue is all about:**
The calender gets opende when the medication field outcome - resolved is selected from the dropdown which is undesirable

**Steps leading to issue:**
Select a patient > Go to medication > Add new > Click on the dropdown for outcome > select resolved

**Screenshots** 
![screenshot from 2018-09-27 21-26-01](https://user-images.githubusercontent.com/16350814/46172894-c0b5cb80-c272-11e8-952e-eec2ca59cc50.png)

**Errors from log files or console if any**
None

**Would you like to solve this issue**
No, any first time contributor can be assigned to this

muarachmann:
@aethelwulffe , @Trodrige can any of you add labels to this thanks. I would have but i aint got sufficient permissions for that

besong23:
Hello guys, I will like to work on this issue

muarachmann:
@besong23 go ahead

besong23:
Okay so after looking through the code the function outcomeClicked() turns to set the End Date value to current date so I have two ways of going about this, 
I can remove the function call as well as the function definition since it looks like the function is being called just ones OR I can remove just the function call and leave the function definition since I don't know if it might be needed later on.
Which way should I go about this?

muarachmann:
We dont want to define functions id they arent being use. There is no point to that. Do what you have to do to fix the issue and make a PR. From there we will review your code and tell you what you ought to do and not. As for finding a suitable method thats up to you 😃

muarachmann:
Oh @aethelwulffe , @besong23 my bad. I didn't think this twice. When an issue is marked resolved, it should be ended and that is why it auto fills the form with the end date marking this stuff over. The issue to resolve here is preventing the opening of the calendar but that field (end date) should be filled with the current date of filling the issue. @besong23 hope you get it now. Let me give you a tip 😺  **a focused field will draw attention to itself in all the form.**

besong23:
So that means it isn't at issue after all right?

muarachmann:
You should learn to read @besong23 I said we want the field filled but the calendar is not to popup

besong23:
Ah okay. Sorry about that :-)

muarachmann:
And i gave you a tip. Its a one line issue stuff **"a focused field will draw attention to itself in all the form."**

besong23:
@muarachmann I have opened a pull request for the issue

[...]

PR #1252, opened 2018-10-07
#1246: Calendar opens on medication outcome select <Resolved>

besong23:
Removed focus from End Date field on the Add issues form so that calendar does not pop up when outcome <Resolved> is selected.

### DIFF
--- a/interface/patient_file/summary/add_edit_issue.php
+++ b/interface/patient_file/summary/add_edit_issue.php
@@ -552,7 +552,6 @@ ActiveIssueCodeRecycleFn($thispid, $ISSUE_TYPES);
    var today = new Date();
    f.form_end.value = '' + (today.getYear() + 1900) + '-' +
     ("0" + (today.getMonth() + 1)).slice(-2) + '-' + ("0" + today.getDate()).slice(-2);
-   f.form_end.focus();
   }
  }
 


interface/patient_file/summary/add_edit_issue.php
```diff
@@ -552,7 +552,6 @@ function outcomeClicked(cb) {
    var today = new Date();
    f.form_end.value = '' + (today.getYear() + 1900) + '-' +
     ("0" + (today.getMonth() + 1)).slice(-2) + '-' + ("0" + today.getDate()).slice(-2);
-   f.form_end.focus();
```
muarachmann: 👍 this does the trick. Please always describe your PR what it does a suitable name would be **Fix: issue title** no need placing the issue number cause github references issues when they are related or mentioned in PRs.
Approving this

muarachmann:
@aethelwulffe , @teryhill this is ok have a look

interface/patient_file/summary/add_edit_issue.php
besong23: Okay, will keep that in mind

aethelwulffe: Nice description like "Changed focus thingie on my first step to becoming a major daily contributor to LibreHealthEHR" or something like that is usually good to have.

besong23: I have added a description to my pull request. Will remember to always do so.

PR #1253, opened 2018-10-10
GSoC 2018: Setup Improvements Continue(libreehr) ** New

muarachmann:


muarachmann:
I prefer tony keeps his squashed branch i will continue the process from here with the subdir looping thing

tmccormi:
141 files for a setup process?   is all that bootstrap stuff needed?

muarachmann:
@tmccormi i had myself a separate module not to worry at the end i will point all to the asset folder

muarachmann:
@aethelwulffe found a way to get all the sqlconfig files. Now it is question of reading the config variable. I need a situation where if the config variable is 0, it should consider that the setup is not installed in this dir and when 1 we detect it in the dropdown list. 
Any ideas would help thanks

modules/setup/includes/functions.inc.php
```diff
@@ -90,4 +90,54 @@ function write_bashScript($sys){
...
+                // Check if the line contains the string we're looking for, and print if it does
+                if(strpos($line, $search1) !== false){
+                    $val = substr($line,9,2);
+                    echo $val;
+                }else{
+                    if(strpos($line, $search2) !== false){
+                        $val = substr($line,9,2);
+                        echo $val;
+                    }
+                }
```
muarachmann: Suceeded in getting the stuff and values now i will place the select dropdown just above this screen below.
screen shot 2018-10-21 at 11 40 20 pm

Also art the force installation button should ask which site you would love to delete or re-install cause you might have default, clinic that are ok but default4 which maybe you messed up with the db. And you want a re-install there. If this is ok please tell me. Will push more commits tomorrow

muarachmann: The 110 are the config variables found in the default site, three and two if i am not mistaken. I placed them there to see the recursiveness will remove them once done. So here is my idea where it detects a one in the sqlconfig, it denotes it as an installed site and vice versa for the other. The 110 denotes there is an active installation in default/ and in three/ so the dropdown will comprise of these. and not of two. But with the force install you can override any directory , or wipe all something of the sort waiting for our ideas on that 😃

modules/setup/includes/functions.inc.php
```diff
@@ -129,11 +129,11 @@ function findString($filename) {
                 // Check if the line contains the string we're looking for, and print if it does
                 if(strpos($line, $search1) !== false){
                     $val = substr($line,9,2);
-                    echo $val;
+                    return $val;
```
muarachmann: Ehhhmm @aethelwulffe here is what i do i have done a recursive read on the sites dir to look for all folders containing the sqlconfig file **first**. When i grab them i now do a read in each file to get the value of the `$config` variable. This permits me to know which site has a 1 or 0 so i know which installations is/are active or not. If cause it is no point asking someone to login with when the config variable has changed. With this set the dropdown would be implemented and you on select you would be redirected to `localhost/lh-ehr?site=<folder_name>`. Now in my force install method. You would be prompted all the available sites whether installed or not installed. Here the role of this is just to write a zero into the config variable and continue the process of installation normally

muarachmann:
Here is how it goes and looks like
screen shot 2018-10-24 at 4 47 46 am

screen shot 2018-10-24 at 4 48 12 am

muarachmann:
Success on my end all set. At least the main features multi-site login implemented. I just tried with only an sql file and a config file in a directory look at the output. 
screen shot 2018-10-24 at 5 09 32 am

And on login 

screen shot 2018-10-24 at 5 08 57 am

aethelwulffe:
@muarachmann  where are we here (sorry I was gone so long!)?  Is this a "run a normal setup from fresh install, then use a musty old install and see what happens" sort of test?

aethelwulffe:
What does "re-install" do?  Wipe out your database?  Is this actually meant to say "Add a new clinic/database/site thingie"?

aethelwulffe:
Running test on this now.

muarachmann:
Thanks @aethelwulffe let me work on this feature. My gsoc joy is complete :) lol

aethelwulffe:
Sorry it took so long.  One, I was hoping you would just take over the project/political BS completely and let me pop in on occasion.  Two, I have been very busy.  Three, I am lazy.

muarachmann:
Hahahaha now i can proceed with it cool by moving all assests used in my setup to the assets folder. I did this as a separate module. But now time to do house keeping. I got this

aethelwulffe:
W00T!

PR #1261, opened 2018-10-25
Added an Installation using MAMP section and fixed broken links

punwai:


punwai:
@muarachmann I messed up and changed sqlconfig on one of the earlier commits (mac added). I fixed it by manually copy and pasting an old version of the file onto sqlconf.php and then pushing it in Update sqlconf.php. Is this OK or do I have to do something else.

Issue #1265, opened 2018-10-29
System allows creation of multiple encounters on a DOS 


teryhill:
Currently the system allows creation of multiple encounters on a DOS when auto creating an encounter. If you set the status of an appointment to arrived and have the auto generate encounters checked in globals the encounter will be created, however if you change the status to something else and save then go back in and change the status to arrived it will create another encounter.

tmccormi:
Is the fix just for that problem?  IE: Auto encounters, because creating more than one encounter per visit (manually) is normal for many practices

teryhill:
Yes just for auto encounters. They should only create i encounter per appointment automatically.

tmccormi:
merged

PR #1266, opened 2018-10-29
Fix for the autocreate encounters

teryhill:
Fix for the autocreate encounters #1265


teryhill:
yes $today=date('Y-m-d'); was not being loaded so the where was always wrong.

PR #1271, opened 2018-10-30
[WIP] Add the initial admin user information for when the NHANES dataset is loaded

robbyoconnor:


tmccormi:
@robbyoconnor Do you have other changes?

robbyoconnor:
Yes, I do

robbyoconnor:
Should have this done soon as the setup tool finishes -- the idea is to have a ready to go system :)

tmccormi:
ok, now we have a reason to turn on CI to build and deploy new nhanes and other demos as stuff is merged into master

robbyoconnor:
I'm going to do that -- I'll wrap it up in this PR We have a Docker Hub account if you want to use that instead of GitLab...I'd be in favor of using both

robbyoconnor:
Just the image push -- no automation for deploys (yet)

tmccormi:
I'm good with whatever you think is best

robbyoconnor:
I will ping you when I'm ready for review

robbyoconnor:
Okay, this expanded significantly: 

I set this up so that docker is now a supported development way -- now people can run EHR in Docker and the changes will be reflected in real-time. There are two compose files, one for development, one for production use. The development one has the lh-ehr code mounted a [bind mount] (https://docs.docker.com/storage/bind-mounts/). I also refactored the Dockerfile and cleaned it up a bit -- it copies the code from the repo. 

I have set up [GitLab CI as well](https://gitlab.com/librehealth/ehr/lh-ehr/pipelines) which pushes images to [GitLab's container registry](https://gitlab.com/librehealth/ehr/lh-ehr/container_registry). Additionally, there is an automated [Docker Hub Image set-up as well](https://hub.docker.com/r/librehealth/lh-ehr/)

robbyoconnor:
Going to merge this to trigger the image build on GitLab CI -- should not break anything.

aethelwulffe:
About 50 junk files.  Great.  Also untestable here.

tmccormi:
These are steps in the right direction for modern development and deployment to the cloud and, yes, refinement and iteration is required as always.

@aethelwulffe not junk files, but right in line with what is required to do a docker supported deployment for developers.    Testing in windows requires docker, which requires a Linux VM to run it.  So it may not even be relevant to windows on devs, though I know many that use docker this way.

@robbyoconnor We appreciate the work your are doing on the project, but, we would have preferred our normal review and testing process BEFORE a merge to the master and announcement of availability.  We typically require 2 integrators to review and approve the pull request before merging.

robbyoconnor:
@tmccormi sorry about that! 

@aethelwulffe they are segregated in a directory - out of the way -- this is a way to have a turnkey docker solution -- that doesn't require you go through the setup process -- which makes development faster.

robbyoconnor:
@tmccormi I have ideas on how we can eliminate the need to mount the default site -- introduce `dotenv` so that I can specify things via env vars -- then it's cleaner. This is a first pass though.

robbyoconnor:
The only place that this needs to happen is in `site/default/sqlconf.php` -- unless I'm missing something -- I'd rather not include the site directory for a configured site (it is  only 3.3M in size)

robbyoconnor:
ACTUALLY -- I have a way to fix this -- I'll make a PR

tmccormi:
The sites/*/ dirs are a not so great mix of data storage (documents, claim batches etc) as well as some code and templates that get customized and some env data (like sqlconf).   Part of this UNF project is to clean that up so it works more smoothly with modern deployment models as well as being able to deploy locally with and without Docker.   So please do open enhancement issues for us to think about and discuss as well as proposed solution P/Rs.   Thks @robbyoconnor

robbyoconnor:
Will do :-)

PR #1280, opened 2018-11-05
Remove non-working submodule that for some reason wasn't set

### LINKED ISSUE #1274: Missing git submodule is throwing the repository into a weird state
I found the culprit [here](https://github.com/LibreHealthIO/lh-ehr/tree/master/interface/tags_filters/vendor/twitter).

robbyoconnor:
Fixes #1274


robbyoconnor:
it also fails on [Docker Hub](https://hub.docker.com/r/librehealth/lh-ehr/builds/bxhnkrvugj9ulddbscunjlp/) as well.

robbyoconnor:
It will run two git commands: 

- git ls-files -zcoi --exclude-standard
- git submodule --quiet foreach 'echo $path' | tr '\\n' '\\0'

In this case, the second errors out because there's a submodule in git, but git doesn't know where the submodule is pointing...removing this should have no issue since no deployments actually have the depndency regardless.

PR #1285, opened 2018-11-08
Fix authenticated SQL injection #1215

prondubuisi:
This attempts to fix the issue mentioned above @tmccormi can you please review

### DIFF
--- a/interface/super/show_groups_popup.php
+++ b/interface/super/show_groups_popup.php
@@ -49,7 +49,7 @@ li {
 <ul>
 <?php
 $res = sqlStatement("SELECT distinct(group_name) FROM layout_options WHERE " .
-                    "form_id = '".$_GET['layout_id']."' ORDER BY group_name");
+                    "form_id = ? ORDER BY group_name", array($_GET['layout_id']));
 while ($row = sqlFetchArray($res)) {
     $gname = preg_replace("/^\d+/", "", $row['group_name']);
     $xlgname = "";


interface/super/show_groups_popup.php
```diff
@@ -49,7 +49,7 @@
 <ul>
 <?php
 $res = sqlStatement("SELECT distinct(group_name) FROM layout_options WHERE " .
-                    "form_id = '".$_GET['layout_id']."' ORDER BY group_name");
+                    "form_id = ? ORDER BY group_name", array($_GET['layout_id']));
```
robbyoconnor: You should _PROBABLY_ use a prepared statement here to properly fix this.

robbyoconnor: See [this URL](http://bobby-tables.com/php).

robbyoconnor:
See http://bobby-tables.com/php

interface/super/show_groups_popup.php
prondubuisi: @robbyoconnor thanks for pointing this out . I was thinking the sqlStatement function was flexible enough for  prepared statements  https://github.com/LibreHealthIO/lh-ehr/blob/master/library/sql.inc#L176

prondubuisi:
Another problem I have is testing  this solution, am at loss when trying to search for the form or functionality that triggers this statement.

muarachmann:
> Another problem I have is testing this solution, am at loss when trying to search for the form or functionality that triggers this statement.

you could always ask for help in our chat channels

prondubuisi:
@muarachmann When i fix this bug, I will need a way to test it has been fixed. That's what i mean when i say i need a way to test it, am currently looking for a way to access the form that posts the GET request

muarachmann:
@prondubuisi are you on linux? you can use the grep command to find use of where show_group..... is used

muarachmann:
I jsut did a grep and this is my output that shows the get request is either in these files. Check that out and verify
screen shot 2018-11-14 at 11 10 56 am

muarachmann:
I think i have answered your question(s) 😃

prondubuisi:
thanks @muarachmann for your prompt response.

muarachmann:
@prondubuisi still have an eye on this?

prondubuisi:
Yes. I just installed Linux on my System for ease. Thanks for the checkup

interface/super/show_groups_popup.php
prondubuisi: > Well, sql.inc could use some modernization...adodb and all that.

@aethelwulffe could you through moe light on this?

prondubuisi:
> @prondubuisi You have any updates for this man? (Welcome aboard btw!)

@aethelwulffe Getting done with this PR will be the best welcome I can get D:

interface/super/show_groups_popup.php
aethelwulffe: Oh man.  This stuff involves what is called the Layout Based Forms, or LBF.
It does a few different things.  One, it is the basis of creating forms and fields for demographics type stuff.  It is also used for making somewhat crude custom patient treatment forms, reading the data etc...

This designer can "group" variables it is going to display.  This is part of that tool (I think).  Look at Administration>Layout.

aethelwulffe: I think what robby is looking for here is to go 
$query = "SELECT distinct(group_name) FROM layout_options WHERE " .
                    "form_id = '".$_GET['layout_id']."' ORDER BY group_name");
                    "form_id = ? ORDER BY group_name", array($_GET['layout_id']);";
$res = sqlStatement($query);

aethelwulffe: The above should be viewed as pseudocode.  I just copied and pasted, not actually looked carefully at the syntax in the strings.  It may not work, and if you can't make something like that work, I am not worried about sql injection here so much as there are not really any non enumerated values that get passed to the query itself.

prondubuisi: Thanks @aethelwulffe

aethelwulffe: This might need something like $res = sqlFetchArray(sqlStatementNoLog($query, array($form_id->id))); or something between the two.  Sorry.  Lazy.

aethelwulffe: Depends on ability to convert to string in sql.inc.

PR #1303, opened 2018-11-27
Fix :FAB doesn't close when cursor moves away

### LINKED ISSUE #1302: FAB doesn't close when focus is lost
fabnotclosing
The Floating Action button expands when hovered on, but doesn't collapse even when another button is clicked on. 
Expected behaviour: FAB collapses when another button is clicked, when the mouse is hovered over any other element, or when the action performed by the FAB is finished.

punwai:
Fixes #1302 

Fixed it so that it no longer stays up forever once the mouse moves 50px away from it.

demo

### DIFF
--- a/interface/main/tabs/main.php
+++ b/interface/main/tabs/main.php
@@ -258,9 +258,9 @@ var webroot_url="<?php echo $web_root; ?>";
     if(checkAdmin($userAuthorized, $authUserId) && $show_icon) {
         //only admin can use the updater
         echo  "<div id='developer-mode' title='Developer Mode' style='display: none;'><i class='fa fa-code'></i></div>
-        <div id='updater-options' style='display: none;' title='updater settings'><i class='fa fa-gear'></i></div><div id='updater-icon'>
+        <div id='updater-options' style='display: none;' title='updater settings'><i class='fa fa-gear'></i></div><div id='updater-icon-wrapper'><div id='updater-icon'>
                 <i class='fa fa-refresh'></i>
-                </div>
+                </div></div>
                 <div id='updater-iframe'></div>
                 <div id='updater-settings-iframe'></div>
                 <div id='developer-mode-iframe'></div>";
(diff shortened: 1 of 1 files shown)


punwai:
@robbyoconnor @muarachmann

muarachmann:
@punwai please edit your title to something of the sort Fix :FAB doesn't close when focus is lost

interface/main/tabs/main.php
```diff
@@ -361,19 +369,44 @@ function isEncounterLocked( encounterId ) {
...
+}, function(){
+    timer = setTimeout(function(){
+        $('#developer-mode').fadeOut(100);
+        $('#updater-options').fadeOut(100);
+        $('#updater-icon').css("border", "none");
+    }, 3000);
+});
+$('#updater-icon-wrapper').hover(function(){}, function(){
+    secondTimer = setTimeout(function(){
+        $('#developer-mode').fadeOut(100);
```
muarachmann: excellent piece of jquery.... works pretty fine

interface/main/tabs/main.php
```diff
@@ -361,19 +369,44 @@ function isEncounterLocked( encounterId ) {
...
+
 $('#updater-icon').click(function () {
     $('#updater-iframe').iziModal('open');
 });
 $('body').hover(function () {
     $('#developer-mode').css("display", "none");
     $('#updater-options').css("display", "none");
     $(this).css("border", "0px solid #000");
+}, function(){
+
```
muarachmann: What this for i calls no method... nor does works

muarachmann:
@Trodrige, @naveen17797 , @robbyoconnor  tested and it works like charm... Left for @punwai to answer some questions before i can give my approval.

interface/main/tabs/main.php
punwai: @muarachmann It's a mistake in my code, I'll remove it. I was meant to put it on another hover function but probably put it here instead. It doesn't do anything though so it won't break the code

punwai: @muarachmann removed it!

interface/main/tabs/main.php
aethelwulffe: I never caught onto this happening before.  I think I should have seen something in the inspector.

aethelwulffe: Wondering if we are setting developer mode by default for superusers ("administrators").

aethelwulffe:
We have a Mua review, approval.  This tests, and giant horking icon fades away now instead of causing accidental upgrades!

PR #1304, opened 2018-11-27
Fix #1301: Calendar title no longer changes

### LINKED ISSUE #1301: Incorrect label on calendar tab after search panel closed
normalcalendar
Normal calendar tab
searchmode
Tab text changes when search button is clicked
backtocalendar
Tab text doesn't revert to its original value when search panel is closed

MigDinny:
The title of the tab is always Calendar now. There is no reason to change it either to "Back to Calendar" or to "Search" or something like that. The tab is still related to Calendar so just keep the name. Added some margins inside the content to fit better.

image

### DIFF
--- a/modules/calendar/search.php
+++ b/modules/calendar/search.php
@@ -39,11 +39,11 @@ function keywordFilter($arr) {
 </head>
 <body>
   
-  <div class="container-fluid">
-    <div class="row title">
+  <div class="container-fluid" style="margin-left: 20px; margin-right: 20px;">
+    <div class="row">
       <h2><button type="button" class="btn btn-default btn-sm" onclick="window.location.href='index.php'">Back to Calendar</button></h2>
     </div>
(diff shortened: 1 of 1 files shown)


modules/calendar/search.php
```diff
@@ -39,11 +39,11 @@ function keywordFilter($arr) {
 </head>
 <body>
   
-  <div class="container-fluid">
-    <div class="row title">
+  <div class="container-fluid" style="margin-left: 20px; margin-right: 20px;">
+    <div class="row">
```
muarachmann: Perfect weird having back to calendar as title tab lol. Didn't ever notice though :). Tested and it works great. As per the margin stuff. Looks okay. I think i will have to do something about a generalised theme for EHR so everything aligns on well not a fan of inline css that much as it might override on media queries etc
Nice one.... i am approving this.

PR #1305, opened 2018-11-28
Fix #1298: Lab Overview Title Not Displaying

### LINKED ISSUE #1298: Lab Overview title not displaying
Tab shows unknown on click of the Lab overview menu.

### Steps to reproduce this ###

Click on the following **Procedures > Lab Overview** 

### Expected behaviour ###
Lab overview tab should open with title Lab overview

### Actual behaviour ###
Shows unknown title tab bar

screen shot 2018-11-24 at 7 44 51 am

jonathanzhang53:
## Description
The issue was that the title was never declared in the respective php file. Just added the tag and it is fixed.

## Screenshot
image

### DIFF
--- a/interface/patient_file/summary/labdata.php
+++ b/interface/patient_file/summary/labdata.php
@@ -81,6 +81,7 @@ $main_spell .= "ORDER BY procedure_report.date_collected DESC ";
 // ####################################################
 echo "<html><head>";
 ?> 
+<title><?php echo xlt('Lab Overview'); ?></title>
 <link rel="stylesheet" href="<?php echo $css_header;?>" type="text/css"> 
 <link rel="stylesheet" href="<?php echo $web_root; ?>/interface/themes/labdata.css" type="text/css"> 
 <script type="text/javascript" src="<?php echo $web_root; ?>/library/js/jquery.1.3.2.js"></script>


interface/patient_file/summary/labdata.php
```diff
@@ -81,6 +81,7 @@
 // ####################################################
 echo "<html><head>";
 ?> 
+<title><?php echo xlt('Lab Overview'); ?></title>
 <link rel="stylesheet" href="<?php echo $css_header;?>" type="text/css">
```
muarachmann: Yop this does the trick. Good catch

jonathanzhang53: Thanks.

aethelwulffe: If "lab overview" does not need to be echoed onto the form (it is already in the tab) then set the title to -hidden.

aethelwulffe:
Mua approved.  AS I already tested my big merges, I guess we can do this tiny one.

Manvi07:
Hi @jonathanzhang53 
Is this issue resolved? What I'm getting is something different, the heading is "Notice". 
screenshot from 2019-02-28 16-16-57

muarachmann:
@Manvi07 yes this issue will be resolved once this PR would be merge. Try working on other issues thanks

interface/patient_file/summary/labdata.php
muarachmann: @jonathanzhang53 i think #1331 needs this

jonathanzhang53: Cool I'll check it out.

PR #1306, opened 2018-11-28
Fix #1275: Clinical Reports Enhancement (Dynamically Sized Inputs)

### LINKED ISSUE #1275: Clinical Reports enhancement
The two new clinic reports need a useability enhancement.  The column head search filter boxes are TINY compared to the width of the columns.

Please make the search box dynamically sized to match or nearly match the width of the data below.

This will help make it obvious to users the these are meant to be used not graphic blips on the screen.

dx-nhanes

jonathanzhang53:
## Description
The table columns adjust their height to the biggest length value in the column. However, the search bars do not follow suit. I added a css styling to make the width of the inputs 100%. The inputs will adjust to the size of the lengthiest data value in the column when data is populated (tested with inspect element in nhanes demo).

## Screenshots
image

### DIFF
--- a/assets/js/DataTables-1.10.16/datatables.css
+++ b/assets/js/DataTables-1.10.16/datatables.css
@@ -822,6 +822,9 @@ a.dt-button.processing:after {
 }
 
 
+table thead tr th input {
+  width: 100%;
+}
 table.dataTable.dtr-inline.collapsed > tbody > tr > td.child,
 table.dataTable.dtr-inline.collapsed > tbody > tr > th.child,
 table.dataTable.dtr-inline.collapsed > tbody > tr > td.dataTables_empty {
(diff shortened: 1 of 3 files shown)


assets/js/DataTables-1.10.16/datatables.css
```diff
@@ -822,6 +822,9 @@ a.dt-button.processing:after {
 }
 
 
+table thead tr th input {
```
aethelwulffe: Comment that this is datatables customization, or otherwise rename the file datatables_custom.css and use it that way somehow.  I think this is the proper way to go about modifying a third party asset.  Anyone else have an opinion on this?

interface/reports/lab_stats_by_demographics_report.php
```diff
@@ -236,13 +236,13 @@ function lab_result_details() {
     <thead>
 
     <tr>
-        <th><input  id = 'column0_search_show_lab_details_table' size="4"></th>
```
aethelwulffe: https://github.com/LibreHealthIO/lh-ehr/pull/1284/files should not affect this, right?  It is another PR coming up.  Probably needs to be merged first, as that is for the race condition.

PR #1307, opened 2018-11-30
Add editorconfig file

robbyoconnor:
This adds an [editorconfig file](https://editorconfig.org/) which most editors/IDEs are designed to honor to allow for consistency.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true


muarachmann:
@robbyoconnor not sure of what this does... I use phpstorm and i have indentations ok  etc.. Not seeing the difference when i test this. Could it be due to the fact my editor works fine or .....

robbyoconnor:
@muarachmann -- it's useful to have here -- It's very much an out of the way type thing -- check out https://
editorconfig.org

aethelwulffe:
Mine doesn't.  Well, one does.  I use three :(
Issues:  
-This commit would be incorrect as we like crlf for git and cross-platform.
-Need to have a phone meeting to see if the folks care, and if things like the tab treatment should be included.  
-This might not the right flavor of thing to have in the repo, as it is an editor specific project file and in a dev package and not something that floats on everyone's web server.
- The git thing to do is git config core.autocrlf, and that should be set already for everyone in their git-config.  Thing is, we can all take care of formatting through git-config to handle tabs and all of that as well....which we stupidly never thought to do.
Let's discuss implementing *something like this* to help format our output and avoid whitespace issues.  Don't close this too fast yet.

robbyoconnor:
Fully within your right -- I just offered the change -- you were under
no obligation to take it :)

PR #1315, opened 2018-12-02
Fix #1314: Style.php no longer overrides PRISM colors

### LINKED ISSUE #1314: Style.php overrides some theme configs
Last year I created the PRISM theme for a GCI Task. Some colors were overriden by style.php system. The idea is nice but its override on other theme's settings can break user experience.

image
image

_Patient Data and Tabs_ are white due to Primary Color setting.

This major problem must be fixed to improve user experience.

_Updated below_

image

MigDinny:
Fixes #1314 

_Updated below_

image

### DIFF
--- a/library/headers.inc.php
+++ b/library/headers.inc.php
@@ -19,8 +19,9 @@
  */
 
 ?>
-<link rel="stylesheet" href="<?php echo $css_header;?>" type="text/css">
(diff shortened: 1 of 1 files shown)


muarachmann:
@MigDinny does this only affects prism theme?

MigDinny:
@muarachmann apparently it only affects prism i think. but anyway, i think style.php is breaking all themes' configurations. if a theme has a color, style.php is overriding its colors with primaryColor setting for example.

I don't think it's beneficial for the program existence of style.php. Each theme should have its own configurations and that's that.

Trodrige:
We need a solution to this multiple theme thing. Fixing one theme affects another, this happens often.

tmccormi:
Since "Themes" are about color choices (mostly) let get ONE theme working, delete all the others and make new themes with different color choices from that working one.

Later we can just have one theme and a user color selection tool (or something).

aethelwulffe:
No changes in this PR

PR #1318, opened 2018-12-04
Fix: Admin Title no longer obstructs nav 

### LINKED ISSUE #1313: Navbar is overlapped by Title
How to recreate: 
Minimize the browser until the Title is overlapping with the navbar elements. This will Overlap the navbar elements. This breaks the hovering of the navbar.
fix1

punwai:
fix3

Fixes #1313


aethelwulffe:
OK.  Came back to re-test and merge with my brain on.
To wit, I have not been able to figure out how to get my OLD version to do any eclipsing of the username (the "admin" you are talking about I suppose).  Might be that "popups" are gone in my config here.

aethelwulffe:
Finally realized that it is about the hover toggling off repeatedly like it tends to do when you need to log out.

Issue #1321, opened 2018-12-04
Administration > Rules Unknown Page


punwai:
<img width="556" alt="screen shot 2018-12-04 at 03 08 33" src="https://user-images.githubusercontent.com/28529491/49416482-fe067000-f771-11e8-9ced-224f4c2b6211.png">
A page in the code is broken. After checking the URL the iframe redirects to (/lh-ehr/interface/super/rules/index.php?action=browse!list), it seems like the URL is correct but the php is incorrect.

Trodrige:
Great job @punwai 

Head to Popups -> chart label. The same thing happens, try to apply your solution here.

See address label and barcode label too.

punwai:
@Trodrige It seems to be a different issue here, the links are working perfectly fine but the title is not displaying because a PDF is returned.

punwai:
I will create an issue on this and try to fix it.

PR #1325, opened 2018-12-04
Fix: Broken links #1299

punwai:
@muarachmann Added this new PR instead. The links in the PHP files were broken #1299.

PR #1337, opened 2018-12-05
Fixing Unknown title for PDFs

### LINKED ISSUE #1336: PDFs pages Unknown titles.
screen shot 2018-12-05 at 23 18 51
On some PDF pages for example, Popups > Chart Label, Popups > Barcode Label and Popups > Address Label, the title is unknown

punwai:
PDFs do not display titles as the pages are not stored with titles in. To add labels to PDF files, a structure similar to the committed changes must be employed. I was tempted to change this for ALL PDF files but was not sure how necessary this would be as they may not all be iframes. I am awaiting for @Trodrige 's view on this. 

Fixes #1336

### DIFF
--- a/interface/patient_file/label.php
+++ b/interface/patient_file/label.php
@@ -75,5 +75,11 @@ for($i=1;$i<=$last;$i++) {
 	$pdf->Add_Label($text);
 }
 
-$pdf->Output();
+$output = $pdf->Output('','S');
+$output = base64_encode($output);
 ?>
+
+<title>Chart Label</title>
+<body style="margin: 0">
+    <embed src="data:application/pdf;base64,<?php echo $output ?>" type='application/pdf' style="position: fixed; width: 100%; height: 100%;">
+</body>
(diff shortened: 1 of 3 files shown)


interface/patient_file/label.php
```diff
@@ -75,5 +75,11 @@
 	$pdf->Add_Label($text);
 }
 
-$pdf->Output();
+$output = $pdf->Output('','S');
+$output = base64_encode($output);
 ?>
+
+<title>Chart Label</title>
```
muarachmann: nice catch this does the trick

Trodrige:
@punwai is this supposed to solve the Admind-> Rules problem too? Or should we expect another PR. I still see this
screen shot 2018-12-11 at 8 54 43 am

Trodrige:
This doesn't solve #1321

punwai:
@Trodrige My bad, I was referring to the issue you mentioned in #1321 rather than the main problem. #1325 fixes #1321 :)

interface/patient_file/label.php
aethelwulffe: Need to check for issues with base64_encode php dependency.

PR #1342, opened 2018-12-07
Fixes #1341 - Turned Button Texts to White

### LINKED ISSUE #1341: Save and Cancel Button texts on New Encounter Form are hardly visible
## Problem
In **New Encounter Form**  the Save & Cancel buttons are hardly visible. (in every theme)

## Screenshots : 
screenshot from 2018-12-07 12-32-20

bauripalash:
Previously the buttons on **New Encounter Form** were hardly visible because of their weird color combinations. I fixed them to white in Light and Prism theme.

Fixes #1341 

screenshot from 2018-12-07 13-05-59

### DIFF
--- a/interface/themes/style_light.css
+++ b/interface/themes/style_light.css
@@ -693,7 +693,7 @@ tr.odd, td.even {
   padding-right: 10px;
   /* sliding doors padding */
   text-decoration: none;
-  color: #101e34 !important;
+  color: white !important;
   background-color: #5ac3ca;
   padding: 5px 12px 5px;
   border: none;
--- a/interface/themes/style_prism.css
+++ b/interface/themes/style_prism.css
@@ -675,7 +675,7 @@ tr.odd {
 }
 .css_button, button {
   background: #063f80;
-  color: #101e34  !important;
+  color: white  !important;
   display: inline-block;
   font-weight: 300;
   margin: 4px;


aethelwulffe:
We should dump the "light" theme.  We really only want a single theme 
that allows for selection of colors, button and layout styles.  Not the 
old css template approach.  I would rather that the CORE application 
only have one theme (plus R-T-L) in any respect, even if we do support 
mod themes.

aethelwulffe:
I use it.

Leaving it with the default settings works great IMO.

It only works with prism (the whole point), but I still think we can 
pare down the list and move forward.  Just my opinion.  Looks like this 
initiative is moving towards approaching code unity between the CSS 
anyway...that is good!

tmccormi:
I agree, I like the proposal to use the css variables to pare things down, there is also some CSS (menu.css) that is not part of any theme, which I don't think is great.

aethelwulffe:
No, that is not optimal..no.

I would like the view model to have a consolidated one-stop shop, 
preferably with the code containerized in the directory structure.  Too 
many bits scattered, too deep in separated directories for something 
that as a core feature is usually a very coherent pile of code in most 
applications.

Trodrige:
@bauripalash I can't observe the changes you made, the colors are the same. Give some spacing between the form fields.

bauripalash:
@Trodrige I think this PR will be pointless because these themes gonna get changed, here's the discussion : https://github.com/LibreHealthIO/lh-ehr/issues/1357

aethelwulffe:
@bauripalash Does this allow me (if I test it) to change the background as you did from blue to whatever in prism on the basis that it is white by default?  If so, then it is not useless and it can be reviewed again and merged.

aethelwulffe:
text color needs to basically be fully dynamic.

PR #1345, opened 2018-12-07
Medical Issue End-date display

aethelwulffe:
In stats_full.php End date for an issue is displayed as a time when the value is NULL.  Looks goofy with fake negative 1969 date in there...this fixes it, especially when you add additional custom patient ISSUE types.


robbyoconnor:
Not sure if I labeled correctly -- been triaging GCI stuff.

PR #1348, opened 2018-12-08
fix: button change to under form

punwai:
**Link:** Fees->Payment
**Work:** - Put buttons immediately under the form
**Changes** - The buttons are now under the form rather than under the table
**After Changes**
screen shot 2018-12-08 at 04 41 17


aethelwulffe:
We have two approvals.  That is a go.  I wonder if under the table is intended to force review of a long list.  Still two ups it is.  If someone hates it, we can get another PR, or actually get to rebuilding the billing module.

PR #1349, opened 2018-12-08
Fix: Select Measure Style

punwai:
**Link**: QA Measures->Select Measures

**Work**: Let “Update” and MIPS specialty… be on one row
	- Change the background of all measures
	- Delete one update button

**Changes**: 
- Added bootstrap rows to structure the form. The checkboxes were broken to start with.
- Deleted one Update button.
- Changed colour of all measures

**Before**: 
screen shot 2018-12-08 at 07 03 21

**After**:
screen shot 2018-12-08 at 23 36 46

### DIFF
--- a/modules/MIPS/measure_activate.php
+++ b/modules/MIPS/measure_activate.php
@@ -25,6 +25,10 @@ $fake_register_globals = false;
 
 require_once '../../interface/globals.php';
 require_once $srcdir.'/api.inc';
+require_once("$srcdir/headers.inc.php");
+//Include Bootstrap
+call_required_libraries(array("bootstrap"));
+
 
 $updateStatus = array(
 	'off' => array(),
@@ -78,6 +82,7 @@ h1, h2, h3 {
 }
 #measures ul {
 	padding-left: 1em;
+	display: inline;
 }
 #measures li {
 	list-style: none;
@@ -119,7 +124,6 @@ h1, h2, h3 {
 	display: none;
 }
 .quickselect {
-	
 	border: 1px solid #333333;
 	border-radius: 4px;
 	color: #333333;
@@ -128,6 +132,13 @@ h1, h2, h3 {
 	margin: 0.25ex 0.15ex 0.15ex 0ex;
 	padding: 0 0.75ex;
 	font-size: 9pt;
+}
+.ma-title {
+	font-size: 20px; 
+	margin-left: 15px;
+}
+.row{
+	margin: 0;
 }
 		</style>
 		<script type="text/javascript" src="../../library/js/jquery-1.9.1.min.js"></script>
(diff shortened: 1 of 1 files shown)


muarachmann:
@punwai a preview gives us some clue on which section you modified. Mind including one please

modules/MIPS/measure_activate.php
```diff
@@ -25,6 +25,10 @@
 
 require_once '../../interface/globals.php';
 require_once $srcdir.'/api.inc';
+require_once("$srcdir/headers.inc.php");
+//Include Bootstrap
+call_required_libraries(array("bootstrap"));
+
```
muarachmann: Great this shows you are getting to know the codebase very well, I like this

modules/MIPS/measure_activate.php
```diff
@@ -123,11 +127,19 @@
...
 	font-family: monospace;
 	margin: 0.25ex 0.15ex 0.15ex 0ex;
 	padding: 0 0.75ex;
 	font-size: 9pt;
+}
+.ma-title {
+	font-size: 20px; 
+	margin-left: 15px;
+}
+.row{
```
muarachmann: Bueeeehhh whats this for??? I think bootstrap has a class called `row` which does pretty what you want to accomplish?? why redefining that class ? @punwai

punwai: @muarachmann This is because rows come with a custom margin-left. I want the bootstrap rows to not have any margins but have the same grid functionality so I have to do this.

modules/MIPS/measure_activate.php
```diff
@@ -309,53 +320,49 @@ function quickSelect(specialty) {
...
-						</ul>
-					</li>
-					
-					<li>
+			<div id="measures">				
+				<div class="row"><p><span class="align-middle"><input class="btn btn-success btn-sm align-middle" type="submit" value="Update" /></span><span class="align-middle ma-title">MIPS Specialty Sets for 2018</span></p></div>
+				<div class="row">
+					<button type="button" class="quickselect" onclick="quickSelect('MIPS-1 Allergy/Immunology')">MIPS-1 Allergy/Immunology</button>
+					<button type="button" class="quickselect" onclick="quickSelect('MIPS-2 Anesthesiology')">MIPS-2 Anesthesiology</button>
+					<button type="button" class="quickselect" onclick="quickSelect('MIPS-3 Cardiology')">MIPS-3 Cardiology</button>
```
muarachmann: Looks pretty good. Great work 

screen shot 2018-12-09 at 8 44 37 am

aethelwulffe:
I can approve this, but this issue only happens with the "Light" theme (and the MIPS module is not being maintained in this repo due to slow commit process for a fast moving thing).
Looks like this normally:
image
The above needs work, but it scales properly with the PRISIM theme, and we can make it pretty if we do a 2019 measure set for the main branch.

PR #1351, opened 2018-12-08
Edit layout_edit styling.

punwai:
Fixed
Link: Admin->Layouts->’Any’
Work: -Table headers(General, Family History, ...) Let these be above the row of buttons

Also Extra problems I've found.
screen shot 2018-12-08 at 06 10 43


punwai:
@muarachmann I will try testing with prism theme, it might be that the border was overridden by another property.

PR #1360, opened 2018-12-09
Fix: Table styling and inline header

punwai:
Link: Admin->Layouts->’Any’

Work: -Table headers(General, Family History, ...) Let these be above the row of buttons

Changes: 
- Table header no longer is inline with buttons but over the buttons.
- Made fields bigger
- Made all tables fill width space
- Removed some table inline styling
- Added borders to all popup boxes

screen shot 2018-12-09 at 23 17 30

screen shot 2018-12-09 at 23 17 43

screen shot 2018-12-09 at 23 17 21


muarachmann:
@punwai let me look at this. This seems to be good. 🙃

PR #1363, opened 2018-12-11
style: patient > add new/search

punwai:
## Task 1 (Grab Bag)
Link: Patient->Search or Add Patient->Insurance
Work: - Change the font-color from red to black(default)
	-Give insurance form fields styling, space out the elements. For example give some space between select fields and the Add buttons

#### - Restyled form and buttons
screen shot 2018-12-11 at 12 45 42

#### - Removed red text from Insurance section
screen shot 2018-12-11 at 12 46 00

#### - Replaced normal buttons with bootstrap buttons
screen shot 2018-12-11 at 12 46 03

#### - Changed checkboxes to bootstrap styled checkboxes
screen shot 2018-12-11 at 12 46 08

### DIFF
--- a/interface/new/new_comprehensive.php
+++ b/interface/new/new_comprehensive.php
@@ -68,7 +68,7 @@
   <head>
     <?php html_header_show();
       //  Include Bootstrap, Fancybox, date-time-picker
-  call_required_libraries(array("jquery-min-3-1-1","bootstrap","datepicker","fancybox-addpatient"));
+  call_required_libraries(array("bootstrap","jquery-min-3-1-1","datepicker","fancybox-addpatient", "font-awesome", "iziModalToast"));
 ?>
 
 <style>
@@ -89,6 +89,40 @@ div.section {
  */
  margin: 0 0 0 10pt;
  padding: 5pt;
+}
+
+input[type="text"]{
+  height: 34px;
+  padding: 6px 12px;
+  font-size: 14px;
+  line-height: 1.42857143;
+  color: #555;
+  background-color: #fff;
+  background-image: none;
+  border: 1px solid #ccc;
+  border-radius: 4px;
+  -webkit-box-shadow: inset 0 1px 1px rgba(0,0,0,.075);
+  box-shadow: inset 0 1px 1px rgba(0,0,0,.075);
+  -webkit-transition: border-color ease-in-out .15s,-webkit-box-shadow ease-in-out .15s;
+  -o-transition: border-color ease-in-out .15s,box-shadow ease-in-out .15s;
+  transition: border-color ease-in-out .15s,box-shadow ease-in-out .15s;
+}
+
+input[type="entry"]{
+  height: 34px;
+  padding: 6px 12px;
+  font-size: 14px;
+  line-height: 1.42857143;
+  color: #555;
+  background-color: #fff;
+  background-image: none;
+  border: 1px solid #ccc;
+  border-radius: 4px;
+  -webkit-box-shadow: inset 0 1px 1px rgba(0,0,0,.075);
+  box-shadow: inset 0 1px 1px rgba(0,0,0,.075);
+  -webkit-transition: border-color ease-in-out .15s,-webkit-box-shadow ease-in-out .15s;
+  -o-transition: border-color ease-in-out .15s,box-shadow ease-in-out .15s;
+  transition: border-color ease-in-out .15s,box-shadow ease-in-out .15s;
 }
 
 </style>
@@ -188,14 +222,14 @@ function ins_search(ins) {
 }
 
 function checkNum () {
- var re= new RegExp();
- re = /^\d*\.?\d*$/;
- str=document.forms[0].monthly_income.value;
- if(re.exec(str))
- {
- }else{
-  alert("Please enter a dollar amount using only numbers and a decimal point.");
- }
+  var re= new RegExp();
+  re = /^\d*\.?\d*$/;
+  str=document.forms[0].monthly_income.value;
+  if(re.exec(str))
+  {
+  }else{
+    alert("Please enter a dollar amount using only numbers and a decimal point.");
+  }
 }
 
 // This capitalizes the first letter of each word in the passed input
@@ -236,14 +270,14 @@ function allowOnlyDigits(elem_name){
       }
 
       function checkNum () {
-       var re= new RegExp();
-       re = /^\d*\.?\d*$/;
-       str=document.forms[0].monthly_income.value;
-       if(re.exec(str))
-       {
-       }else{
-        alert("Please enter a dollar amount using only numbers and a decimal point.");
-       }
+        var re= new RegExp();
+        re = /^\d*\.?\d*$/;
+        str=document.forms[0].monthly_income.value;
+        if(re.exec(str))
+        {
+        }else{
+          alert("Please enter a dollar amount using only numbers and a decimal point.")
+        }
       }
 
       // This capitalizes the first letter of each word in the passed input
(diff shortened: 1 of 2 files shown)


interface/new/new_comprehensive.php
```diff
@@ -395,46 +429,46 @@ function validate(f) {
         msg += "\n<?php echo htmlspecialchars(xl('Please fill them in before continuing.'),ENT_QUOTES); ?>";
 
         if ( errMsgs.length > 0 || ins_error == '1' ) {
-               alert(msg);
-               return false;
+          alert(msg);
```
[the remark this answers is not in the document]
punwai: @muarachmann Also, what is the purpose of the yellow boxes here, it seems pretty pointless to me.

aethelwulffe:
The RED indicates REQUIRED fields.
Also, there is a reason for the red text in the insurance data.  The red stuff is the stuff that cannot be gotten wrong on data entry, and cannot be user config altered or very very bad stuff happens.

aethelwulffe:
The red borders are a nice way of doing it, but take it easy on the white space.  The users must be able to see a lot of data all at once.
There may be a PR out there right now from @MigDinny that implemented "accordion" for this stuff.

interface/new/new_comprehensive.php
```diff
@@ -236,14 +270,14 @@ function ins_search(ins) {
...
-       }else{
-        alert("Please enter a dollar amount using only numbers and a decimal point.");
-       }
+        var re= new RegExp();
+        re = /^\d*\.?\d*$/;
+        str=document.forms[0].monthly_income.value;
+        if(re.exec(str))
+        {
+        }else{
+          alert("Please enter a dollar amount using only numbers and a decimal point.")
```
aethelwulffe: "Monetary", not dollar.  Dollars are semi-country specific unless you wrap these in xlt() with an ADDITIONAL xlt that replaces local currency name as well as putting it in in linguistic syntax order.  Big challenge there...so monetary or currency vs. dollar.

interface/new/new_comprehensive.php
```diff
@@ -380,61 +414,73 @@ function validate(f) {
...
         }
         <?php } ?>
-        var msg = "";
         var ins_error = '0';
-        msg += "<?php echo htmlspecialchars(xl('The following fields are required'),ENT_QUOTES); ?>:\n\n";
         for ( var i = 0; i < errMsgs.length; i++ ) {
-               msg += errMsgs[i] + "\n";
+          iziToast.error({
+            title: 'Error',
+            message: "<?php echo htmlspecialchars(xl('The following field is required: '),ENT_QUOTES); ?>" + errMsgs[i],
```
aethelwulffe: xl or xlt ?

interface/new/new_comprehensive.php
```diff
@@ -380,61 +414,73 @@ function validate(f) {
...
+            position: 'topRight',
+          })
         }
         <?php if($GLOBALS['primary_insurance_required']){ ?>
           if($('#i1provider option:selected').val() == '') {
-             var ins_error = '1';
-             msg += "<?php echo htmlspecialchars(xl('Primary Insurance Provider Entry Required.'),ENT_QUOTES); ?> \n";
+            var ins_error = '1';
```
aethelwulffe: The "required" list...are all of these configured somewhere?  The Layout engine has a tool for setting layout elements as required or not (only works on some, probably means they are handled separately).  Being able to control "required" fields for creating new patients and the like (and for duplicate checking) is darn important...and that is a feature we don't have so good.  State/region is one of those items that winds up creating a lot of screwups if it is not entered.
-So, when I test this, Do I just need to create a patient and hit various fields, right?  What if I change some demographics fields in the layout options?

interface/new/new_comprehensive.php
muarachmann: @aethelwulffe i have also been wondering on this from time to time i know the `xl` , `xlt` are used for translation. What is the difference between the two

aethelwulffe: xl() is defined in library/translations.inc.php, and has a number of functions that we don't actually use, like special label versions etc...
These functions are supposedly not compatible with HTMLSPECIALCHAR.
```
C:\AlphaFiles\bidenis\REPOS\BRANCH_TESTS\interface\language\lang.info.html:
   62  IV. Developer Instructions
   63  
   64: The translation is done by the xl() custom function, which can be found
   65  at libreehr/library/translation.inc.php.
   66  Its semantics are very simple.
   ..
   85  http://librehealth.io/wiki/index.php/Development_Policies#Internationalization
   86  
   87: There are also wrappers to the xl() function that are used infrequently in specific situations. Please
   88  refer to libreehr/library/translation.inc.php for more details.
   89  

C:\AlphaFiles\bidenis\REPOS\BRANCH_TESTS\interface\orders\orders_results.php:
   87      list($order_id, $order_seq, $report_id, $result_id) = explode(':', $line_value);
   88  
   89:     // Not using xl() here because this is for debugging only.
   90      if (empty($order_id)) die("Order ID is missing from line $lino.");
   91  

C:\AlphaFiles\bidenis\REPOS\BRANCH_TESTS\library\htmlspecialchars.inc.php:
   93  	} else {
   94  		trigger_error(
   95: 			'Translation via xl() was requested, but the xl()'
   96  			. ' function is not defined, yet.',
   97  			E_USER_WARNING
   ..
  102  
  103  /**
  104:  * Translate via xl() and then escape via text().
  105   *
  106   * @param string $key The string to escape, possibly including "&", "<",
  ...
  113  
  114  /**
  115:  * Translate via xl() and then escape via attr().
  116   *
  117   * @param string $key The string to escape, possibly including (&), (<),
  ...
  124  
  125  /*
  126: Translate via xl() and then escape via addslashes for use with javascript literals
  127  */
  128  function xls($key){

C:\AlphaFiles\bidenis\REPOS\BRANCH_TESTS\library\translation.inc.php:
   53  }
   54  
   55: // ----------- xl() function wrappers ------------------------------
   56  //
   57: // Use above xl() function the majority of time for translations. The
   58  //  below wrappers are only for specific situations in order to support
   59  //  granular control of translations in certain parts of LibreEHR.

C:\AlphaFiles\bidenis\REPOS\BRANCH_TESTS\library\authentication\privDB.php:
   81      if ($recordset === FALSE) {
   82          
   83:       // These error messages are explictly NOT run through xl() because we still
   84        // need them if there is a database problem.
   85        echo "Failure during database access! Check server error log.";

C:\AlphaFiles\bidenis\REPOS\BRANCH_TESTS\library\Smarty\plugins\function.xl.php:
    4   * @package Smarty
    5   * @subpackage plugins
    6:  * simplified xl() version for smarty templates
    7   * Christian Navalici 2007
    8   */

C:\AlphaFiles\bidenis\REPOS\BRANCH_TESTS\modules\language_translations\collectConstants.pl:
  208     }
  209  
  210:    # skip $. Raally rare usage of xl() function.
  211     # There are about 25 lines of this in entire codebase
  212     # and likely just several contants. Can put in manually
  ...
  230     # Skip line with the js_xl function
  231     if ($de eq "m") {
  232:     print LOGFILE "MESSAGE:  Special case character 'm' skipped. Done to skip the function js_xl() line.\n";
  233      print LOGFILE $editvar2."\n";
  234      next;
```
You can find the xlt() in htmlspecialchars.inc.php here:
```
C:\AlphaFiles\bidenis\REPOS\MIPS_Gateway\library\htmlspecialchars.inc.php:
  108   * @return string The string, with "&", "<", and ">" escaped.
  109   */
  110: function xlt($key) {
  111  	return text(hsc_private_xl_or_warn($key));
  112  }
```

muarachmann: so which are we to use henceforth?

muarachmann:
@punwai can you resolve the conflicts on this

aethelwulffe:
@punwai @muarachmann 
Your conflicts are a library call function that has a different order to the arguments (and a couple of spaces), some izitoast error methods that have a little different syntax using semicolons, and a final error label with "title=error" thingie.  You can resolve them with the in-line editor I think.

interface/new/new_comprehensive.php
aethelwulffe: The functions in htmlspecialchars.inc.php (xlt, xla etc...) include special character functions sorta.  Use what you need, just know what those are.
xlt() includes a function also defined in htmlspecialchars.inc.php that handles that stuff, and is not supposed to be called directly (hidden).

aethelwulffe:
Bump....

muarachmann:
Getting stale ... closing could be re-opened anytime if necessary

PR #1365, opened 2018-12-12
style: form administration tables

punwai:
The forms were quite hard to look at to start with, I added some bootstrap styling to the table and the surroundings to make it more organized.

Task 20
Link: Admin->Other->Form
Work: Style the whole page, horrible mix of colors here, centralize fields(Already open issue)

### Before:
screen shot 2018-12-11 at 21 33 52
### After:
screen shot 2018-12-12 at 00 22 55

### DIFF
--- a/interface/forms_admin/forms_admin.php
+++ b/interface/forms_admin/forms_admin.php
@@ -34,9 +34,9 @@ $bigdata = getRegistered("%") or $bigdata = false;
 <head>
 <?php html_header_show();?>
 <link rel="stylesheet" href="<?php echo $css_header;?>" type="text/css">
+<?php call_required_libraries(array('bootstrap')) ?>
 </head>
 <body class="body_top">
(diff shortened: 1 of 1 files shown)


interface/forms_admin/forms_admin.php
```diff
@@ -61,150 +61,167 @@
...
+        else
+            $phpState =  "PHP extracted";
+        ?>
+        <tr >
+            <td width="20%" class="buttoncol">
+                <?php
+                    $form_title_file = @file($GLOBALS['srcdir']."/../interface/forms/$fname/info.txt");
+                            if ($form_title_file)
+                                    $form_title = $form_title_file[0];
+                            else
```
muarachmann: okay looks good

aethelwulffe: Looks *sooo much better*.  I think it is a lot simpler to use.  Links to enable an install db work and everything.

PR #1369, opened 2018-12-13
Improve install notes and test the process

tmccormi:


docker/README.md
```diff
@@ -98,14 +123,18 @@ Login with credentials provided above:
 * Initial User: `admin`
 * Initial User Password: `password`
 
-## Cleanup
+## Cleanup (Shutdown?)
```
robbyoconnor: Why is there a question mark here

robbyoconnor: Fixing this so no need to act.

robbyoconnor:
Merged and a follow-up fix in fd7a7493068a4918f52a902b9c70242bd829ca2b

robbyoconnor:
I revamped most of it in #1370

PR #1382, opened 2019-01-08
Fix #1380: SQL Injection vulnerability

### LINKED ISSUE #1380: SQL Injection vulnerability
Hi,
I believe to have found an SQL Injection vulnerability that exists once the datatable for Physical Exam Diagnoses has been installed. Because I don't know how you prefer to handle security issues like these, I have decided not to post any more information yet. I would appreciate it if you could get back to me whether I should describe the details here or somewhere else.

Regards

DustinBorn:
Use a Prepared Statement to prevent SQL Injection.

Fixes #1380

### DIFF
--- a/interface/forms/physical_exam/edit_diagnoses.php
+++ b/interface/forms/physical_exam/edit_diagnoses.php
@@ -61,7 +61,7 @@
 
  $dres = sqlStatement(
   "SELECT * FROM form_physical_exam_diagnoses WHERE " .
-  "line_id = '$line_id' ORDER BY ordering, diagnosis"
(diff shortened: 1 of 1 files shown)


interface/forms/physical_exam/edit_diagnoses.php
```diff
@@ -61,7 +61,7 @@
 
  $dres = sqlStatement(
   "SELECT * FROM form_physical_exam_diagnoses WHERE " .
-  "line_id = '$line_id' ORDER BY ordering, diagnosis"
+  "line_id = ? ORDER BY ordering, diagnosis", array($line_id)
```
muarachmann: I think this is okay @DustinBorn.

PR #1388, opened 2019-01-20
Update gitignore file

muarachmann:
This adds `.idea` , all  `sites/*/sqlconf.php` to the git ignore. Other files could be added as well

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 *.env.*
 .htaccess
+.idea/
+sites/*/sqlconf.php
+sites/*/profile_pictures/*
+


.gitignore
```diff
@@ -1,2 +1,6 @@
 *.env.*
 .htaccess
+.idea/
+sites/*/sqlconf.php
+sites/*/profile_pictures/*
```
robbyoconnor: Move line 5 into your other pull request.

.gitignore
```diff
@@ -1,2 +1,6 @@
 *.env.*
 .htaccess
+.idea/
+sites/*/sqlconf.php
```
robbyoconnor: @tmccormi, @aethelwulffe -- should this be ignored?

.gitignore
muarachmann: ok

.gitignore
muarachmann: IMO the .idea yes. For `sqlconf.php` i thought it wise and safer cause i guess we had an issue like this where someones sqlconf was merged into the codebase and cause some disruptions.  If needed to change in the future it can be done by just removing  from the gitignore  of course i think one needs to know ahead that it was ignored - thats the only bad to newbies

aethelwulffe:
Review this in ref to the profile pic fix merge.

muarachmann:
@aethelwulffe is the idea of adding sqlconfig to the git ignore suitable? So i adjust appropriately

aethelwulffe:
Each site directory is something that should be ignored...and yes, this is just fine as a stop-gap for folks that are testing from their actual repo...which they should not be doing anyway...and they should not be using the /default site...and the setup should suggest "/mysite" and not "default".  
So:

- -/default site should never be used on install in any case.  We wanted that as a real case instance for the "installer".  Technically, the directories under /sites should be empty
- -We don't want an overall gitignore in the /sites when actually changing the sites/default directory.
- -When you are feeding an active install on a production server, you should only be using git pull.  That means that if you have additional active sites, they will be unaffected.  That is good.
- -Thus, there isn't any reason to add a gitignore to the base project.  If you want this for your own dev instance, it is fine, but NOT NECESSARILY GOOD for production environments.

Think about that, and tell me what all extra we should do.

aethelwulffe:
Changes requested here are for another PR.  There were two other related PR's, one of which merged and created a minor conflict. Merging this (in lieu of a second review) to ensure conflicts are minimized.  This is plain and simple a totally acceptable gitignore file.

PR #1391, opened 2019-01-21
deleted useless file

muarachmann:
When i attempt to delete this PHPstorm shows me its usage in `new_patien.php` and `new_patient_save.php`. But i dont think it is actually ever used.

### DIFF
--- a/interface/new/new.php
+++ b/interface/new/new.php
@@ -1,3 +1,0 @@
-<?php
-//Useless
-?>


muarachmann:
@aethelwulffe i think this is a special case cause it was renamed to `new_comprehensive.php` so i cant know other instances. If you can point them i would refactor and  re-commit

aethelwulffe:
This has been sitting here a while.  Merging.

PR #1394, opened 2019-01-22
translate text corrections to english

aethelwulffe:
Round 1.  Fixing the ENGLISH before adding the Spanish translations.  To be checked against whatever list someone else has come up with.
-Note, one item involves removing an ICD-9 code from Bronchitis form.  The other translatable items in there need to have entries made for ICD-10, probably push to the fee sheet, or simply delete the old funky form.


aethelwulffe:
@teryhill You think I found all of them?

teryhill:
If you did not we will see them eventually cause they will not translate.

aethelwulffe:
Someone else check this so we can merge it and get the translation folks on track.

VirginiaBalseiro:
Any idea when this will be merged?
I need those Greek tags in the Spanish version gone ASAP for the docs I'm working on.
Thanks.

robbyoconnor:
Perfect -- I will update the demo right now!

aethelwulffe:
This is only the English text changes.  Working on PR for Spanish.

robbyoconnor:
Still pulled in -- let me know when it's done

PR #1398, opened 2019-01-26
fix authenticated sql injection #1215

prondubuisi:
@muarachmann @robbyoconnor , I attempted a re-fix for this issue, please review

### DIFF
--- a/interface/super/show_groups_popup.php
+++ b/interface/super/show_groups_popup.php
@@ -48,8 +48,10 @@ li {
 <h1><?php xl('Layout groups','e'); ?></h1>
 <ul>
 <?php
-$res = sqlStatement("SELECT distinct(group_name) FROM layout_options WHERE " .
-                    "form_id = '".$_GET['layout_id']."' ORDER BY group_name");
+
+$query = "SELECT distinct(group_name) FROM layout_options WHERE form_id = ? ORDER BY group_name";
+$res = sqlStatement($query, array($_GET['layout_id']));
+
 while ($row = sqlFetchArray($res)) {
     $gname = preg_replace("/^\d+/", "", $row['group_name']);
     $xlgname = "";


interface/super/show_groups_popup.php
```diff
@@ -48,8 +48,11 @@
 <h1><?php xl('Layout groups','e'); ?></h1>
 <ul>
 <?php
-$res = sqlStatement("SELECT distinct(group_name) FROM layout_options WHERE " .
-                    "form_id = '".$_GET['layout_id']."' ORDER BY group_name");
+
+$query .= "SELECT distinct(group_name) FROM layout_options WHERE ";
+$query .= "form_id = ? ORDER BY group_name";
+$res = sqlStatement($query, array($_GET['layout_id']));
+
```
muarachmann: LGTM i think this is ok just a ques, y the split of the `$query`. Cant it be on the same line? @prondubuisi

muarachmann: We need to think coding style henceforth

muarachmann:
Also consider closing this #1285

interface/super/show_groups_popup.php
prondubuisi: made the change @muarachmann thanks

muarachmann: okay looks cleaner now @prondubuisi thanks

interface/super/show_groups_popup.php
```diff
@@ -48,8 +48,10 @@
 <h1><?php xl('Layout groups','e'); ?></h1>
 <ul>
 <?php
-$res = sqlStatement("SELECT distinct(group_name) FROM layout_options WHERE " .
-                    "form_id = '".$_GET['layout_id']."' ORDER BY group_name");
+
+$query .= "SELECT distinct(group_name) FROM layout_options WHERE form_id = ? ORDER BY group_name";
```
aethelwulffe: If you want to go multi-line (which is good for line length and reviews of the query elements) you can always just end your line of the query by putting in a space, closing the quote, adding a dot then adding the next quoted line vs. declaring the variable again with a .=
That space is important.  I personally like breaking out FROM, WHERE, AND, statements as much as practical unless it is a short line with AND ((statement) OR (statement)).
  I also like the concat dot and space at the beginning of the line, such as
"FROM layout_options"
." WHERE (statement);";
It is best to end your query strings with a semi-colon within the quotes.  I there is optional content, I prefer to concat a .";"; after all criteria is added, just so that no matter what, the final closing semicolon for the query and the final PHP statement end are always treated properly.

Issue #1401, opened 2019-02-02
Authenticated Unrestricted file upload when editing/updating patient profile picture


prondubuisi:
After a patients profile picture has been initially uploaded one can replace the profile picture with any type of file using the Edit Profile picture feature. 
![edit patient profile](https://user-images.githubusercontent.com/22311928/52161021-577e1300-26bf-11e9-81ef-78b9a98d765e.png)
![php file upload](https://user-images.githubusercontent.com/22311928/52161027-61077b00-26bf-11e9-8e7c-4bfde5007199.png)

prondubuisi:
I will like to fix this bug, I its confirmed.

aethelwulffe:
Isn't there a PR for this already? or was it just an issue?
-This one is a case of someone with superuser access being able to hack their own computer.  Still bad.

prondubuisi:
> Isn't there a PR for this already? or was it just an issue?
> -This one is a case of someone with superuser access being able to hack their own computer. Still bad.

@aethelwulffe , I looked through, there is no PR for this, can I work on it?

PR #1404, opened 2019-02-06
fix authenticated user unrestricted file upload while updating user profile picture #1401

prondubuisi:
create methods for image verifcation in other to avoid code duplication
 renamed variables for better code readabilty

This fix ensures that users can only edit profile images  with images and not other files


muarachmann:
I like the idea will check this once on PC

interface/patient_file/summary/demographics.php
```diff
@@ -1787,4 +1770,27 @@ function setMyPatient() {
   //show failure message.
 }
 
+//image mime check with all image formats.
+function has_right_mime($image_properties) {
+  $mime = $image_properties["mime"];
```
muarachmann: @prondubuisi nice idea to me. Just a thought we have many places where uploads is being done. Why not write this as a library then import so that we use them everywhere in the codebase. Just a suggestion. Lets here what @aethelwulffe has to say

aethelwulffe:
Depends on how sandboxed you want each of them.  I suppose a ?=pagefrom
can handle arguments for different behavior.

You have documents and image files of a variety of sorts going to the
patients (literally, this could be any word processor, DICOM, pdf or any
of a huge number of other file types) and then the photos, then
executable stuff.

Personally, I think we should look at Dropzone.js instead of treeview or
the others for that sort of thing.  The EHR need not become an image
editor either, but having the ability to add a warning when the file
properties are parsed by the JS to 1. indicate if the file upload will
work due to php settings, and 2. check the resolution to advise the user
that they are uploading a 2400dpi  color scan of a printed text document.

prondubuisi:
@aethelwulffe am a bit curious when you say 

> You have documents and image files of a variety of sorts going to the patients (literally, this could be any word processor, DICOM, pdf or any of a huge number of other file types) and then the photos, then executable stuff.

what do you mean by executable stuff? Like .exe files?

aethelwulffe:
Like .SQL in demo, dev, and certain production modes where the EHR is
used as a Quality Reporting tool.  Not really executable stuff.

prondubuisi:
By Library I think @muarachmann was referring to a means to verify uploads(images for now) in the backend. Will dropzone.js be able to make backend verification checks? maybe we could use it just in the frontend  @aethelwulffe

muarachmann:
take a look https://www.dropzonejs.com

aethelwulffe:
We are using it for the MIPS module specifically.  There is an
implementation of dropzone in the /sites/filemanager directory.  I
actually used two different tools to further isolate the upload from the
download/view.  Not sure if that helps much.  I used to have the code in
modules/MIPS, but with sandboxing that only allows the application to
view it's own root directory, you must have the .js directly up the
directory tree from the files it handles.  This could be changed, but as
I am not a security expert, I was not willing to attempt that at the
time.  All user content must, of course, be held in the sites/ directory.

prondubuisi:
@muarachmann whats the final take on this pull request, can you confirm that it solves the issue #1401

muarachmann:
@prondubuisi sorry haven't had time to test. From first glance, i would say all is ok but we have also pic uploads, document uploads in various iframes all around the EHR. and this will cause you to repeat thesame functions. So i would say you get the dropzone library as requested by @aethelwulffe make a test case then submit a PR, then if ok find similar cases where uploads are found and implement across EHR. You could make it like a milestone or gradually. IMHO. Else approving this will lead to more work done on this in the future. Lets make something worthwhile :)

prondubuisi:
@muarachmann am wondering, what role will the dropzone library play in a use case like this? Display role or Validation role or both? This will aid me in understanding what I am supposed to do

muarachmann:
its all of that @prondubuisi it will handle the upload, validations etc. u will just have to use iziToast to communicate the response

prondubuisi:
Is there any example of where the dropzone library is currently used in the codebase? I would love to take a look @muarachmann

muarachmann:
nope i dont think so its an enhancement to the system @prondubuisi i could include it and make a demo then you stem from there

prondubuisi:
@aethelwulffe I have updated the description for this pull request, it aims at fixing unrestricted file uploads only for the case of issue #1404

aethelwulffe:
Gotcha.  Understood.

robbyoconnor:
Resolve merge conflicts

prondubuisi:
@robbyoconnor this review took months hence the resulting merge conflicts. Resolving the conflict it might still take more months to merge leaving us in a infinite loop of resolving conflicts D:

aethelwulffe:
@prondubuisi Sailor's honor I will get this merged ASAP.  Sorry, but this is my dead-heat running time of the year, and we are lacking reviewers.  You might look at my own commit history and see that most stuff just dies of old age because no-one ever reviews it.
-Straight up telling you that if you want on the integration team, just start review testing and you are IN.

robbyoconnor:
I resolved the conflicts for ya @prondubuisi

robbyoconnor:
er cleaning up

robbyoconnor:
Yes --  was an accident on my part of resolving the conflicts

aethelwulffe:
Let's get them out, and if you would immediately pull it down and make sure that the pic system still works and all that, we can do this thing.

aethelwulffe:
You closed the PR Robby.

robbyoconnor:
ugh hold on

robbyoconnor:
(Sorry for trying to help and making a mess)

robbyoconnor:
@prondubuisi sorry for making a mess Pull in changes from here https://github.com/LibreHealthIO/lh-ehr/pull/1434 -- and then fix it :)

aethelwulffe:
OK.  this is a bit confusing.  Good enough that we now all actually know what the confusion is.

robbyoconnor:
Sorry @aethelwulffe

PR #1405, opened 2019-02-12
English xl() round 2

aethelwulffe:
Modifying UI text to make translation possible.

Currently working on the new translation list automated tool.
Interim solution is that I have taken all of the instances of:
xl()
xlt()
xls()
xla()
and put them into a spreadsheet.  This is due to the c++ parsing tool mysteriously truncating my results.  Probably an oversized array issue.

In iterations with the above, I am making some of the text more translatable.  There are instances where a single sentence had several breaks for punctuation or printed variables, with xlt() elements between.  THIS DOES NOT WORK.  Punctuation *must be* in the translated phrase or sentence so that the translator understands the context of that particular usage, and can convey the meaning.  Breaking it into pieces makes it impossible to translate, as no other languages have === English syntax.

Once the English base definitions (lang_constants) is done, that will supply the basis of our translation maintenance system.  We want a GUI for allowing the user to search in the constants, then modify the matching translations table entry for their language.

While I intend to retain the list of languages as options, I will be dropping all translated content, as it is so out of date as to be useless.  This is a do-over, but the hope is that the results even with automated translation values will be superior to what currently exists.  At that point, specialists can far more easily correct the translations.


aethelwulffe:
Anyone care to help review this thing so that translations can move forward?

muarachmann:
okay looking and this and kinna lost i have come accross only `xlt()` and `xl()` . Can you please tell me how it works that is how they work and the whole translation thingie :)

aethelwulffe:
Uh,  That is going to take me some time to give a rundown of everything in this.  Whew.  Basically, there is a lot of weird stuff in the translation functions.  Just searching the code base for one of the stranger ones like "xla(" grep should put you on the trace of where all the functions are defined, but now necessarily why.
  Most of the stuff in this relates to testing files to make sure they still put out their text properly in default language.

muarachmann:
OK just curious on why many of them xl , xla , xlt etc.
Are they all performing something different?

aethelwulffe:
Yes, but no-one really knows how or why!
Some take arguments.  Some don't.  Others are "hidden" functions we are not supposed to use.  Mess.  None of them were created or implemented by anyone that understands language or how to go about any of that.
  For years, the policy was that you could not include punctuation or things like colons in the sentence you were translating.  You speak more than one language; you know that is pure nuts.  You also can't break up a concept into parts, then translate those individually even if there are printed variables in them.  Syntax of the language you are translating to will almost never allow that.  Thus, it is a mess.

naveen17797:
what if implement a JS based translation system, so that it will decouple the translation logic from php and it will remove all the xlt, xl, xla tags from the existing pages, translation should occur once the page has been rendered in the browser (just like google translation plugin)
@aethelwulffe

aethelwulffe:
Tell me more...
Notes:  
We still need hard-coded definitions.  Medical terms and context are too important here.
Still need the (intended) functionality of having the in-application language editor, such as administration/other/language is supposed to do.
Any sources of data for ANYTHING in this should be in the database, not a combination of all of the DB and language files and some google doc somewhere.  On install, it should ONLY install required definitions, and those should be selected in setup.

-This may not be the path to go.  Maybe having the variables print out, then translate the screen can work...but I have my doubts.  We can't use a third party asset call out that allows data to be passed elsewhere.  Try a proof of concept with your google translate, by all means, and tell me what you think...unless you are thinking something that I am not.

naveen17797:
Actually we do the translation with our own translator but with JS, ie shift all the translation logic from php to js using a background threading method like ajax which can translate once the page has been rendered, according to my understanding /var/www/html/gramps-stuff/modules/language_translations/current_spreadsheet.tsv , the current translation system works on the based on the files where the corresponding values are mapped, we make a js function which parses the html and replace the text contents with the translated values.

muarachmann:
Correct Naveen this approach is great

aethelwulffe:
That is fine...BUUUT...we still need fixes for the STRUCTURE of the constants.  This means re-writing text strings so that they can make sense even when they have variables in them.  Whole concept translation.
So maybe we can do a little review and actual page-by-page testing here sometime?

naveen17797:
".we still need fixes for the STRUCTURE of the constants. " 
Couldn't get you

aethelwulffe:
@naveen17797 Means we can't translate strings that are:
text.variable.text
And have it come out with just any language.
We need "Full Concept Text:  ".$variable

Example:
English:  "My name is " . $Art
Latin:  $Art. "nomen meum."

This is real bad when the variable is in the middle of a sentence, and you have two function calls for each part of the sentence, each bit lacking context.  There are plenty of other instances when we need the context (location) of the string that will be translated (by someone) identified.  "E.T." can be Emergency Technician, Entertainment Tonight Magazine, or Extra Terrestrial.  We have to be able to give a user some context for what they are translating.  A big one is "No."  This can be the opposite of Yes, or it can be an abbreviation for "Number".

This stuff is tricksy.

aethelwulffe:
Constant === English String value echoed on the screen.

naveen17797:
yeah i get that now @aethelwulffe, altering sentence format might be little tricky, there should be a way to do it, translations should resemble like template format.Like we wold be able to add the translation like
"my name is  {name_var}" and the latin version be like "{name_var} nomen meum" stored in a csv file and there are lot of other validations may arise in the module.

Issue #1361, opened 2018-12-09
Selecting Procedure in Batch Results Opens Up Inappropriate Module


jonathanzhang53:
## Important Note

This will require you to test locally to fully understand (unless I just wrote such a great description). The issue is not really a clear code fix.

## Description

When you try to see batch results for a certain procedure, selecting a procedure leads to a list of Types of Orders and Results which you can select, but clicking save does not do anything and does not select the procedure. I have tried different combinations of interactions but there's no way to actually select a procedure.

## Reproduce

Login to EHR locally.
Create a procedure if one is not yet created by going to Procedures > Configuration and adding a top level (the easiest to fill out is a group type procedure).
Go to Procedures > Batch Results.
Click the procedures input box to reveal a popup module.
Click save or add or anything else but a procedure still isn't selected for the input

## Fix

The popup has the functionality of Procedures > Configuration but that is not what is needed in this case. All that is needed is a list of the orders and the ability to select one and confirm selection.

## GIF

![selectprocedure](https://user-images.githubusercontent.com/41968151/49196748-bf2b8100-f351-11e8-8b6c-be613e938cd3.gif)

Again, I recommend going on your own demo to test this.

robbyoconnor:
Not 100% sure these labels -- someone can fix them...just trying to triage things for GCI metric purposes.

jonathanzhang53:
@robbyoconnor I'm not going to approach this for GCI. I just found it and I wanted to report it.

punwai:
I'll try and work for this although it is off my tasks, I'm doing grab bag. Is that fine by you? @robbyoconnor

aethelwulffe:
@tmccormi @teryhill 
Where are we with this particular form?  Does any of it also relate to LIMS?  This isn't something we are looking to deprecate or anything is it?  I don't use the system at all, so I don't know much about it.

mshubhankar:
I'm a new contributor and am finding my first issue to work on. This seems like a good start. What is exactly expected ? 

These are my observations
- The cancel button needs to be fixed.
- The save button needs to close the window instead of popping it back up again.
-  The add for each row opens up procedures->configurations module which needs to be removed and instead the ability to select a row needs to be added. 

Have I got it correctly ?

jonathanzhang53:
[...]

PR #1410, opened 2019-02-23
Solved issue #1361

mshubhankar:
Changes made:
-> Ported the find-procedure popup to iziModal style.
-> Fixed the cancel button.
-> Save button name changed to Refresh.

### DIFF
--- a/interface/orders/orders_results.php
+++ b/interface/orders/orders_results.php
@@ -231,7 +231,8 @@ function sel_proc_type(varname) {
     console.log(event.target.href);
     $("#modal-iframe").iziModal({
     title: 'Find procedure',
-    subtitle: 'Select procedure from the table below', headerColor: '#88A0B9',
+    subtitle: 'Select procedure from the table below',
+    headerColor: '#88A0B9',
     iframe: true,
     iframeHeight: 500,
     width: 700,
--- a/interface/orders/orders_results.php
+++ b/interface/orders/orders_results.php
@@ -208,17 +208,41 @@ a, a:visited, a:hover { color:#0000cc; }
 <script type="text/javascript" src="../../library/dialog.js"></script>
 <script type="text/javascript" src="../../library/textformat.js"></script>
 <script type="text/javascript" src="../../library/js/jquery-1.9.1.min.js"></script>
-
+<?php call_required_libraries(array("jquery-min-3-1-1","bootstrap", "iziModalToast")); ?>
 <script language="JavaScript">
 
 
-// This invokes the find-procedure-type popup.
+// This invokes the find-procedure-type popup. Old popup style iframe replaced with iziModal. 
 var ptvarname;
+// function sel_proc_type(varname) {
+//  var f = document.forms[0];
+//  if (typeof varname == 'undefined') varname = 'form_proc_type';
+//  ptvarname = varname;
+//  dlgopen('types.php?popup=1&order=' + f[ptvarname].value, '_blank', 800, 500);
+// }
+
+//This invokes the find-procedure-type popup in iziModal iframe.
 function sel_proc_type(varname) {
  var f = document.forms[0];
  if (typeof varname == 'undefined') varname = 'form_proc_type';
  ptvarname = varname;
- dlgopen('types.php?popup=1&order=' + f[ptvarname].value, '_blank', 800, 500);
+ // dlgopen('types.php?popup=1&order=' + f[ptvarname].value, '_blank', 800, 500); // This calls the old style popup
+ event.preventDefault();
+    console.log(event.target.href);
+    $("#modal-iframe").iziModal({
+    title: 'Find procedure',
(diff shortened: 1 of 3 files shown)


interface/orders/orders_results.php
```diff
@@ -208,17 +208,41 @@ function QuotedOrNull($fld) {
...
+ // dlgopen('types.php?popup=1&order=' + f[ptvarname].value, '_blank', 800, 500); // This calls the old style popup
+ event.preventDefault();
+    console.log(event.target.href);
+    $("#modal-iframe").iziModal({
+    title: 'Find procedure',
+    subtitle: 'Select procedure from the table below', headerColor: '#88A0B9',
+    iframe: true,
+    iframeHeight: 500,
+    width: 700,
+    height: 500,
```
muarachmann: @mshubhankar good work done here. Just some few stuffs to be done. izi headers on the other modals

mshubhankar: I'm sorry I couldn't fully understand. Which other modal are you exactly taking about ? Is it the Add button or Add top level modal ?

muarachmann:
The add

mshubhankar:
The add and add top level iziModals are written in the same definition. I have added a generic header which will work for both modals.

interface/orders/types.php
```diff
@@ -191,6 +191,8 @@ function refreshFamily(id, haskids) {
     event.preventDefault();
     console.log(event.target.href);
     $("#modal-iframe").iziModal({
+    title: 'Add procedure',
+    subtitle: 'Fill in procedure', headerColor: '#88A0B9',
```
muarachmann: a break here will do .... headerColor is a key value

mshubhankar: Yes you are right. I have made the necessary changes to both new definitions.

muarachmann:
sorry @mshubhankar havent had time, testing this in a bit

mshubhankar:
@muarachmann Haven't heard from you in a while. I just wanted to check if everythings okay with the changes I made.

muarachmann:
> @muarachmann Haven't heard from you in a while. I just wanted to check if everythings okay with the changes I made.

Yes actually have been pretty busy!. I think your PR is cool will review it some other time please. You can solve other issues :) Thanks and good work

interface/orders/types.php
```diff
@@ -192,6 +191,9 @@ function refreshFamily(id, haskids) {
     event.preventDefault();
     console.log(event.target.href);
     $("#modal-iframe").iziModal({
+    title: 'Add procedure',
+    subtitle: 'Fill in procedure',
```
aethelwulffe: Translations?  Meaning use of xlt() function.

muarachmann: yes use those i forgot thanks @aethelwulffe

aethelwulffe:
Also, the way you called the libraries is pretty much the common way it is done, but I believe something like this:
	<?php call_required_libraries(['bootstrap', 'jquery-min-1-9-1']); ?>
Also works fine without using array()

interface/orders/orders_results.php
```diff
@@ -208,17 +208,42 @@ function QuotedOrNull($fld) {
...
+<?php call_required_libraries(array("jquery-min-3-1-1","bootstrap", "iziModalToast")); ?>
 <script language="JavaScript">
 
 
-// This invokes the find-procedure-type popup.
+// This invokes the find-procedure-type popup. Old popup style iframe replaced with iziModal. 
 var ptvarname;
+// function sel_proc_type(varname) {
+//  var f = document.forms[0];
+//  if (typeof varname == 'undefined') varname = 'form_proc_type';
```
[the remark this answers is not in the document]
mshubhankar: Yeah, you're right. I will remove the junk.

interface/orders/types.php
mshubhankar: Sorry but I do not understand fully.

muarachmann: xlt() is a translation function @mshubhankar so basically just pass the texts as parameter to it so it displays a correct message in the appropriate language

mshubhankar:
@aethelwulffe Yeah the libraries can be called without array() as well. I refered to some previous code and make it this way. Here are some of the instances :
screenshot 2019-03-08 at 1 30 17 pm

aethelwulffe:
Yeah, you did good, it was just that when I wrote the function I assumed that it would be passed as an array without anyone needing to be explicit.  Since then most all the implementations have been explicit.  Just now noticed.

aethelwulffe:
Mua,
Review, merge.

mshubhankar:
Hey @muarachmann, I hadn't pushed my last commit. I had to make some changes in the PR suggested by @aethelwulffe.

mshubhankar:
I have created another PR with the uncommitted files. Kindly refer #1424

Issue #1413, opened 2019-03-01
Able to add invalid email when adding pharmacy


Ngai-E:
A)  **Outreachy Username:** elizabeth

B) **Issue Title:** Able to add invalid email when adding pharmacy.

C) **Bug Report Date**: Thursday, 28th  February 2019.

D) **Site Affected:** Documentation Site and NHANES

E) **OS/ Browser Used**: Chrome  on Ubuntu 16.04

F) **Which Workflow Module in LHEHR**: Administration => Practice Settings

G) **Steps to Reproduce the Bug:**
   1. Login to the LibreHealth EHR platform.
   2. Move on to the Administration menu item and select Practice from the drop down options.
   3. Click on Pharmacies and then click on Add a Pharmacy.
   4. Fill in a pharmacy Name (which is required).
   5. Fill in a wrong Email say “thisIsWrongEmail Right?”.
   6. Click on the Save button and it will be saved successfully to the database without throwing a 
      validation error.

H) **At Point of the Bug, the Expected Behavior:** The expected behavior was to prompt the user to input a valid email address before clicking save.

I) **Details of What Actually Happened:** Typed in an invalid email address when adding a pharmacy and it was accepted.

J) **Provide Relevant Screenshot and Explain the Screenshot:** 
![image](https://user-images.githubusercontent.com/22508583/53660744-eb131700-3c5e-11e9-8fff-480125537a89.png)
Illustration 1: Select Practice from Administration

![image](https://user-images.githubusercontent.com/22508583/53660880-3299a300-3c5f-11e9-92b7-2fb5d8a30e66.png)
Illustration 2: Select Pharmacies

![image](https://user-images.githubusercontent.com/22508583/53660897-3c230b00-3c5f-11e9-979a-931795a07c5b.png)
Illustration 3: Click on Add a Pharmacy button

![image](https://user-images.githubusercontent.com/22508583/53660911-480ecd00-3c5f-11e9-9081-e2dbf3660c93.png)
Illustration 4: Fill in invalid email as shown in this example and click save which will then save successfully

K) Estimation of The Bug Severity: Module continues with ongoing errors (Critical).

muarachmann:
thanks for your report @Ngai-E we are here to help you fix this is you want

Ngai-E:
> thanks for your report @Ngai-E we are here to help you fix this is you want

[...]

PR #1423, opened 2019-03-08
Solved issue #1413

Ngai-E:
1. changed the email input type from 'text' to 'email'
2. validated filled email entry input to check fo validity using javaScript regex pattern match before saving.

### DIFF
--- a/templates/pharmacies/general_edit.php
+++ b/templates/pharmacies/general_edit.php
@@ -33,7 +33,7 @@
 <tr>
     <td  ><?php echo xlt("Email");?></td>
     <td >
-        <input type="text" class="form-control input-sm" NAME="email" SIZE="35" VALUE="<?php echo $this->pharmacy->email;?>" onKeyDown="PreventIt(event)" />
+        <input type="email" class="form-control input-sm" NAME="email" SIZE="35" VALUE="<?php echo $this->pharmacy->email;?>" onKeyDown="PreventIt(event)" />
     </td>
 </tr>
 <tr>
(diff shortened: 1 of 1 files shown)


muarachmann:
yes Elizabeth thank you so much. I have some few request though :)

Ngai-E:
> yes Elizabeth thank you so much. I have some few request though :


> yes Elizabeth thank you so much. I have some few request though :)

okay what are the requests?

Ngai-E:
> @Ngai-E Elizabeth, if you are happy with the 35 character field size (SHOULD be enough!) then this is a good clean simple PR, and seems to function as desired.

Thanks

aethelwulffe:
Yeah dude, what are the requests?  What did I miss in the review specifically?  Should we have used a different database type for this as well?  I didn't look into that.

templates/pharmacies/general_edit.php
```diff
@@ -83,10 +83,24 @@
...
+
+            if(regex.test(email)){
+                top.restoreSession();
+                document.pharmacy.submit();
+                //Z&H Removed redirection
+            }
+
+            else{
+                alert("please enter a valid email address");
+            }
```
muarachmann: Here we intend to communicate messages to the user using izi-Toast - http://izitoast.marcelodolce.com/ it is included already in the codebase see its usage here https://github.com/LibreHealthIO/lh-ehr/pull/1392/files#diff-ac491a14571b3e954dca9d6842deaeb2 and

muarachmann: Also remove some trailing spaces and you are good to go

muarachmann:
> Yeah dude, what are the requests? What did I miss in the review specifically? Should we have used a different database type for this as well? I didn't look into that.

oh nothing so complicated its actually good. just an enhancement :)

Ngai-E:
Ok I will take a look at it @muarachmann  thanks for the review.

muarachmann:
Great just that and you are good to go

Issue #1415, opened 2019-03-05
Invalid date inputs were accepted while creating MIPS report


Priyanka488:
A) <b>Outreachy username :</b> priyanka.yadav

B) <b>Issue title :</b> Invalid date inputs were accepted while creating MIPS report

C) <b>Site affected:</b> The documentation site

D)<b> Bug report date :</b> March 5, 2019

E) <b>OS/ browser used: </b>Windows/Chrome

F) <b>Which workflow module in LHEHR : </b>“Create MIPS Report” under “QA Measures”

G) <b>Steps to reproduce the bug : </b>
1. ) Login with your account credentials.
2. ) Click on QA Measures from the top menu , and select Create MIPS Report.
3. ) Enter Begin date and End date in such a way that End date is earlier than the Begin Date.
4. ) Click on Submit.

H) <b>At point of bug, the expected behavior :  </b> When we enter invalid begin and end dates, there should be a prompt refraining the successful submission of the report , and user should be asked to enter valid dates.

I) <b>Details of what actually happened : </b>There was no prompt or alert box for the invalid dates. The details were successfully submitted and the option to print appeared.

J) <b>Provide relevant screenshots :</b>
A. <b>Entering invalid begin and end dates.</b>

![screenshot 139](https://user-images.githubusercontent.com/33698440/53821549-026e3f00-3f94-11e9-844b-7e5ed400d27d.png)

B.<b> After successful submission, print option appeared  :</b>

![screenshot 140](https://user-images.githubusercontent.com/33698440/53821563-08fcb680-3f94-11e9-932b-e5ef2f86d958.jpg)

K) <b> Estimation of the bug severity :</b> module function continues with ongoing errors.

PR #1425, opened 2019-03-09
Fix:#1415 - Validation of dates done

Priyanka488:
Changes made in modules/MIPS/clinicalmeasures.php to check for valid end and begin dates.

### DIFF
--- a/modules/MIPS/clinical_measures.php
+++ b/modules/MIPS/clinical_measures.php
@@ -227,6 +227,17 @@ function GenXmlMIPS(sNested) {
 
 
 function validateForm() {
+var d = document.forms[0];
+
+  FromDate = d.form_begin_date.value;
+  ToDate = d.form_target_date.value;
+
+  if(FromDate.length > 0 && ToDate.length > 0) {
+    if(FromDate > ToDate) {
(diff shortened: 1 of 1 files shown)


modules/MIPS/clinical_measures.php
```diff
@@ -227,6 +227,17 @@ function GenXmlMIPS(sNested) {
...
 
 function validateForm() {
+var d = document.forms[0];
+
+  FromDate = d.form_begin_date.value;
+  ToDate = d.form_target_date.value;
+
+  if(FromDate.length > 0 && ToDate.length > 0) {
+    if(FromDate > ToDate) {
+      alert("<?php echo xls('End date must be later than Begin date!'); ?>");
```
muarachmann: use iziModalToast lib here to report msgs to the user http://izitoast.marcelodolce.com/ and see some sample usage here https://github.com/LibreHealthIO/lh-ehr/pull/1395/commits/db8f6b670c555193d15de3f866ec479c6a89fb53

Priyanka488: @muarachmann Okay ! will replace the alert statement.

Issue #1429, opened 2019-03-10
Deleting Reports can not be determined if successful or not until page is refreshed


Ngai-E:
**A) Outreachy Username:**  elizabeth

**B) Issue Title**: Deleting Reports can not be determined if successful or not until page is refreshed.

**C) Bug Report Date:** Friday, 10th  March 2019.

**D) Site Affected:** Documentation Site and NHANES

**E) OS/ Browser Used:** Chrome  on Ubuntu 16.04

**F) Which Workflow Module in LHEHR:** Report History/Results

**G) Steps to Reproduce the Bug:**
1. Login to LibreHealth EHR.
2. Select the View Reports drop down option from the QA Measures menu item.
3. Click on Delete on any of the reports.

**H) At Point of the Bug, the Expected Behavior:** User expects the  report to be deleted and not to be visible anymore.

**I) Details of What Actually Happened**: Report is still visible even after it was deleted.

**J) Provide Relevant Screenshot and Explain the Screenshot:** 
![records](https://user-images.githubusercontent.com/22508583/54090114-fc4fd800-4370-11e9-91e0-f700bf59464b.png)
Illustration 1: Showing records 

**K) Estimation of The Bug Severity**: Module function continues with workarounds (Major).

**L) Workaround:** The user can successfully delete the record if the record is deleted twice, Refresh button is clicked or the entire page is refreshed.

muarachmann:
@Ngai-E not sure of this. I can actually delete reports @koniKodes please verify this

PR #1431, opened 2019-03-12
Solved #1429 Deleting Reports can not be determined if successful or not until page is refreshed 

Ngai-E:
removed the html row of the where the delete button belongs to when ajax returns success so as to sync backend with front end code.


Ngai-E:
Thanks

aethelwulffe:
No, thank YOU!

Issue #1438, opened 2019-03-13
MySQL errors showing up in search submission


Ngai-E:
A) **Outreachy Username:**  elizabeth

B) **Issue Title:** MySQL errors showing up in search submission exposing database structure of LibreHealth EHR.

C) **Bug Report Date:** March 13,  2019.

D) **Site Affected:** Documentation Site and NHANES

E) **OS/ Browser Used:** Chrome  on Ubuntu 16.04

F) **Which Workflow Module in LHEHR:** Report - Sales by Item

G) **Steps to Reproduce the Bug:**
1. Login to LibreHealth EHR.
2. Select Reports menu item.
3. Hover over Financial drop down option and click on Sales by Item.
4. Choose a facility from the Facility drop down menu which is not –All Facilities--.
5. Choose a From and To date range  (Optional to reproduce bug).
6. Click on Submit.
7. Similarly, choose a provider from the Provider drop down which is different from –All Providers-- and repeat steps 5 and 6.

H) **At Point of the Bug, the Expected Behavior:** User expects to see a list of sold items displayed in an orderly manner, say, a table for example.

I) **Details of What Actually Happened:** User is presented with a bunch of MySQL error complaining of syntax in SQL statement and a dump of failed database querry.

J) **Provide Relevant Screenshot and Explain the Screenshot:**
![sqlerror](https://user-images.githubusercontent.com/22508583/54292902-4a521f00-45af-11e9-995e-189928cc8152.png)
Illustration 1: Submit button returns MySQL errors

K) **Estimation of The Bug Severity:** Module function continues with ongoing errors (Critical).

aethelwulffe:
This is due to your server error reporting settings.  This is not a bug, just a developer-level php.ini configuration issue.  
I use something like this in my config 
// toggle this to change the setting
define('DEBUG', true); 
// you want all errors to be triggered
error_reporting(E_ALL); 

if(DEBUG == true)
{
    // you're developing, so you want all errors to be shown
    display_errors(true);
    // logging is usually overkill during dev
    log_errors(false); 
}
else
{
    // you don't want to display errors on a prod environment
    display_errors(false); 
    // you definitely wanna log any occurring
    log_errors(true); 
}
Otherwise just turn off display_errors in php.ini

aethelwulffe:
What caused this is possibly a (currently) bad query, or lack of data without sanity checking database  or globals configuration.  I doubt the warehouse option is turned on in globals for the site you are looking at.

[...]

PR #1443, opened 2019-03-15
Solve #1438 MySQL errors showing up in search submission

Ngai-E:
@aethelwulffe @muarachmann  I succeeded to make this fix kindly review and give me feedback.
Screenshot from 2019-03-15 12-26-30
After inputting selecting desired drop down menu options and clicking submit, results are shown as needed

### DIFF
--- a/interface/reports/reports_controllers/SalesByItemController.php
+++ b/interface/reports/reports_controllers/SalesByItemController.php
@@ -261,14 +261,13 @@ function prepareQuery() {
     "JOIN drugs AS d ON d.drug_id = s.drug_id " .
     "JOIN form_encounter AS fe ON " .
     "fe.pid = s.pid AND fe.encounter = s.encounter AND " .
-    "fe.date >= '$from_date 00:00:00' AND fe.date <= '$to_date 23:59:59' " .
-    //"fe.date >= ? AND fe.date <= ? " .
+    "fe.date >= ? AND fe.date <= ? " .
     "WHERE s.fee != 0";
-    array_push($sqlBindArray,$from_date,$to_date);
+    array_push($sqlBindArray,$from_date . '00:00:00',$to_date . '23:59:59');
   // If a facility was specified.
   if ($form_facility) {
     $query .= " AND fe.facility_id = ?";
-     array_push($sqlBindArray,$form_facility);
+    array_push($sqlBindArray,$form_facility);
   }
   if ($form_provider) {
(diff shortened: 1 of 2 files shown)


interface/reports/reports_controllers/SalesByItemController.php
```diff
@@ -262,21 +262,21 @@ function prepareQuery() {
...
     "fe.date >= '$from_date 00:00:00' AND fe.date <= '$to_date 23:59:59' " .
-    //"fe.date >= ? AND fe.date <= ? " .
     "WHERE s.fee != 0";
     array_push($sqlBindArray,$from_date,$to_date);
   // If a facility was specified.
   if ($form_facility) {
-    $query .= " AND fe.facility_id = ?";
-     array_push($sqlBindArray,$form_facility);
+    $query .= " AND fe.facility_id = $form_facility";
+     //array_push($sqlBindArray,$form_facility);
```
aethelwulffe: drop old code.

aethelwulffe:
@muarachmann You see any issues here?  Test is good, Just trying to figure out what the heck happened to the array bind that presumably worked at one time (or did it?)

Ngai-E:
@aethelwulffe  changes done!

KoniKodes:
Hi, is it possible to have a new reviewed or merged?
This is the last week of the application round and I'm trying to get all the ducks in a row...

Ngai-E:
@muarachmann @aethelwulffe
Made the changes you requested kindly review. I have been waiting

muarachmann:
> @muarachmann You see any issues here? Test is good, Just trying to figure out what the heck happened to the array bind that presumably worked at one time (or did it?)

I too wondering. Just worried about sql injection. But i think it is a good fix for the errors

Ngai-E:
@muarachmann will the array bind prevent the mysql injection? If so, i think u better guide me how i can go about implementing it. I would love to improve the code if possible except you are ok with the solution so far

muarachmann:
Yes the array bind will replying from phone. You could try by doing a dump
and die before the query runs to see the output. I am sure the array is
empty reason for that error. Just try and see how to better parse it

Ngai-E:
@muarachmann i did print the content of the array and noticed it was empty. You are right that was the reason for the error

muarachmann:
Okay... So ensure you are getting it correctly then the bind should work

Issue #1441, opened 2019-03-14
Performing patient Ledger search keeps displaying error message


Ngai-E:
A) **Outreachy Username:**  elizabeth

B) **Issue Title**: Can not select search patient Ledger in module Report – Financial -     Patient Ledger By Date .

C) **Bug Report Date**: March 14, 2019.

D) **Site Affected**: Documentation Site and NHANES

E) **OS/ Browser Used**: Chrome  on Ubuntu 16.04

F) **Which Workflow Module in LHEHR**: Report – Financial - Patient Ledger By Date

G) **Steps to Reproduce the Bug:**
1. Login to LibreHealth EHR.
2. Select Reports menu item.
3. Hover over Financial drop down option and click on Patient Ledger By Date.
4. Choose a From and To date range  (Optional to reproduce bug).
6. Select a patient and click on Submit.

H) **At Point of the Bug, the Expected Behavior**: User expects to see a list of patient ledgers displayed in an orderly manner, say, a table for example.

I) **Details of What Actually Happened**: User keeps getting the error message ‘No data to display. Select patient and try search again.’ and whether or not the user selects a patient who visited the hospital in a particular selected date range, user still is unable to see patient ledger with the same error message ‘No data to display. Select patient and try search again.’.

J) **Provide Relevant Screenshot and Explain the Screenshot:**

			
	
![image](https://user-images.githubusercontent.com/22508583/54392455-068f1080-46a8-11e9-8c23-c1ea27af844e.png)
Illustration 1: showing visual  steps to reproduce bug 

![PatientLedger](https://user-images.githubusercontent.com/22508583/54392618-6e455b80-46a8-11e9-853c-bb326ed8c99f.png)
Illustration 2: Error message still persist even after selecting patient with correct date.

 

K) **Estimation of The Bug Severity:** Module function continues with ongoing errors (Critical).

aethelwulffe:
Patient has to have authorized billing and payments for this report to display data.  We will look into the queries it runs to display the same.
  Despite the feelings of certain parties, we cannot use the Documentation or NHANES site to do troubleshooting, as there is no access to the database (phpmyadmin specifically) to determine the state of the data or make direct changes to affect that.  Thus, using a local copy with some data is a good idea for anyone wishing to actually troubleshoot or do development.
  I will hunt what I can after work.

Ngai-E:
> Patient has to have authorized billing and payments for this report to display data. We will look into the queries it runs to display the same.
[...]

PR #1457, opened 2019-03-21
Solved #1441: Ledger search displays inappropriate error message

Ngai-E:


Ngai-E:
errorMessagefix
Fix showing a more explicit error message

Ngai-E:
@aethelwulffe @muarachmann I improved on the error message. kindly review

interface/reports/pat_ledger.php
```diff
@@ -213,7 +213,7 @@ function sel_patient() {
 if (! $_REQUEST['form_csvexport']) {
   if ( $_REQUEST['form_refresh'] && $orow <= 0) {
     echo "<span style='color:red;'>";
-    echo xlt('No data to display. Select patient and try search again.');
+    echo xlt('No data to display for this patient.');
```
muarachmann: Hmmmm was this reported as a bug?

muarachmann: I don't think the two statement makes any difference ie `No data to display. Select patient and try search again.` and `No data to display for this patient.`

Ngai-E: > I don't think the two statement makes any difference ie `No data to display. Select patient and try search again.` and `No data to display for this patient.`

The issue was user already selected a patient before performing search. So the current error message is confusing and user might keep trying to select the same patient over and over without knowing the actual cause of no results displaying. You can check the issue and comments for more clarification.

KoniKodes: @muarachmann This was an approved issue. It was suggested that @Ngai-E create a better error message. 
> But what I expect to happen if no billing data was found, is an appropriate response such as 'Sorry no billing data was found for this patient' instead of 'No data to display. Select Patient and try again.' meanwhile a patient was selected. User can get confused as to the actual cause of failure.

aethelwulffe:
This report is kinda junk anyway...but this is an OK change.

KoniKodes:
Thank you.

PR #1471, opened 2019-05-09
Report generator disable

aethelwulffe:
Disabling the report generator due to install errors.  .env file related stuff in library/installer.class.php cause the stuff to break.  Testing a full install now, and will merge once fixed.  Code did not affect existing installations.


aethelwulffe:
Install now works.  Merging this.

PR #1473, opened 2019-05-10
Version 2.02 RC from SunCoast Connection.

aethelwulffe:
Release working branch for a probable 2.1.0 version
Features 2018 QA measures
Adds V22 of the Health Condition Category standards
Adds Drop and Drag appointments (Terry Hill)
Adds Enhanced Authorization features (Terry Hill Preliminary stuff)
Adds some enhanced Insurance and Facility Administration (Terry Hill)
Adds some ACL base on facility features
Some change files reflect minor whitespace difference.

Release issues:

- [x] issue 1
- [ ] issue 2
- [ ] issue 3
- [ ] issue 4
- [ ] issue 5
- [ ] issue 6
- [ ] issue 7
- [ ] issue 8
- [ ] issue 9
- [ ] issue 10
- [ ] issue 11
- [ ] issue 12
- [ ] issue 13
- [ ] issue 14
- [ ] issue 15

PR #1479, opened 2019-05-16
fixed setup installation script

muarachmann:
This PR attempts to fix the setup installation process for EHR now found in `/modules/setup` .

=====
[x] create temporal file for for temporal track of setup installation process
[x] removed unsed files
[] move all styles and js, images to `assets/`
[] optimize the code to show real time percentages from mysql query (like measure actual execution of each query to provide appropriate feedback to user)
[] Revamp multi-site login (re-installs)

.....

NB - This currently installs EHR and works fine. Except for clone source database which i dont know why the function doesn't work. Looking forward to solving this

```
private function dumpSourceDatabase() {
        global $OE_SITES_BASE;
        require("$OE_SITES_BASE/$this->source_site_id/sqlconf.php");

        if (empty($config)) die("Source site $this->source_site_id has not been set up!");

        $backup_file = $this->get_backup_filename();
        $cmd = "mysqldump -u " . escapeshellarg($this->login) .
            " -p" . escapeshellarg($this->pass) .
            " --opt --quote-names -r $backup_file " .
            //" --opt --skip-extended-insert --quote-names -r $backup_file " .  Comment out above line and enable this to have really slow DB duplication.
            escapeshellarg($this->dbname);

        $tmp0 = exec($cmd, $tmp1=array(), $tmp2);
        if ($tmp2) die("Error $tmp2 running \"$cmd\": $tmp0 " . implode(' ', $tmp1));

        return $backup_file;
    }
```
PS - Should be merged after #1473


aethelwulffe:
We also need to rebuild the database.sql file.
Two choices.
1.  Create a separate file for each table, maintain each separately, and have the installer run files in sequence, THEN do insert files.
2.  Clean database.sql up, organize it, then re-write or dump all the inserts so they do not do 12 million queries.  Installing the db can take about 5 seconds when optimized.  The way it is now, it can potentially time out.

muarachmann:
Okay i see the error here. The issue is the dump was only done considering two systems linux and windows. Adding the patch for mac.

aethelwulffe:
Assets move next?  You can keep them in the setup module if you like, but since it is (or is to be) a core module, nothing wrong with using the standard assets.

Multi-site (adding sites, updating sql for sites), requiring a localhost or pre-set install access auth to run setup (which can be done without root db pass remotely if db is local to apache), and removing all other setup tools to prevent use by unauthorized access (while not requiring anyone to remove setup.php) are higher on the list than the install countdown.  Frankly, if we optimize database.sql, it might be so quick we don't need a timer.  I think that issue might already be taken care of in the SCCBranch monster PR.  It runs much faster now, and still has room for improvement, after I remove a few thousand semicolons and combined insert queries.

PR #1494, opened 2019-08-06
Update .gitignore

### LINKED ISSUE #1477: Modify .gitignore to exclude all directories in /sites dir except default
When installing EHR for first or n times, you might want to have multiple instances where you can access via some `site_id=xxxx` doing this will lead to creation or mirror of an already existing sites dir in the `/sites` dir.  We dont need this to show up on git commits

- ignore all folders in `/sites` except for default site 
NB - sqlconfig.php for `/sites/default` is not needed equally you can check this #1388

Patil2099:
Fixes #1477

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,7 @@
 .idea/
 sites/*/sqlconf.php
 sites/*/profile_pictures/*
+sites/*
+!sites/default
 
 backups/
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,6 @@
 .idea/
 sites/**
 !sites/default
+sites/default/sqlconfig
 
 backups/
--- a/.gitignore
+++ b/.gitignore
@@ -1,9 +1,7 @@
 *.env.*
 .htaccess
 .idea/
-sites/*/sqlconf.php
-sites/*/profile_pictures/*
-sites/*
+sites/**
 !sites/default
 
 backups/


.gitignore
```diff
@@ -3,5 +3,7 @@
 .idea/
 sites/*/sqlconf.php
 sites/*/profile_pictures/*
+sites/*
+!sites/default
```
muarachmann: I think this will do @Patil2099 . Only one thing removing this `sites/*/profile_pictures/*` cause `sites/*` will take care of that and also `sites/*/sqlconf.php` . Now here is the thing we want the stuffs under **default** to show up if modified , altered by git except for its `sqlconf.php` cause absolutely this has caused trouble :). The rest can be discarted.

Once more welcome and Nice having you onboard!!

Patil2099:
@muarachmann  Is it okay now ?

muarachmann:
> @muarachmann Is it okay now ?

Nope we need the default site. except for its sqlconfig

.gitignore
```diff
@@ -1,7 +1,7 @@
 *.env.*
 .htaccess
 .idea/
-sites/*/sqlconf.php
-sites/*/profile_pictures/*
+sites/**
+!sites/default
```
muarachmann: This is ok but we need not `default's sqlconfig`

Patil2099: should i add !sites/*/sqlconf.php @muarachmann

muarachmann:
We don't need the default sites sqlconfig. but we need its other contents.
I think that is self explanatory right?

**!!sites/default/sqlconfig**

Patil2099:
@muarachmann Is it okay now?

muarachmann:
> @muarachmann Is it okay now?

LGTM. Thanks @Patil2099

muarachmann:
Sorry this took long @Patil2099 . Thanks for your contribution and welcome on board :). Hope to see you around. If you have any questions feel free to reach us in the chat [here](https://chat.librehealth.io/channel/general)

PR #1497, opened 2019-08-13
Fixed filter function of the Patient List Reports module

mhn94:
Fixes Issue #1496 

Before, the Patient List Report module would show the same results no matter the input. Now, the module filters correctly accordingly to input.

For example, now:

- submitting the form with the default values ("All Providers" and no date range selected) outputs:

no-input-results-fixed

- submitting the form with a provider selected, but no date range outputs:

w-patient-no-date-selected-fixed

- submitting the form with a provider and date range outputs:

w-patient-from-and-to-dates-fixed

### DIFF
--- a/interface/reports/reports_controllers/PatientListController.php
+++ b/interface/reports/reports_controllers/PatientListController.php
@@ -33,45 +33,45 @@ function qescape($str) {
 	return str_replace('"', '\\"', $str);
 }
 
-// $from_date = fixDate($_POST['form_from_date'], date('Y-01-01'));
-// $to_date   = fixDate($_POST['form_to_date'], date('Y-12-31'));
-$from_date = fixDate($_POST['form_from_date'], '');
-$to_date   = fixDate($_POST['form_to_date'], '');
-if (empty($to_date) && !empty($from_date)) $to_date = date('Y-12-31');
-if (empty($from_date) && !empty($to_date)) $from_date = date('Y-01-01');
-
-$form_provider = empty($_POST['form_provider']) ? 0 : intval($_POST['form_provider']);
-
 /* This prepares and displays results for Patient List report
  * @params: None
  * @return: Echo HTML
  * @author: Tigpezeghe Rodrige K. <tigrodrige@gmail.com>
+ * @author: Maggie Negm <maggiehn94@gmail.com>
  * */
 function prepareAndShowResults() {
 	$totalpts = 0;
+
+	$from_date = fixDate($_POST['form_from_date'], '');
+	$to_date   = fixDate($_POST['form_to_date'], '');
+	if (empty($to_date) && !empty($from_date)) $to_date = date('Y-12-31');
+	if (empty($from_date) && !empty($to_date)) $from_date = date('Y-01-01');
+
+	$form_provider = empty($_POST['form_provider']) ? 0 : intval($_POST['form_provider']);
+
   	$query = "SELECT " .
 	   "p.fname, p.mname, p.lname, p.street, p.city, p.state, " .
 	   "p.postal_code, p.phone_home, p.phone_biz, p.pid, " .
-	   "count(e.date) AS ecount, max(e.date) AS edate, " .
+	   "count(e.pc_eventDate) AS ecount, max(e.pc_eventDate) AS edate, " .
 	   "i1.date AS idate1, i2.date AS idate2, " .
 	   "c1.name AS cname1, c2.name AS cname2 " .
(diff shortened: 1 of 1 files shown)


interface/reports/reports_controllers/PatientListController.php
```diff
@@ -33,45 +33,45 @@ function qescape($str) {
...
   	} else {
    		if ($form_provider) {
-    		$query .= "JOIN form_encounter AS e ON " .
-    			"e.pid = p.pid AND e.provider_id = '$form_provider' ";
+    		$query .= "JOIN libreehr_postcalendar_events AS e ON " .
+    			"e.pc_pid = p.pid AND e.pc_aid = '$form_provider' ";
    		} else {
-    		$query .= "LEFT OUTER JOIN form_encounter AS e ON " .
-    			"e.pid = p.pid ";
+    		$query .= "LEFT OUTER JOIN libreehr_postcalendar_events AS e ON " .
```
muarachmann: Will test this when i can. Can you repopulate back the fields with the search criteria? e.g if you chosed  a date it should fill back after the search is done.

mhn94: Thank you, and I can do that. I will push those changes when I'm done.

mhn94:
1) Previously, the prepareAndShowResults() function would generate an invalid date range if no date range or an incomplete date range was entered. For example, if a user entered the to date as 07-01-2018, the function would have generated 01-01-2019 as the from date, because it generated a date based on the present year the form was submitted. Now, it generates a date based on the input submitted. Unfortunately, this doesn't prevent users from entering an invalid date range. This can be created into an issue report.

2) The form now shows the date range entered after each submission.
date-range-now-shows
Since this function is also used by other reports that have a date range, it will now display correctly in those reports as well. :)

mhn94:
I have created an issue report regarding the form accepting invalid date ranges: [#1499 ](https://github.com/LibreHealthIO/lh-ehr/issues/1499)

library/report_functions.php
```diff
@@ -90,25 +90,31 @@ function dropDownPayors() {
...
+
     echo "<td class='label'>";
     echo xlt('From');
     echo ":
         </td>
          <td>
             <input type='text' name='form_from_date' id='form_from_date' size='10'
-                value='"; htmlspecialchars(oeFormatShortDate(attr($from_date)));
-    echo "' />
+                value='"; echo $from_date;
```
muarachmann: Hmmm something seems not right here ....

interface/reports/reports_controllers/PatientListController.php
```diff
@@ -33,45 +33,45 @@ function qescape($str) {
...
 	   "p.fname, p.mname, p.lname, p.street, p.city, p.state, " .
 	   "p.postal_code, p.phone_home, p.phone_biz, p.pid, " .
-	   "count(e.date) AS ecount, max(e.date) AS edate, " .
+	   "count(e.pc_eventDate) AS ecount, max(e.pc_eventDate) AS edate, " .
 	   "i1.date AS idate1, i2.date AS idate2, " .
 	   "c1.name AS cname1, c2.name AS cname2 " .
 	   "FROM patient_data AS p ";
 
   	if (!empty($from_date)) {
-   		$query .= "JOIN form_encounter AS e ON " .
```
muarachmann: Y did you delete the query? any specific reasons @mhn94, I think it did the Job.

muarachmann:
Well without wanting to spoil anything or take any credits, she could get
back on this on October 1st.

Let me know if it is urgent

KoniKodes:
It's more frustrating than anything right now. We can wait until she's back - hope she's accepted!
The Education Assessment forms are more important at the moment, I think.

interface/reports/reports_controllers/PatientListController.php
mhn94: > Y did you delete the query? any specific reasons @mhn94, I think it did the Job.

I deleted it because it wasn't displaying the right data when I ran the app on my machine. Most of the time, it wouldn't pull any appointment info when there should have been. Does the original query work for you?

I changed the db to  "libreehr_postcalendar_events," because I saw that when new events were created/old events were edited (in /modules/calendar/add_edit_event.php), this db was used. For me, the new query seems to be displaying the correct info.

library/report_functions.php
mhn94: I believe I fixed this. I just pushed a change!

KoniKodes:
@muarachmann @Ngai-E Can you check this please? 
Thank you.

PR #1498, opened 2019-08-19
contributed forms

Ngai-E:


Ngai-E:
> Not sure if the sqlconfig file is to be deleted. Can you remove this and commit again

I have tried removing that file from the commit but only ended up deleting it. Can you help me out?

muarachmann:
> > Not sure if the sqlconfig file is to be deleted. Can you remove this and commit again
> 
> I have tried removing that file from the commit but only ended up deleting it. Can you help me out?

Okay it is either you create back the file with original contents from github repo and remove that file from being tracked by git `git rm --cached pat/to/sqlconfig.php`.  I think henceforth it is being addressed here #1388.  Then force push again

Ngai-E:
> > > Not sure if the sqlconfig file is to be deleted. Can you remove this and commit again
> > 
> > 
> > I have tried removing that file from the commit but only ended up deleting it. Can you help me out?
> 
> Okay it is either you create back the file with original contents from github repo and remove that file from being tracked by git `git rm --cached path/to/sqlconfig.php`. I think henceforth it is being addressed here #1388. Then force push again

Thanks

Issue #1531, opened 2019-10-13
Duplicate Modules Open in the Same Screen


mhn94:
A) **Outreachy Username:** maggie.negm
B) **Issue Title:** Duplicate Modules Open in the Same Screen
C) **Report Date:** 10/13/19
D) **Sites Affected:** Documentation, NHANES
E) **OS/Browser Used:** Chrome browser in Linux OS
F) **EHR Workflow Module:** Calendar, Flow Board, and other default module options (except Messages)
G) **Steps to Reproduce the Issue:**
It depends on which modules open by default once a user logs in. In this example, let’s assume the “Calendar” and “Flow Board” modules are already opened by default.

1. Select “Calendar” from the main menu.
2. Select “Flow Board” from the main menu.
       
H) **Expected Behavior:** After trying to open another instance of an already opened module, the already opened module should stay open, but refresh itself.
I) **What Actually Happened:** A duplicate module opens. (However, if you try to open the module again, the second window refreshes itself.)
J) **Screenshots with Explanation:**
In this example, the default modules that are opened are Calendar and Flow Board.
![open-default-modules](https://user-images.githubusercontent.com/8771586/66723412-eb73b300-ede6-11e9-8faa-04ad155b6271.png)

1. Select “Calendar” from the main menu.
![main-menu-calendar-outline](https://user-images.githubusercontent.com/8771586/66723424-02b2a080-ede7-11e9-820c-765a02414dc8.png)
Another Calendar module opens.
![duplicate-modules-1-w-outline](https://user-images.githubusercontent.com/8771586/66723432-19f18e00-ede7-11e9-8e1c-6095b55001d1.png)

2. Select “Flow Board” from the main menu.
![main-menu-flowboard-outline](https://user-images.githubusercontent.com/8771586/66723435-2a096d80-ede7-11e9-9c51-2d18c06db2b8.png)
Another Flow Board module opens.
![duplicate-modules-2-w-outline](https://user-images.githubusercontent.com/8771586/66723440-355c9900-ede7-11e9-8082-d12c4a15115d.png)

The following is the associated code within the Chrome Inspector:
![chrome-inspector-code-w-outline](https://user-images.githubusercontent.com/8771586/66723446-43121e80-ede7-11e9-8931-6f4d0a6dcf05.png)

K) **Estimation of Severity:** Critical – module function continues with ongoing errors
[...]

PR #1535, opened 2019-10-31
Fixes Issue #1531 - can no longer open duplicate modules on same screen

mhn94:
I decided to go ahead and push this solution until we can further discuss if there needs to be certain modules that have the option and capability to be opened in two tabs.

Right now, none of the default tabs can be opened more than once. If the user tries to open a default module that is already opened, the same tab will reload.

The root cause was that the navigateTab() function that determines if a tab is already opened was not able to correctly identify the default tabs. The default tab URL and names are hard coded in, and after all of the database info initially loads, the URL and name are supposed to both change to reflect the chosen modules in the global settings. However, the URL changed, but not the name.

My solution correctly changes the associated name with each default module URL when the default tabs are initially loaded.

### DIFF
--- a/interface/main/tabs/main.php
+++ b/interface/main/tabs/main.php
@@ -79,7 +79,16 @@ var webroot_url="<?php echo $web_root; ?>";
 <?php require_once("menu/menu_json.php"); ?>
 <?php $userQuery = sqlQuery("select * from users where username='".$_SESSION{"authUser"}."'"); ?>
 <script type="text/javascript">
-    <?php if(isset($_REQUEST['url']))
+    <?php 
+        $default_tabs_info = array(
+            "/interface/main/main_info.php" => "cal",
+            "/interface/main/finder/dynamic_finder.php" => "fin",
+            "/interface/new/new_comprehensive.php" => "pat",
+            "/interface/patient_tracker/patient_tracker.php?skip_timeout_reset=1" => "flb",
+            "/interface/main/messages/messages.php?form_active=1" => "pat"
+        );
+        
+        if(isset($_REQUEST['url']))
         {
 
             $tab_one_default=$web_root.$GLOBALS['default_tab_1'];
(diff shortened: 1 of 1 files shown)


KoniKodes:
@muarachmann @VirginiaBalseiro @Ngai-E 
Can we work on this please?

Ngai-E:
looking into this

Ngai-E:
Ok it works fine. But is it not possible the use might actually want another instance of same tab opened? @muarachmann  @KoniKodes .

interface/main/tabs/main.php
```diff
@@ -97,13 +106,15 @@ function isEncounterLocked( encounterId ) {
...
             app_view_model.application_data.tabs.tabsList()[0].url(<?php echo $tab_one_contents; ?>);
+            app_view_model.application_data.tabs.tabsList()[0].name(<?php echo json_encode($default_tabs_info[$GLOBALS['default_tab_1']]); ?>);
         <?php
         }
         if(isset($GLOBALS['default_tab_2']))
         {
         ?>
             tab_defaults[1]=<?php echo json_encode($web_root.$GLOBALS['default_tab_2']); ?>;
             app_view_model.application_data.tabs.tabsList()[1].url(tab_defaults[1]);
+            app_view_model.application_data.tabs.tabsList()[1].name(<?php echo json_encode($default_tabs_info[$GLOBALS['default_tab_2']]); ?>);
```
muarachmann: Testing this ...

mhn94:
The original issue targets the default tabs, since I noticed those were opening twice. This PR fixes that issue.

The default tabs are the two tabs that automatically load when you login. So if the two default tabs were Flow Board and Calendar, you could open a second tab of the Flow Board or Calendar. However, I noticed that almost all/all of the other tabs don’t have this behavior, which is why I created an issue for it.

Although, if you take a look at Issue 1531, Art mentioned that sometimes you may want duplicate tabs in some cases. If so, I think that would mean that feature would have to be further implemented.

Ngai-E:
> The original issue targets the default tabs, since I noticed those were opening twice. This PR fixes that issue.
> 
> The default tabs are the two tabs that automatically load when you login. For example, if the two default tabs were Flow Board and Calendar, you could open a second tab of the Flow Board or Calendar. However, I noticed that almost all/all of the other tabs don’t have this behavior, which is why I created an issue for it.
> 
> Although, if you take a look at Issue 1531, Art mentioned that sometimes you may want duplicate tabs in some cases. If so, I think that would mean that feature would have to be further discussed and implemented.
> 
> Please let me know if I'm misunderstanding something :)

ok.

KoniKodes:
@robbyoconnor Can we please have a second approval and merge for this?

muarachmann:
@KoniKodes Merged

KoniKodes:
Thank you @muarachmann

PR #1539, opened 2019-11-02
Fix Issue 1537

yamamashakaa:
After using Export to CSV button  The Submit Button in Patient_list.php behaves like the Export to CSV Button and don't submit values.
Solution: Correct the condition in the if statement and set the value of
form_csv to false when clicking submit.
Issue: 1537

### DIFF
--- a/interface/reports/patient_list.php
+++ b/interface/reports/patient_list.php
@@ -31,10 +31,9 @@
 require_once "reports_controllers/PatientListController.php";
 
 // In the case of CSV export only, a download will be forced.
-if ($_POST['form_csvexport'] && !$_POST['form_refresh']) {
-  csvexport('patient_list'); // CSV headers. (TRK)
+if ($_POST['form_csvexport']) {
+    csvexport('patient_list'); // CSV headers. (TRK)
 } else {
-    $_POST['form_csvexport']=false;
 ?>
 <html>
 <head>
--- a/interface/reports/patient_list.php
+++ b/interface/reports/patient_list.php
@@ -31,9 +31,10 @@
 require_once "reports_controllers/PatientListController.php";
 
 // In the case of CSV export only, a download will be forced.
-if ($_POST['form_csvexport']) {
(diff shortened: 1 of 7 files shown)


Ngai-E:
This works fine thanks.
One suggestion try creating a separate branch when working on a fix or feature then send a PR for it other than messing up with you masters branch.

library/report_functions.php
```diff
@@ -145,7 +145,7 @@ function showSubmitPrintButtons($form_csvexport='') {
                     <div style='margin-left:5px'>
                         <a href='#' class='css_button cp-submit' onclick='$(\"#form_refresh\").attr(\"value\",\"true\"); $(\"#theform\").submit();'>
                             <span>";
-                                echo xlt('Submit');
+                         echo xlt('Submit');
```
muarachmann: Hmmmm should this be here?

interface/usergroup/addrbook_edit.php
```diff
@@ -252,7 +252,7 @@ function typeSelect(a) {
     invalue('form_notes')         . ", " .
     invalue('form_abook_type')    . " "  .
    ")");
-
+
```
muarachmann: Well surely the editor doing stuffs

interface/reports/patient_list.php
```diff
@@ -31,9 +31,10 @@
 require_once "reports_controllers/PatientListController.php";
 
 // In the case of CSV export only, a download will be forced.
-if ($_POST['form_csvexport']) {
-  csvexport('patient_list'); // CSV headers. (TRK)
+if ($_POST['form_csvexport'] && !$_POST['form_refresh']) {
+    csvexport('patient_list'); // CSV headers. (TRK)
 } else {
+     $_POST['form_csvexport']=false;
```
muarachmann: LGTM

yamamashakaa:
Happy that it got merged. Thanks for the advise, yes next time I will create a branch isntead of playing with the master

PR #1540, opened 2019-11-04
Education form

Ngai-E:
Ok i have redesigned this @muarachmann  hope this is better.

PR #1542, opened 2019-11-08
patient care plan

Ngai-E:


muarachmann:
I guess this is in #1540 cause i did test, anything new? Here is what I have
Screen Shot 2019-12-14 at 6 15 51 PM

Ngai-E:
yes I did not notice

> I guess this is in #1540 cause i did test, anything new? Here is what I have
> Screen Shot 2019-12-14 at 6 15 51 PM

yes I did not notice

Issue #1525, opened 2019-10-08
Duplicate "Test" on Flow Board Category


HenriettaWilla:
A) Outreachy Username: NDum.
B) Issue Title: Duplicate "Test" on Flow Board "Category"
C) Report Date: 10/06/19.
D) Sites Affected: Documentation Site.
E) OS/Browser Used: Chrome browser in Linux OS.
F) EHR Workflow Module: Duplicate "Test" on Flow Board Category. When I drop it down, I see Test twice. On the console the line affected is “option value = 16”.
G) Steps to Reproduce the Issue:
Login with your account details.
Click on the menu group “Flow Board”.
Click on the select element corresponding to “Category”.
View the duplicate option “Test”.
H) Expected Behaviour:
I expected to see the option “Test” just once.
 I) What Actually Happened: The option “Test” was duplicated.
J) Screenshots with Explanation:
![image](https://user-images.githubusercontent.com/38100233/66437109-d41c7a80-ea20-11e9-8eba-80f26ed09115.png)
Figure 1

![image](https://user-images.githubusercontent.com/38100233/66437130-e26a9680-ea20-11e9-9d03-f13b44f40a9b.png)
Figure 2

Figure 1 above shows the duplicate option "Test" under Category
Figure 2 shows the chrome inspector indicating where the issue is coming from.
K) Estimation of Severity: Minor: distracting or confusing but workflow can continue correctly.

L) Solution: The solution is to remove the duplicated option.

KoniKodes:
@HenriettaWilla 
In section G and H:
"Text" should be "Test". Just as it is seen in the image. Please change this report to reflect that and then create the PR.

HenriettaWilla:
ok thanks @KoniKodes. I tried to rectify the issue but the clone repository does not have such error

HenriettaWilla:
The categories were different from the one online

muarachmann:
status??

PR #1545, opened 2019-12-18
added check for duplicate category names

mhn94:
This is the beginning attempt to create a solution for Issue #1525.

I was able to delete the duplicate "Test" category from the Category List through the Calendar Admin Tool. (I actually deleted both, since I originally created the "Test" category and it's not needed anymore.)

What the category list looks like now within the docs server app:
result-after-deletion-w-outline

Now, I'm trying to solve the underlying issue: the Calendar Admin Tool allowing category names to be reused. The category names should be unique values. My solution is to have a check when the form is submitted to check if there is a category with that name that already exists. If so, display a message to users using iziModal and keeping the entered form info displayed by not refreshing the form. Otherwise, the new category is created.

I created the check and message, but I am receiving a console error when I try to display a message using iziModals library:
console-error

I'm also having trouble with displaying the entered info after the user receives the message that they cannot use a category name. Instead, the form refreshes and all the entered info is deleted.

### DIFF
--- a/modules/calendar/admin.php
+++ b/modules/calendar/admin.php
@@ -7,22 +7,14 @@ require_once("includes/admin_helper.php");
 call_required_libraries(array("jquery-min-3-1-1", "bootstrap", "iziModalToast"));
 
 // if update category button is used
+$nameCheck = false;
 if($_POST['updateCat'] == 1) {
   // if this is a new category, check if there exists a category already with the same name
-  $nameCheck = false;
   $categories = getCategories(); // retrieve up-to-date list of categories from admin_helper
   if ($_SESSION['category'] == "__NEW__") {
     foreach ($categories as $category) {
       if ($category['pc_catname'] == $_POST['catName']) {
         $nameCheck = true;
-        echo ('
-        <script type="text/javascript">
-            iziToast.warning({
-              title: "Caution:",
-              message: "This category name already exists. You must enter a new category name.",
-            });
-        </script>
-        ');
         break;
       }
     }
@@ -117,7 +109,8 @@ if(isset($_SESSION['category']) && $_SESSION['category']!=NULL) {
             <label class="control-label col-md-2" for="catName"><?php echo xlt('Name');?></label>
             <div class="col-md-10">
               <input type="text" class="form-control" id="catName" name="catName" placeholder="<?php echo xlt('Category Name');?>" required 
-              <?php if(!empty($selectedCat)) echo " value=\"" . $selectedCat['pc_catname'] . "\"";  ?> >
+              <?php if(!empty($selectedCat)) echo " value=\"" . $selectedCat['pc_catname'] . "\"";
+              else if ($nameCheck) echo " value=\"" . $_POST['catName'] . "\""; ?> >
             </div>
           </div>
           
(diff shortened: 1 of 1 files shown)


modules/calendar/admin.php
```diff
@@ -4,15 +4,35 @@
...
+call_required_libraries(array("jquery-min-3-1-1", "bootstrap", "iziModalToast"));
 
 // if update category button is used
 if($_POST['updateCat'] == 1) {
-  // call for creation/updation of category
-  createUpdateCategory($_SESSION['category'], $_POST['catName'], $_POST['catCol'], 
-    $_POST['catDes'], $_POST['catType'], $_POST['catDur'], 
-    $_POST['catAllDay'], $_SESSION['category'] == "__NEW__");
+  // if this is a new category, check if there exists a category already with the same name
+  $nameCheck = false;
```
muarachmann: This can get outside and in turn called at the bottom end
`$nameCheck = false`

modules/calendar/admin.php
```diff
@@ -4,15 +4,35 @@
...
+  $categories = getCategories(); // retrieve up-to-date list of categories from admin_helper
+  if ($_SESSION['category'] == "__NEW__") {
+    foreach ($categories as $category) {
+      if ($category['pc_catname'] == $_POST['catName']) {
+        $nameCheck = true;
+        echo ('
+        <script type="text/javascript">
```
muarachmann: This wont work call it at the bottom of the script so that it can affect the document
```
if(namecheck){
echo
}
```

muarachmann: You might want to alert this at the bottom of the script rather see https://github.com/LibreHealthIO/lh-ehr/blob/master/interface/usergroup/facilities_add.php It should help

mhn94:
Thank you for your feedback, @muarachmann!

mhn94:
Hey @muarachmann,

So I've incorporated your feedback into my solution. When a user enters a category name that already exists, a message pops up and the form displays previously entered info. I also made sure that no negative numbers or zero can be entered as a duration value.

A concern I have with my solution is that the message doesn't pop up instantaneously after submission and the form appears to be a bit laggy. I would appreciate any input in regards to this, if you have any.

modules/calendar/admin.php
```diff
@@ -7,22 +7,14 @@
...
         $nameCheck = true;
-        echo ('
-        <script type="text/javascript">
-            iziToast.warning({
-              title: "Caution:",
-              message: "This category name already exists. You must enter a new category name.",
-            });
-        </script>
-        ');
         break;
```
muarachmann: LGTM testing in a while

muarachmann:
> Hey @muarachmann,
> 
> So I've incorporated your feedback into my solution. When a user enters a category name that already exists, a message pops up and the form displays previously entered info. I also made sure that no negative numbers or zero can be entered as a duration value.
> 
> A concern I have with my solution is that the message doesn't pop up instantaneously after submission and the form appears to be a bit laggy. I would appreciate any input in regards to this, if you have any.

Lagging could be maybe your server's response time, I will simulate this

muarachmann:
@mhn94 works fine on my end. I just tested your PR and response too acts quickly. Maybe you had many things running on your desktop juts try to run this only and it should be fine. Nice work :)

mhn94:
@muarachmann yay, good to hear! :)

Ngai-E:
works fine. Just that I almost missed the pop-up warning since it appeared way at the bottom. 
image

mhn94:
Thanks @Ngai-E! I've made the changes you've requested. I chose to display the iziToast notification in the top center of the module rather than center, so it doesn't block the form if the user decides to make changes before the notification disappears.

modules/calendar/admin.php
```diff
@@ -229,9 +229,10 @@
       if($nameCheck) {
         echo ('
           <script type="text/javascript">
-            iziToast.warning({
+            iziToast.error({
               title: "Caution:",
               message: "This category name already exists. You must enter a new category name.",
+              position: "topCenter"
```
Ngai-E: perfect!

Issue #1536, opened 2019-11-01
White space/block present in the top right corner 


j-pro:
**A) Outreachy Username:** j-pro
**B) Issue title:** White Block present in the top right corner 
**C) Report Date:** Oct 31, 2019 
**D) Sites Affected:** Documentation Site
**E) OS/ browser used:** Windows / Chrome
**F) EHR Workflow module:** Not specifically a module, as it is present in main menu bar

**G) Steps to Reproduce the Issue:** 
1.	Login with your account credentials.
2.	In the upper right corner, a white block is seen above the username.

**H) Expected Behavior:** The white block should not have appeared.

**I) What Actually Happened:** White block appears in the top right corner, next to the ‘Help’ menu. 

**J) Screenshots with Explanation:** 

![ws-01](https://user-images.githubusercontent.com/1905329/68007615-3aa15b00-fca4-11e9-9d46-2e351f4226a5.png)
 Two white spaces appear in the top right corner, just above username.  As it is present in the main menu, it is seen in all the modules.

![ws-02](https://user-images.githubusercontent.com/1905329/68007630-468d1d00-fca4-11e9-9c2c-c543976d5bbd.png)
In the smaller screen, the space appears to be present in both sides of the username. It seems to be present from the ‘nav’ class.

**K) Estimation of Severity:**
Trivial: cosmetic defect only; has no effect on Workflow or Module's actual function.

**L) Solution:** 
As the list items are limited, there is extra space in the navigation bar.  The same space has appeared as white background.  It can be eliminated by setting the background color of navigation bar as same as the present color of the header.
 
![ws-03](https://user-images.githubusercontent.com/1905329/68007663-5b69b080-fca4-11e9-980f-4cd587c17d68.png)
I have added ‘background-color: #fedf98; in the ‘nav’ class.  It seems to fix the issue.

j-pro:
The file I am running in local server has white background color in the menu bar. This is screenshot of my localhost:
![issue_white_bg](https://user-images.githubusercontent.com/1905329/68214878-59865100-0006-11ea-8c38-5584e560bb4d.png)
This issue #1536 is present in the docs site running in main server. 
I find difference in these two.

mhn94:
I've replicated the template settings (color theme and background colors) within my local app to mirror the app on the Docs server, and this issue doesn't exist in my local app, as shown below:
[...]

PR #1547, opened 2020-01-07
removed white spaces in menu header (fixes Issue #1536)

mhn94:
The white spaces (which is actually the primary color chosen in the Global settings) in the menu header and around the user's name have been removed and no longer show no matter the screen size. It now looks like this:

Full Screen View
uniform-color-full-screen

Less than Full Screen but Menu isn't Hidden View
uniform-color-not-full-view

Hidden Menu View
uniform-color-mobile-view

### DIFF
--- a/interface/main/tabs/main.php
+++ b/interface/main/tabs/main.php
@@ -148,7 +148,7 @@ var webroot_url="<?php echo $web_root; ?>";
 </style>
 <div id="mainBox">
     <div id="dialogDiv"></div>
-    <div class="navbar-header">
+    <div class="navbar-header" id="navbar-header">
         <button type="button" class="navbar-toggle" data-toggle="collapse" data-target="#navbar-collapse">
         <span class="sr-only"><?php echo xlt("Toggle navigation"); ?></span>
         <span class="icon-bar"></span>
--- a/interface/themes/style.php
+++ b/interface/themes/style.php
@@ -64,10 +64,9 @@ else {
 	$button_font_color = $secondary_font_color;
 }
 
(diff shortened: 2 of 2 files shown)


mhn94:
I may have found a bug while working on this, but would like it if someone else could confirm it's a bug.

I believe that the primary color chosen in the Global Settings (Administration > Globals > Appearance tab > Primary color) isn't showing properly in the menu header. The Global Settings allow you to choose a different color than the default one (the tan color above - #fedf98), but it doesn't show in the app. 

I believe it's because when the app is first loaded, there's a call to the style.php file with preloaded color choices, as shown below:
style-php-call-outline
However, I was unsuccessful in locating the original call to style.php. The inspector says it came form main.php, but I can't find the call to style.php in there. So, I'm stuck and would appreciate any help!

KoniKodes:
@muarachmann @Ngai-E @VirginiaBalseiro Can someone help Maggie track down this issue?

mhn94:
> I may have found a bug while working on this, but would like it if someone else could confirm it's a bug.
> 
> I believe that the primary color chosen in the Global Settings (Administration > Globals > Appearance tab > Primary color) isn't showing properly in the menu header. The Global Settings allow you to choose a different color than the default one (the tan color above - #fedf98), but it doesn't show in the app.
> 
> I believe it's because when the app is first loaded, there's a call to the style.php file with preloaded color choices, as shown below:
> style-php-call-outline
> However, I was unsuccessful in locating the original call to style.php. The inspector says it came form main.php, but I can't find the call to style.php in there. So, I'm stuck and would appreciate any help!

I've since then figured out the user preferences takes priority over the global settings, so this is no longer an issue.

KoniKodes:
Then do we want to close this?

mhn94:
Not yet! The original issue's PR solution stills needs to be checked.

KoniKodes:
Oh. Alright.

PR #1551, opened 2020-01-24
Update README.md

aashimgarg:
These guidelines will make the user more comfortable for contributing.


robbyoconnor:
I'm not sure about this. Since the source repo has been deleted, no changes can be requested.

Issue #1550, opened 2020-01-24
Patient List Creation Reports Don't Filter


mhn94:
A) **Outreachy Username:** maggie.negm
B) **Issue Title:** Patient List Creation Filtering Incorrectly
C) **Report Date:** 01/21/2020
D) **Sites Affected:** Docs Site, NHANES
E) **OS/Browser Used:** Chrome browser in Linux OS
F) **EHR Workflow Module:** Report – Patient List Creation
G) **Steps to Reproduce the Issue:**

1. Select “Reports” from the main menu.
2. Select “Clients” under “Reports.”
3. Select “Patient List Creation” under “Clients.”
4. Fill out and submit the form.

H) **Expected Behavior:** The correct results should be returned.
I) **What Actually Happened:** Depending on the input entered, especially the “Option” chosen, the form isn’t returning the correct results.

- Demographics option: returns the same list
- Problems option: doesn’t return anything
- Medications option: doesn’t return anything
- Allergies option: doesn’t return anything
- Lab Results option: doesn’t return anything
- Communication option: returns the same list

J) **Screenshots with Explanation:**

- Example 1:
**Input:**
From: 01/01/2019
To: 08/01/2019
Option: Demographics
Gender: Female
**Output:**
Total Patients: 104
![filtering1](https://user-images.githubusercontent.com/8771586/73092179-18396b00-3eaa-11ea-9369-f691530e69fe.png)

- Example 2
**Input:**
Option: Demographics
Patient ID: 123456
Gender: Male
**Output:**
Total Patients: 104
![filtering2](https://user-images.githubusercontent.com/8771586/73092203-24bdc380-3eaa-11ea-83a2-748651605e7f.png)

- Example 3
**Input:**
Option: Allergies
**Output:**
No records found.
![filtering-3](https://user-images.githubusercontent.com/8771586/73092220-2d15fe80-3eaa-11ea-8b78-d4ecdf0a77d9.png)

The following is the code that displays the form within Chrome Inspector:
![patientlistcreationinspector](https://user-images.githubusercontent.com/8771586/73092245-3a32ed80-3eaa-11ea-9373-38762192fb6e.png)

K) **Estimation of Severity:** Critical – module function continues with ongoing errors
M) **Solution:** I would start by looking at the filtering function and the database queries being called when the form is submitted within the PatientListController.php file.

Similar Issue: #1496

PR #1554, opened 2020-02-14
Patient List Creation report now filters correctly (Fixes Issue #1550)

mhn94:
The "Patient List Creation" Reports module now filters results correctly based on input. 

- Form input is now accessible within the prepareAndShowResults() function through the usage of the keyword "global." 

- The "From" and "To" dates are now queried properly based on the input.


mhn94:
I added a new column called "Diagnosis Status" when users choose to display results concerning patient medications (Option > Medications), per Harley's request. It shows whether the diagnosis status is active, inactive, or resolved. 

new-col-diagnosis-status

muarachmann:
I think I need data to test this ... Looking for my old backups in my old mac ..... wheeeepppp

PR #1567, opened 2020-03-14
'Find Available' button no longer outputs an infinite loop of appt times

### LINKED ISSUE #1566: 'Find Available' button outputs an infinite loop of appt times
A) **Outreachy Username:** maggie.negm
B) **Issue Title:** Find Available Calendar Bug
C) **Report Date:** 03/14/2020
D) **Sites Affected:** Docs Site, NHANES
E) **OS/Browser Used:** Chrome browser in Linux OS
F) **EHR Workflow Module:** Calendar
G) **Steps to Reproduce the Issue:**

1. Open the Calendar module if it isn't already opened.
2. Click in a column under a provider's name.
3. Select the "Provider" tab.
4. Select the "Find Available" button in the bottom left corner.

H) **Expected Behavior:** Available provider times should be displayed.
I) **What Actually Happened:** An infinite loop of the time "12:00" appears and the app eventually crashes.
J) **Screenshots with Explanation:**
Below, the image shows the infinite loop of times that is displayed when the "Find Available" is selected in the "Provider" tab.
infinite-times

K) **Estimation of Severity:** Blocker - module's function ceases at this point or crashes/disables the EHR itself
M) **Solution:** The root cause of this infinite loop is when a negative or zero value of "evdur" is included in the dialog URL that displays the available times in either the "Patient" or "Provider" tab. Thus, to fix this, the "evdur" value will only be included in the URL if it's positive.

mhn94:
Fixes #1566 

Now, both "Patient" and "Provider" tabs no longer show an infinite loop of times. Instead, the correct available times are shown or a message is displayed letting the user know no there are no available times.
no-infinite-loop

### DIFF
--- a/modules/calendar/add_edit_event.php
+++ b/modules/calendar/add_edit_event.php
@@ -1297,8 +1297,12 @@ var weekDays = new Array(
             f = document.forms[0].facility.options[f.selectedIndex].value;
         <?php }?>
         var c = document.forms[0].form_category;
-    var formDate = document.forms[0].form_date;
-        dlgopen('<?php echo $GLOBALS["web_root"]; ?>/modules/calendar/find_appt_popup.php' +
+        var formDate = document.forms[0].form_date;
+
+        // evdur must be a positive number; if evdur <= 0, it will output an infinite stream of times
+        // If evdur is positive, include it in URL.
+        if (document.forms[0].form_duration.value > 0) {
+            dlgopen('<?php echo $GLOBALS["web_root"]; ?>/modules/calendar/find_appt_popup.php' +
                 '?providerid=' + s +
                 '&catid=' + c.options[c.selectedIndex].value +
                 '&facility=' + f +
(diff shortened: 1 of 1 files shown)


muarachmann:
I still need to test this.

muarachmann:
Boom... I tried with the Nhanes data and yes I had this infinite loop. Now I am not sure if the data is corrupt or not consistent. I am applying your patch now ....

muarachmann:
As you see here is a query

```
string(330) "SELECT pc_eventDate, pc_endDate, 
pc_startTime, pc_duration, pc_recurrtype,
pc_recurrspec, pc_alldayevent, pc_catid, 
pc_prefcatid, pc_apptstatus FROM 
libreehr_postcalendar_events WHERE pc_aid = ? AND pc_eid != ? 
AND ((pc_endDate >= ? AND pc_eventDate < ? ) 
OR (pc_endDate = '0000-00-00' AND pc_eventDate >= ? AND pc_eventDate < ?))"
```

Looking at the imported db I had a bunch of pc_endDate with `0000-00-00`
Screen Shot 2020-03-25 at 8 31 26 AM

So I am tempted to say a better solution will be to validate the date before storing in future. This works though since it restricts those

modules/calendar/add_edit_event.php
```diff
@@ -1307,6 +1311,16 @@ function find_available(extra) {
                 '&eid=<?php echo 0 + $eid; ?>' +
                 extra,
                 '_blank', 500, 400);
+        } else { // If evdur is zero or negative, do not include it in URL.
+            dlgopen('<?php echo $GLOBALS["web_root"]; ?>/modules/calendar/find_appt_popup.php' +
+                '?providerid=' + s +
+                '&catid=' + c.options[c.selectedIndex].value +
+                '&facility=' + f +
+                '&startdate=' + formDate.value +
```
muarachmann: I should guess this is the breaker :0

PR #1569, opened 2020-03-17
modified grant_priviledges function for mysql8

MathurinNkenfack:
In mysql8 you need to create a user first before granting it privileges and I had to use mysql_native-password for it to connect to the database

### DIFF
--- a/library/classes/Installer.class.php
+++ b/library/classes/Installer.class.php
@@ -146,7 +146,16 @@ class Installer
   }
 
   public function grant_privileges() {
+	$version =  mysqli_get_server_version($this->dbh);
+	if($version >= 80000){
+		$this->execute_sql("drop user $this->login@$this->loginhost");
+		$this->execute_sql("FLUSH PRIVILEDGES");
+		return ( ($this->execute_sql("CREATE USER '$this->login'@'$this->loginhost' IDENTIFIED with mysql_native_password  by '$this->pass'"))
+		&&($this->execute_sql("GRANT ALL PRIVILEGES ON $this->dbname.* TO '$this->login'@'$this->loginhost'")));
+	}
+	else{
     return $this->execute_sql( "GRANT ALL PRIVILEGES ON $this->dbname.* TO '$this->login'@'$this->loginhost' IDENTIFIED BY '$this->pass'" );
+		}
   }
 
   public function disconnect() {


library/classes/Installer.class.php
```diff
@@ -146,7 +146,16 @@ public function drop_database() {
   }
 
   public function grant_privileges() {
+	$version =  mysqli_get_server_version($this->dbh);
+	if($version >= 80000){
+		$this->execute_sql("drop user $this->login@$this->loginhost");
+		$this->execute_sql("FLUSH PRIVILEDGES");
```
muarachmann: I need to spin up a MySQL-8 to test this. From first glance it might do the trick. Will test this in a while

robbyoconnor: Do it in Docker -- change to `mysql:5.7` to `mysql:8.0` -- just don't commit it

library/classes/Installer.class.php
```diff
@@ -146,7 +146,16 @@ public function drop_database() {
...
+		return ( ($this->execute_sql("CREATE USER '$this->login'@'$this->loginhost' IDENTIFIED with mysql_native_password  by '$this->pass'"))
+		&&($this->execute_sql("GRANT ALL PRIVILEGES ON $this->dbname.* TO '$this->login'@'$this->loginhost'")));
+	}
+	else{
     return $this->execute_sql( "GRANT ALL PRIVILEGES ON $this->dbname.* TO '$this->login'@'$this->loginhost' IDENTIFIED BY '$this->pass'" );
+		}
   }
```
robbyoconnor: This is vulnerable to sql injection...

library/classes/Installer.class.php
robbyoconnor: Nevermind, code needs to be reworked.

library/classes/Installer.class.php
muarachmann: @MathurinNkenfack address this properly and also consider trying this with docker by replacing the appropriate files with the necessary codes

muarachmann: If you need help feel free to reach out

MathurinNkenfack: I tried as much as could but to no avail that's why I instead used mysql 5 and secured the installing using prepared statements  in the following pull request #1586

PR #1570, opened 2020-03-17
Fix issue #1213

realJema:
This fixes the issue #1213


sites/default/sqlconf.php
```diff
@@ -1,26 +1,29 @@
...
 
 $host	= 'localhost';
 $port	= '3306';
 $login	= 'libreehr';
-$pass	= 'libreehr';
+$pass	= 'toor';
 $dbase	= 'libreehr';
 
+global $disable_utf8_flag;
+$disable_utf8_flag = false;
```
muarachmann: This is NOT suppose to be in this commit. Please remove this file from this commit

realJema: Done

realJema:
editing

PR #1574, opened 2020-03-18
Fixed URL for js modules

xurror:
closes https://github.com/LibreHealthIO/lh-ehr/issues/1553

### DIFF
--- a/interface/globals.php
+++ b/interface/globals.php
@@ -198,7 +198,7 @@ $javascript_dir = $GLOBALS['standard_js_path']; //Make path available as a varia
 $GLOBALS['current_version_js_path'] = "$web_root/assets/js/current_version";
 
 //module configurations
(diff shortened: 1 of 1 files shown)


interface/globals.php
```diff
@@ -198,7 +198,7 @@ function undoMagicQuotes($array, $topLevel=true) {
 $GLOBALS['current_version_js_path'] = "$web_root/assets/js/current_version";
 
 //module configurations
-$GLOBALS['modules_dir']  = "$webserver_root/modules/";  //CURRENT modules directory.
+$GLOBALS['modules_dir']  = "$webroot/modules/";        //CURRENT modules directory.
```
muarachmann: Perfect! ... sometimes we miss little things :). I am approving this works fine and LGTM!

robbyoconnor: You have trailing whitespace, could you clean that up, please?

robbyoconnor: Ingore previous comment

Issue #1340, opened 2018-12-07
Calendar Labels Overlap When User Has A Long Name


bauripalash:
## Problem : 
When viewing the calendar, if the user has a long name calender labels overlap

### Screenshots :
![screenshot from 2018-12-07 11-33-17](https://user-images.githubusercontent.com/16154204/49630454-41800900-fa14-11e8-9579-f01a31f326b6.png)

## Possible Fix : 
Create a line break in the cell if the user has a long name

aethelwulffe:
https://github.com/LibreHealthIO/lh-ehr/pull/1347#event-2197677059 was closed by punwai.  This is still very much an active issue, and a really bad one.

realJema:
I will like to propose a fix to this issue

muarachmann:
Testing this ....

muarachmann:
I think this is fixed and was address by @punwai , @realJema did you try to simulate this before sending a patch

PR #1575, opened 2020-03-18
Calendar Labels overlap when user has long name

realJema:
fix issue #1340

### DIFF
--- a/modules/calendar/css/index.css
+++ b/modules/calendar/css/index.css
@@ -81,6 +81,10 @@ body {
   color: #000;
 }
 
+.fc-resource-cell {
+  word-break: break-all;
+}
+
 .tooltipevent {
   padding:5px;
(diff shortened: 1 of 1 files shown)


muarachmann:
See here - https://github.com/punwai/lh-ehr/blob/5f4b96e15ed9efbbf8ab67d8b0806ad43d0542bb/modules/calendar/css/index.css#L84-L86

muarachmann:
Maybe it was never merged

muarachmann:
@realJema can you confirm this?

realJema:
> See here - https://github.com/punwai/lh-ehr/blob/5f4b96e15ed9efbbf8ab67d8b0806ad43d0542bb/modules/calendar/css/index.css#L84-L86

This fix was merged but I believe its inadequate because the problem will still occur for long wordss

PR #1579, opened 2020-03-20
White space/block present in the top right corner #1536

realJema:
Fix issue #1536

### DIFF
--- a/interface/themes/style.php
+++ b/interface/themes/style.php
@@ -65,14 +65,14 @@ else {
 }
 
 echo " .body_title, .body_top, .body_nav, .body_filler, .body_login, .table_bg, .bgcolor2, .textcolor1, .highlightcolor, .logobar, .dropdown-menu>li>a, .dropdown-toggle, #menu, .dropdown, .nav>li>a, .glyphicon, #userdata .dropdown-menu>li, #userdata{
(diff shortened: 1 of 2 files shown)


mhn94:
@realJema , taking a look at this. Do you mind attaching a pic as to what your screen looks like with this solution?

realJema:
> @realJema , taking a look at this. Do you mind attaching a pic as to what your screen looks like with this solution?


body background-color: green 
body_title background-color: transparent 
navbar-collapse background-color: transparent 
with fix 
Capture4


body background-color: green 
body_title background-color: blue
navbar-collapse background-color: white
without fix 
image

The previous fix just put everything with a background-color:white which is flawed in certain situations

realJema:
@mhn94 could you review and close this. Thanks

mhn94:
Hey @realJema . It seems like your fix isn't really targeting the issue at hand, which is to make the white block space blend in with the rest of the background color chosen. 

Your fix removes the background colors (i.e. the primary and secondary color settings) of the elements aka making those areas of the screen transparent.
transparent-UI

Rather than changing the background-color of those elements to "transparent," have you tried to see if the HTML element(s) in the white block is missing from that CSS statement block/ruleset?

Issue #1576, opened 2020-03-19
Text on ‘Cancel’ button in Edit 'Demographics' not visible.


ahanipeitra1994:
A) **Outreachy  Username**: peitra
B) **Issue Title**:  Text on ‘Cancel’ button in Edit Demographics not visible
C)  **Report Date:**   03/18/2020
D) **Sites Affected**:  Local Instance
E)  **OS/browser used**:  windows 10, Mozilla FireFox, Chrome
F)  **EHR Work Flow Module**:  Patient Finder
G)  **Steps to Reproduce the Issue:** 
1)  Hover over ‘Administration’ on the navigation bar.
2) On the dropdown click on ‘Globals’. 
3) A tap opens go to ‘Theme’ it has a dropdown then click on Tan.
4) Take ‘Save’ at the top left hand of the tap.
5) Hover over ‘Patient/Client’ on the navigation bar. 
6) On the dropdown click on ‘Finder.’
7) On the Patient Finder tap click on any patients name.
8) A new tap opens then click on ‘Edit Demographics.’
9) Click ‘Cancel’ and then click ‘Edit Demographics’ again.

H)  **Expected Behaviour**: The text on the ‘Cancel’ button should be readable. 
I)  What Actually Happened:  When the ‘Edit Demographics’ is clicked the text on the ‘Cancel’ button on the new tap is white and not visible.
 J)  **Screenshots  With Explanation.** From the screenshot below it  can easily be seen that  ‘Cancel’  button shows a white text.
![image](https://user-images.githubusercontent.com/60817596/77066851-3328e700-69e4-11ea-9268-fa7d364db9c0.png)

 
K)  **Estimation of Severity:**  Major,Module function continues with workaround.

L)  **Solution**: It can be corrected using CSS on the visited button to make the text visible as in the screenshot below. I changed the color of the visited button from white to brown.  
![image](https://user-images.githubusercontent.com/60817596/77067181-d7ab2900-69e4-11ea-80cf-a98a820d0bf9.png)

Ngai-E:
This is really killing us now. Please fix it.

PR #1581, opened 2020-03-21
Fixed text on button changing to white after clicking button

ahanipeitra1994:
@Ngai-E @mhn94  @muarachmann this is a fix for #1576. The text is now visible because I used a brown color. See screenshot below.
image

### DIFF
--- a/interface/themes/style_tan.css
+++ b/interface/themes/style_tan.css
@@ -673,7 +673,7 @@ tr.odd {
 }
 /* set the text color inside button to white when it is visisted or hovered */
 .css_button:visited{
-	color:white;
+	color:brown;
 }
 .css_button:hover{
 	color:white;


interface/themes/style_tan.css
```diff
@@ -673,7 +673,7 @@ tr.odd {
 }
 /* set the text color inside button to white when it is visisted or hovered */
 .css_button:visited{
-	color:white;
+	color:brown;
```
Ngai-E: great you did this in the theme file.

Ngai-E:
I think this was neatly done. Works fine for me @mhn94 @muarachmann  what do you think. I need another eye here.

interface/themes/style_tan.css
muarachmann: LGTM . I hope the new UI fixes all these

mhn94:
LGTM

PR #1586, opened 2020-03-24
Securing Installation from Sql injection

MathurinNkenfack:
1.preventing sql injection with prepared statements
2. change utf8 to utf8mb4

### DIFF
--- a/library/classes/Installer.class.php
+++ b/library/classes/Installer.class.php
@@ -1,461 +1,471 @@
 <?php
+
 /* Copyright © 2010 by Andrew Moore */
 /* Licensing information appears at the end of this file. */
 
 class Installer
 {
 
-  public function __construct( $cgi_variables )
-  {
-    // Installation variables
-    $this->iuser                    = $cgi_variables['iuser'];
-    $this->iuserpass                = $cgi_variables['iuserpass'];
-    $this->iuname                   = $cgi_variables['iuname'];
-    $this->iufname                  = $cgi_variables['iufname'];
-    $this->igroup                   = $cgi_variables['igroup'];
-    $this->server                   = $cgi_variables['server']; // mysql server (usually localhost)
-    $this->loginhost                = $cgi_variables['loginhost']; // php/apache server (usually localhost)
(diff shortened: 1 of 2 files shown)


muarachmann:
Give this a proper title ...

library/classes/Installer.class.php
```diff
@@ -473,304 +483,329 @@ public function write_configuration_file() {
...
+            $results = $this->dbh->query($sql);
+        }else{
+            $stmt = $this->dbh->prepare($sql);
+            if($count == 1){
+                $stmt->bind_param($str,$value[0]);
+            }
+            if($count == 2){
+                $stmt->bind_param($str,$value[0],$value[1]);
+            }
+            if($count == 3){
```
muarachmann: No this is baaaaaad. Try another approach how about escaping the query then binding the array in here. Try another alternative.

muarachmann: What if they were 20 params

MathurinNkenfack: had issues with knowing how many elements the array has to index it and as I saw the number number coming from the user did not exceed 5...

MathurinNkenfack: number of inputs*

MathurinNkenfack: php > 5 doesn't support variable decomposition also

MathurinNkenfack: if you have any clue...I will gladly like to hear

muarachmann:
We are not using php5 and we wont offer support for that. A for loop can
do. Try to research more on that and ideally, you should be
escaping the parameters too. Something like
```$this->mysql_escape($this->login)```.

BR,
Mua

library/classes/Installer.class.php
robbyoconnor: This is where you communicate and ask questions.

MathurinNkenfack: @robbyoconnor  thanks

MathurinNkenfack:
@muarachmann please can u review the changes please

Issue #1590, opened 2020-03-25
The “Add Role” form appears cramped up at top of modal.


ahanipeitra1994:
A) Outreachy  Username : peitra
B) Issue Title: The  “Add Role” form appears cramped up at top of modal
C)  Report Date:  03/20/2020
D) Sites Affected: Local  Instance
E)  OS/browser used:  Windows 10/ Mozilla Firefox
F)  EHR Work Flow Module: Administration
G)  Steps to Reproduce the Issue:
1)	Click on  “Administration “ on the main menu.
2)	Click on “ Manage Roles.”
3)	Click on “ Add Roles.”
4)	On the “Add Roles” tab click on the expand icon at the top right side of the tab.

H)  Expected Behavior: I expected to see a good view of the “Add Roles” form when I expand the modal.
I)  What Actually Happened: The “Add Role” form is too cramped up even when there is enough space in the modal. This makes it hard to manipulate the input fields.
 J)  Screenshot With Explanation. 
![image](https://user-images.githubusercontent.com/60817596/77533851-e9cf1080-6e97-11ea-89f9-080267ec761b.png)

 
  

K)  Estimation of Severity: Major, module function continues with work around.
L) Solution: I added height of 100% !important. It can be seen now that the modal is not cramped anymore.
![image](https://user-images.githubusercontent.com/60817596/77533645-8a710080-6e97-11ea-9640-ed8a72281b88.png)

Ngai-E:
waiting for this, please.

PR #1592, opened 2020-03-27
Fix modal height in firefox

ahanipeitra1994:
Hi this is fix of the issue #1590. I added a height of 100% important.


assets/css/iziModalToast/iziModal.css
```diff
@@ -244,6 +244,7 @@
     margin: 0 0 -6px 0;
     width: 100%;
     transition: height 0.3s ease;
+    height: 100% !important;
```
muarachmann: Lets give it `95%`

Ngai-E:
ok get the fix in

assets/css/iziModalToast/iziModal.css
ahanipeitra1994: Ok i will do that and see how it looks like thanks @muarachmann

ahanipeitra1994:
Muarachman I have changed the height to 95% .

muarachmann:
> Muarachman I have changed the height to 95% .

Thanks

PR #1600, opened 2020-03-31
Authenticate  File Write in letter

Falence:


robbyoconnor:
1. Handled in #1571 so this isn't needed
2. Your fix is incorrect and introduces a CSRF vuln. potentially.

Issue #1596, opened 2020-03-31
Add button won’t add options


laxmikandel:
A)	Outreachy Username: Laxmi

B)	Issue title: Add button won’t add options.   
C)	Report Date: 3/30/2020
D)	Sites Affected: Local Instance
E)	OS/ browser used: Windows 10, Mozilla Firefox.

F)	EHR Workflow module: Patient/Client

G)	Steps to Reproduce the Issue: 
1.	Hover over the “Patient/Client” on the navigation bar.
2.	Click on “Add New/Search.”
3.	Search existing patient.
4.	Click “Edit” on Demographic section. 
5.	Try to add option  through add buttons in the section named “Face Sheet”, “Contact” ,”Employer”

H)	Expected Behavior: It is expected to add if existing options button doesn’t have the required option while clicking on “add” buttons.
I)	What Actually Happened: There is no link pop up to add options.
J)	Screenshots with Explanation:   
This is from contact tab. 
 
![image](https://user-images.githubusercontent.com/22347000/78002384-e4217100-7356-11ea-8b73-b643a5f52a72.png)

This is from Employer tab.
![image](https://user-images.githubusercontent.com/22347000/78002038-504fa500-7356-11ea-8f04-4d1454e3e1c2.png)

 
This is from Face Sheet Tab. This is also not working.

![image](https://user-images.githubusercontent.com/22347000/78001670-c273ba00-7355-11ea-9caf-3d409a71a863.png)
K)	Estimation of Severity: 
a.	Major: Module function continues with workaround

L)	Workaround: If the system kept all possible options up to date, then we even no need to have that button. It will works fine. But in case of letting user to add them through “add” buttons then it must work. 

M)	Solution: Adding code function on add buttons can fix this issue. It can take maximum of two days.

Ngai-E:
Thanks can you fix it?

laxmikandel:
Yes

PR #1604, opened 2020-04-05
Fixed Add button won’t add options #1596

laxmikandel:
In interface/patient_file/summary/demographics_full.php file, jquery dependencies loaded before loading jquery and that dependencies is datetimepicker. That's why error was encounter. It is now fixed by placing them in correct sequence.


Ngai-E:
Ok I see. 
image
But I needed to refresh before seeing my option
image
Can you arrange this?

Issue #1594, opened 2020-03-30
 The Edit Amendments Tab Has No Back Button.


ahanipeitra1994:
A) **Outreachy**  Username: Peitra
B) **Issue Title:** The Edit Amendments Tab Has No Back Button
C)  **Report Date:**  03/29/2020
D) **Sites Affected**: Local Instance
E)  **OS/browser Used**: Windows 10/Mozilla FireFox
F)  **EHR Work Flow Module:** Patients/Client
G)  **Steps to Reproduce the Issue:**
1)	Hover over the  “Patient/Client” on the navigation bar.
2)	Click on “Finder.” 
3)	Click on a “patient’s” name.
4)	Click on “ Edit” button next to Amendments section.
5)	Click on the “List” button on the Amendments tab.

H)  **Expected Behavior**: I expected to see a back button to leave the tab.
I)  **What Actually Happened:**  When I click on “List”  in the Amendments tab, there is no back button to leave the tab. I only have to refresh the browser.
 J)  **Screen With Explanation**. From the screenshot below it can be seen that there is no back button on the tab.
![image](https://user-images.githubusercontent.com/60817596/77865020-13839100-7224-11ea-891e-60bad1d945fb.png)

 

  

K)  **Estimation of Severity:** Minor distraction or confusing but workflow can continue correctly.
L)  **Solution:** The  solution is to put a “Back” button on the Amendments tab.

Ngai-E:
I noticed even the Prescription tab lacks a back button, kindly do for both.

PR #1606, opened 2020-04-07
Add back to patient button on Prescriptions and Amendments

ahanipeitra1994:
I have added a "Back Button" on Amendments and the prescription tab. Issue #1594

### DIFF
--- a/interface/patient_file/summary/rx_left.php
+++ b/interface/patient_file/summary/rx_left.php
@@ -18,6 +18,13 @@ include_once("../../globals.php");
 <span><?php xl('Add','e');?></span></a>
 </td>
 </tr>
+<tr>
+		<td>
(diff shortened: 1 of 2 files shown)


maggienegm:
@ahanipeitra1994 the buttons' UI and positioning looks fine. 

What browser did you use to test this? It works fine in Firefox for me, but when I run it in Chrome and click the "Back to Patient" button, it opens the patient profile in a new tab. Do you get the same result?

interface/patient_file/summary/rx_left.php
```diff
@@ -18,6 +18,13 @@
 <span><?php xl('Add','e');?></span></a>
 </td>
 </tr>
+<tr>
+		<td>
+			<a class="css_button large_button" href="demographics.php" target="pat" onclick="top.restoreSession()">
```
maggienegm, line 23: Where did the target name "pat" come from in both files?

PR #1608, opened 2020-04-22
Create a whitelist to prevent LFI attacks & directory traversals

maggienegm:
Removing the url parameter at the end of the app's main endpoint (libreehr/interface/main/tabs/main.php?url=TAB_ONE_DEFAULT) will prevent users from uploading harmful files aka local file inclusion attacks.


mhn94:
By default, two tabs are chosen and displayed when a user logs in. A user can go in the settings and change that if needed. (One way to do this is to choose “User Preferences” under username in top right corner > Appearance tab.) I don't think there's a need for users to be able to "upload" a tab locally using the url parameter.

aethelwulffe:
Better check Role Based Menus and Single Application Access before you
chop that out. 

Don't remove a stone wall from the middle of a field until you make sure
it isn't actually a Levee....

robbyoconnor:
@aethelwulffe would you like me to add you back to the org so you can review these? We have a lot of work on this likely to last the summer and could use another eye

KoniKodes:
@aethelwulffe It is good to see you sir. Your guidance would be much appreciated by my team members.

They have also been working on the GACL and ACL roles - hoping this will be helpful as well.

mhn94:
Thanks for the guidance @aethelwulffe . I'll take a look, and lesson learned!

mhn94:
My new changes start a whitelist of acceptable file names from the interface/main/main_screen.php file, since it uses the url parameter. This will help prevent LFI attacks and directory traversals.

I had trouble finding how Role Based Menus and Single Page Apps use the url parameter as suggested by @aethelwulffe . I would appreciate any guidance regarding this!

interface/main/tabs/main.php
```diff
@@ -90,15 +90,112 @@ function isEncounterLocked( encounterId ) {
...
+                /*
+                To prevent local file inclusion (LFI) attacks and directory traversals, use a whitelist of acceptable urls that will be compared against.
+                The current list of whitelisted urls (and their required and optional parameters) come from interface/main/main_screen.php.
+                */
+                $inputURL = $_REQUEST['url'];
+                $allowedURLs = array(
+                    "pwd_expires_alert.php",
+                    "../patient_file/summary/demographics.php" => array(
+                        "required" => "set_pid",
+                        "optional" => "set_encounterid"
```
muarachmann, line 109: I like this approach ...

muarachmann: Kinna held up will have a look when I can :)

KoniKodes: Thank you @muarachmann
Your help is greatly appreciated.

mhn94: Thank you @muarachmann ! I really appreciate this! :)

interface/main/tabs/tab_redirect_error_message.php
```diff
@@ -0,0 +1,3 @@
+<?php
+    print("Please enter a valid URL.");
```
muarachmann, line 2: You can just point them to a 404 custom or page not found

interface/main/tabs/main.php
```diff
@@ -90,15 +90,112 @@ function isEncounterLocked( encounterId ) {
...
+                                        $year = substr($inputParamVal, 0, 4);
+                                        $month = substr($inputParamVal, 4, 2);
+                                        $day = substr($inputParamVal, 6, 2);
+                                        if(!checkdate($month, $day, $year)) {
+                                            $check = false;
+                                            break;
+                                        }
+                                    } else if(!preg_match($queryPatterns[$inputParamName], $inputParamVal)) {
+                                        $check = false;
+                                        break;
```
muarachmann, line 178: A lot of conditioning ... Maybe we should have some time to check out ourselves with sample examples

muarachmann: Looks good to me,

muarachmann (approved):
LGTM

Issue #1459, opened 2019-03-21
Inappropriate Module Name


Meghana122:
A) **Outreachy Username**: meghana 

B) **Issue title**: Inappropriate module name.
 
C) **Report Date**: March 21, 2019 

D) **Sites Affected**: Both Documentation and NHANES sites. 

E)  **Browsers used**: Microsoft Edge and Google Chrome on Windows 10.

F) **EHR Workflow module**: Miscellaneous => Portal Dashboard, there click on the “Mail” button. 

G) **Steps to Reproduce the Issue**:  
         1) Login to the LibreHealth documentation site with your login credentials. 
         2) On the main screen, under “Miscellaneous” click on “Portal Dashboard”. 
         3) In Portal Dashboard module, you will find a “Mail” button, click on that. 

H) **Expected Behavior**: The Module name should be appropriate and related to Mail. 

I) **What Actually Happened**: The Module name was “1”, which is inappropriate, for a module related to Mail. 

J) **Screenshots with Explanation**:   
        1) The module name is “1”, as shown in the highlighted portion.
         
![WrongModuleName](https://user-images.githubusercontent.com/40365198/54763324-ca196280-4c1b-11e9-9104-bfe4074c96e4.png)

 2) But the module is related to Mail, and the name related to mailing can be considered appropriate.

K) **Estimation of Severity**: Workflow does not get disturbed but it may confuse the user(Minor).

M) **Suggested Solution**:  
             The most possible solution can be, changing the name of the module according to the function which the module performs. In this case, the module name can be changed to “Patient Mail” from “1”.
          
![AppropriateName](https://user-images.githubusercontent.com/40365198/54763899-cb975a80-4c1c-11e9-8e2b-0611fba5ca0d.png)

muarachmann:
Nice one there, quick fix for this would be appreciated. You've got this @Meghana122

Meghana122:
Thank you.

Meghana122:
@muarachmann I have tried finding the correct file to fix the issue, but I couldn't. Please could you suggest a way to figure that out.

muarachmann:
@Meghana122 right click on the frame and chose view frame source code. I am sure you will have your answers :)

Meghana122:
Thank you for helping me with that.

aethelwulffe:
[...]

PR #1611, opened 2020-05-24
A fix inappropriate module name

ahanipeitra1994:
I did a fix on the issue #1459.
image
I added the module name header.


muarachmann:
Can you remove the other secure message below? sounds redundant 
Screen Shot 2020-06-10 at 4 04 41 AM

ahanipeitra1994:
Ok i will remove that

maggienegm:
@ahanipeitra1994 By default, the tab title shows within the tab, which is why "Secure Message" shows underneath the tab title. Since the module already has a name ("Patient Messaging"), I would make that the title, which will require you to move that piece of code up into where you added a title, so the tab name will become "Patient Messaging" and logo + title within the tab will display the way it's currently being displayed.

patient_portal/messaging/messages.php
```diff
@@ -86,7 +86,7 @@ function getAuthPortalUsers()
 <!DOCTYPE html>
 <html lang="en">
 <head>
-<meta http-equiv="content-type" content="text/html; charset=UTF-8">
+<!-- <meta http-equiv="content-type" content="text/html; charset=UTF-8"> -->
 <meta charset="utf-8">
```
maggienegm, line 90: Is there a reason why you removed some stuff from the <meta> tag?

Ngai-E:
Works fine now. @maggienegm   the previous format is kind of obsolete with majority of the web using  HTML5. `content-type` was used to specify character encoding but now we can just use this `<meta charset="utf-8">`  directly. Thanks for taking out time to refactor the code base as well @ahanipeitra1994.

patient_portal/messaging/messages.php
```diff
@@ -86,7 +86,7 @@ function getAuthPortalUsers()
 <!DOCTYPE html>
 <html lang="en">
 <head>
-<meta http-equiv="content-type" content="text/html; charset=UTF-8">
+<!-- <meta http-equiv="content-type" content="text/html; charset=UTF-8"> -->
```
Ngai-E, line 89: This comment is not necessary, remove it.

patient_portal/messaging/messages.php
```diff
@@ -86,7 +86,7 @@ function getAuthPortalUsers()
 <!DOCTYPE html>
 <html lang="en">
 <head>
-<!-- <meta http-equiv="content-type" content="text/html; charset=UTF-8"> -->
+<meta http-equiv="content-type" content="text/html; charset=UTF-8">
```
Ngai-E: I asked to remove the comment, not uncomment it.

patient_portal/messaging/messages.php
```diff
@@ -110,11 +109,7 @@ function getAuthPortalUsers()
 <!--<script type='text/javascript' src="./appmail.js.php"></script>  -->
 
 </head>
-<!-- style='background:#f1f2f7;' -->
-<body class="body_top">
-<div class='title'>
-        <h2><img style='width:25%;height:auto;' class='logo' src='<?php echo $GLOBALS['images_path']; ?>/logo-full-con.png'/>  <?php echo xlt('Patient Messaging'); ?></h2>
-        </div>
```
maggienegm: @ahanipeitra1994 it looks like with your last commit, you removed the solution you pushed earlier, so now the original issue still exists. Make sure to always test before you push!

maggienegm:
Tested and LGTM. @Ngai-E and @muarachmann ?

patient_portal/messaging/messages.php
```diff
@@ -111,6 +110,9 @@ function getAuthPortalUsers()
 
 </head>
 <body ><!-- style='background:#f1f2f7;' -->
+<div class='title container'>
+        <h2><img style='width:25%;height:auto; class='logo' src='<?php echo $GLOBALS['images_path']; ?>/logo-full-con.png'/>  <?php echo xlt('Patient Messaging'); ?></h2>
+        </div>
```
muarachmann, line 115: I think it should be `patient portal`

ahanipeitra1994:
Ok thanks I will change it

Issue #1602, opened 2020-04-03
Buttons Style In Message and Reminder Center


meshubhangi:
**A) Outreachy Username:** shubhangi.pandey 	
**B) Issue title:** Buttons Style In Message and Reminder Center
**C) Report Date:** 31st March, 2020
**D) Sites Affected:** Local Instance
**E) OS/ browser used:** Mozilla Firefox
**F) EHR Workflow module:** Message And Reminder Center
**G) Steps to Reproduce the Issue:** 

- Login as Admin
- Click on “Messages” tab on the top navbar
- At the bottom of “Message and Reminder Center” you can find the “Add New” and “Delete” button not styled correctly

**H) Expected Behavior:** Buttons styled in a good manner.
**I) What Actually Happened:** Whenever I open the Message and Reminder Center and try to send a message, I come across the “Add New” button which is styled improperly. Also, when I want to delete any previous message sent, I come across “Delete” button which too is styled improperly. I am facing this issue in Chrome, Firefox as well as Chromium.
**J) Screenshots with Explanation:** 
 1. Here’s the “Add New” and “Delete” button referred in this issue report.
![bug_shubhangi_pandey](https://user-images.githubusercontent.com/22342441/78389927-456c6d00-7601-11ea-9430-484449b3d159.png)

2. On inspecting, found the background theme URL of button missing. 
![screenshot2](https://user-images.githubusercontent.com/22342441/78390050-7fd60a00-7601-11ea-923f-9ab98352263e.gif)
**K) Estimation of Severity:** Trivial: Cosmetic defect only; has no effect on Workflow or Module's actual function.
**L) Workaround:** NA
**M) Solution:** The CSS of the button can be fixed
![solution-screenshot](https://user-images.githubusercontent.com/22342441/78390120-a3995000-7601-11ea-918b-ac85acea798e.gif)

Ngai-E:
Thanks please fix it if you can.

meshubhangi:
Yes, I can.

Paras2Bhardwaj:
i want to do this

PR #1612, opened 2020-06-03
A Fix on Message and Reminder Button

ahanipeitra1994:


ahanipeitra1994:
This is a fix of issue #1602

maggienegm:
The buttons within this module (as shown in the image) look great! The only thing I would change is centering the text within the buttons. 

However, take a look at the following buttons in other places of the app:

- Demographics page after choosing a patient from "Patient Finder" module
demographics


- Reports > Clients > Patient List
pt_list_report

Your changes affect all other instances in the app that use the "css_button" class and a[role="button"], so you have to be careful. Can you fix that too?

ahanipeitra1994:
ok let me check it out i never noticed

PR #1616, opened 2020-06-12
XSS File Updates

maggienegm:
This PR mainly includes file updates with any missing HTML encoding.

Another change includes updating the files to the current XSS functions that are defined [htmlspecialchars.inc.php](https://github.com/LibreHealthIO/lh-ehr/blob/master/library/htmlspecialchars.inc.php), such as:
    • `htmlspecialchars()` or `htmlspecialchars(xl())` function calls are changed to one of the following: `xlt()`, `xla()`, `text()`, or` attr()`, depending on the parameters passed in.
    • If any `xl()` function calls include the `‘e’` parameter, the `‘e’` parameter is removed and `echo` is used instead, if it hasn’t already been so.


maggienegm:
Another XSS prevention is doing input validation before the data is used

controllers/C_Document.class.php
```diff
@@ -241,7 +241,7 @@ function note_action_process($patient_id) {
         } elseif ($_POST['identifier'] == "yes") {
 
             // send the document via email
-            $d = new Document($_POST['foreign_id']);
+            $d = preg_match("/^\d+$/", $_POST['foreign_id']) ? new Document($_POST['foreign_id']) : new Document();
```
muarachmann, line 244: I guess this was making sure it's a digit only right.

maggienegm: Yep, and it instantiates a Document object accordingly

Issue #1555, opened 2020-02-16
Patient Issues Show Incorrect Dates When Left Blank


mhn94:
A) **Outreachy Username:** maggie.negm
B) **Issue Title:** Patient Issues Show Incorrect Dates When Left Blank
C) **Report Date:** 2/14/2020
D) **Sites Affected:** Documentation Site, NHANES
E) **OS/Browser Used:** Chrome browser in Linux OS
F) **EHR Workflow Module:** Issues
G) **Steps to Reproduce the Issue:** 

1. Go to a patient profile.
- There are many ways to do this. One way is to use the Patient Finder (“Patient/Client” from the main menu > “Finder” under “Patient/Client”) and choose a patient.
2. Select “Issues” under the patient’s name.
3. Create any issue by clicking the “Add” button next to an issue section.
4. Fill out the form and leave the “Begin Date,” “End Date,” or both blank.
5. Select the “Save” button.

H) **Expected Behavior:** When no begin or end date is entered during an issue’s creation, no timestamp should show under the date columns (“Begin” and “End”) in the “Issues” summary.
I) **What Actually Happened:** When no date is entered during an issue’s creation, the “Issues” module shows a specific timestamp. In the Docs server, it shows “1969-12-31 19:00,” and in the NHANES server, it shows “01/01/1970 00:00.”
J) **Screenshots with Explanation:**
After creating issue(s) within a patient’s profile (using the Docs server), the “Issues” module will list the patient issues in the following format: 

![issues-w-outline](https://user-images.githubusercontent.com/8771586/74613710-9ae6cc00-50de-11ea-8961-a9c564183e81.png)

- The “asthma” issue was created with a begin date of 01/09/2020 and no end date. The begin date is displayed correctly, but the “End” column shows the timestamp “1969-12-31 19:00.”

- The “codeine” issue was created with no begin or end date. Both the “Begin” and “End” columns show the timestamp “1969-12-31 19:00.”

- The “iodine” issue was created with no begin date and an end date of 02/12/2020, and the “Begin” column shows the timestamp “1969-12-31 19:00.” and the “End” column shows the correct date.

The following is the code that displays the form within Chrome Inspector:
![patient-stats-console](https://user-images.githubusercontent.com/8771586/74613717-a9cd7e80-50de-11ea-9b8f-26ca6d83ac74.png)

K) **Estimation of Severity:** Minor - distracting or confusing but workflow can continue correctly
[...]

PR #1617, opened 2020-06-19
Fix in Patient Issues Showing Incorrect Date

ahanipeitra1994:
This is  a fix in issue #1555. The Patient 'Issues' now does not display incorrect date without entry.

### DIFF
--- a/interface/patient_file/summary/stats_full.php
+++ b/interface/patient_file/summary/stats_full.php
@@ -285,8 +285,8 @@ foreach ($ISSUE_TYPES as $focustype => $focustitles) {
 
     echo " <tr class='$bgclass detail' $colorstyle>\n";
     echo "  <td style='text-align:left' data-text='$disptitle' class='$click_class' id='$rowid'>" . text($disptitle) . "</td>\n";
-    echo "  <td>" . text(date(DateFormatRead(true), strtotime($row['begdate']))) . "&nbsp;</td>\n";
-    echo "  <td>" . text(date(DateFormatRead(true), strtotime($row['enddate']))) . "&nbsp;</td>\n";
+    echo "  <td>" . text($row['begdate']) . "&nbsp;</td>\n";
+    echo "  <td>" . text($row['enddate']) . "&nbsp;</td>\n";
     // both codetext and statusCompute have already been escaped above with htmlspecialchars)
     echo "  <td>" . $codetext . "</td>\n";
     echo "  <td>" . $statusCompute . "&nbsp;</td>\n";


Ngai-E:
> This is a fix in issue #1555. The Patient 'Issues' now does not display incorrect date without entry.

please can you send us a screenshot of the fix.

ahanipeitra1994:
ok let me do that now

ahanipeitra1994:
This is after the fix. The beginning and end date does not show again
unless the field is entered.
[image: image.png]

interface/patient_file/summary/stats_full.php
```diff
@@ -285,8 +285,8 @@ function newEncounter() {
 
     echo " <tr class='$bgclass detail' $colorstyle>\n";
     echo "  <td style='text-align:left' data-text='$disptitle' class='$click_class' id='$rowid'>" . text($disptitle) . "</td>\n";
-    echo "  <td>" . text(date(DateFormatRead(true), strtotime($row['begdate']))) . "&nbsp;</td>\n";
-    echo "  <td>" . text(date(DateFormatRead(true), strtotime($row['enddate']))) . "&nbsp;</td>\n";
+    echo "  <td>" . text($row['begdate']) . "&nbsp;</td>\n";
+    echo "  <td>" . text($row['enddate']) . "&nbsp;</td>\n";
```
muarachmann, line 289: This is not the right approach! Just check if the value is `NULL` or `""` then show a default `-`, the functions there are not be removed

ahanipeitra1994:
ok

ahanipeitra1994:
I will fix that

interface/patient_file/summary/stats_full.php
```diff
@@ -285,8 +285,8 @@ function newEncounter() {
 
     echo " <tr class='$bgclass detail' $colorstyle>\n";
     echo "  <td style='text-align:left' data-text='$disptitle' class='$click_class' id='$rowid'>" . text($disptitle) . "</td>\n";
-    echo "  <td>" . text($row['begdate']) . "&nbsp;</td>\n";
-    echo "  <td>" . text($row['enddate']) . "&nbsp;</td>\n";
+    echo "  <td>" . text(date(DateFormatRead(true), date($row['begdate']))) . "&nbsp;</td>\n";
+    echo "  <td>" . text(date(DateFormatRead(true), date($row['enddate']))) . "&nbsp;</td>\n";
```
muarachmann, line 289: Yes we need some if/else statement here for blank dates

ahanipeitra1994:
I tried the if/else statement but it was not functional then noticed the
problem was caused by the "strtotime()" which outputs 1970 01 01 as date by
default so thats y i changed it to a normal date

maggienegm:
@ahanipeitra1994 it seems to still show the default timedate for me, even with date(). Can you push the if/else statements that you used? We can help guide you to a working solution if we can see the code.

ahanipeitra1994:
ok i will do that

PR #1621, opened 2020-06-29
begin work on csrf fixes

Ngai-E:
This is the first test commit towards fixing the csrf vulnerability. Basically, we create a token to validate requests that can manipulate data on the system and verify this each time such request is made. 

This approach makes use of the openssl php extension which can be activated as follows;
1. Open php.ini file located under php installation folder.
2. Search for extension=php_openssl.dll.
3. Uncomment it by removing the semi-colon(;) in front of it.
4. Restart the Apache Server.

So far only the letter.php endpoint has been handled. More commits to come.

### DIFF
--- a/library/sanitize.inc.php
+++ b/library/sanitize.inc.php
@@ -55,5 +55,34 @@ function image_has_right_size($size) {
   return $size < 20971520;
 }
 
+// Generate csrf token for authentication
+function generateCsrfToken()
+{
+    if (!extension_loaded('openssl')) {
+      error_log("ERROR: openssl extension not enabled, needed for proper functioning of LibreHealth");
+        die("Error: Systems needs openssl.");
+    }
+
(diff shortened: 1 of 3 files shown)


library/sanitize.inc.php
```diff
@@ -55,5 +55,34 @@ function image_has_right_size($size) {
...
+
+    return $csrfToken;
+}
+
+// Function to verify a csrf token
+function verifyCsrfToken($token)
+{
+    if (hash_equals($_SESSION['token'], $token)) {
+        return true;
+    } else {
```
muarachmann, line 81: Great initiative @Ngai-E looking forward to this ... One thing can you create a class for this say something like `CsrfToken::class`

Ngai-E: Thanks, @muarachmann we will surely do that.

PR #1624, opened 2020-07-11
Removed calls to files that weren't used && w/ security issues

maggienegm:


maggienegm:
One feature within Admin > Files allows admin users to change source code within the app. Rather than allowing users to edit code files within the app and allowing security vulnerabilities to exist, programmers can edit the files.

PR #1625, opened 2020-07-15
fit remote code execution in history and template imports functionality

Ngai-E:

PR #1634, opened 2020-08-01
CSRF Fixes and Remote Code Execution Prevention

maggienegm:
This fix removes CSRF vulnerabilities and also prevents remote code execution attempts.


maggienegm:
> Its necessary to type cast the inputs before supplying to hash_hmac, interesting things might occur if that is not done ( type juggling vulnerability ), if $secondToken is a user controlled input, if an array is passed, hash_hmac will return null.
> For more details:
> https://www.youtube.com/watch?v=MpeaSNERwQA

Added type casting to hash_hmac()'s input per @naveen17797 's comment in PR #1625

PR #1640, opened 2020-08-11
Added multi-site admin functionality

maggienegm:
Added an admin.php file to handle multiple sites and their respective setup and configuration. You have to manually type the URL to access admin.php, which requires user authentication before its contents are accessed. Currently, it includes the following functionalities:

- Upgrade Database (`setup/sql_upgrade.php`)
- Access Control Upgrade (`acl_upgrade.php`)
- Patch Database (`sql_patch.php`)

This will help prevent unauthenticated info disclosure and unauthenticated admin actions.

Context: [PR 1638](https://github.com/LibreHealthIO/lh-ehr/pull/1638)

### DIFF
--- a/acl_upgrade.php
+++ b/acl_upgrade.php
@@ -107,6 +107,7 @@
 */
 
 $ignoreAuth = true; // no login required
+$aclUpgradeConfig = true; // to correctly redirect to login page if needed
 
 require_once('interface/globals.php');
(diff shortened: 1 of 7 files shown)


maggienegm:
Hmmm, I thought "default" was just a value; "default" is a site ID name and is the default value when creating a site using `setup.php`.

PR #1687, opened 2023-03-24
Update INSTALL.md

anmode:
I was following the installation guide and found this error. Thanks

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -467,7 +467,7 @@ Clone the forked repository to a local directory with:
 #### Installation
 Launch MAMP Control Panel and start your server.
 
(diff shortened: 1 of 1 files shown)


anmode:
But i think it needs to be updated as when we git clone folder named `lh-ehr` comes where whole source code is present not the `ibrehealthehr`. at this `localhost/librehealthehr/setup.php` it shows no page , Please

tmccormi:
I agree that the process needs to be vetted and refined.

anmode:
>  I agree that the process needs to be vetted and refined.

Should i give it a try. ?
I really face problem...but finally successfully setup the project

